PR #1511, opened 2025-08-02
build: use cargo-zigbuild

DavSanchez:

While compiling the Rust packages while using different container engines for macOS I stumbled upon issues with `cross` and emulation (e.g. `qemu`), as the `x86_64-unknown-linux-musl` image used by `cross` is not available for `aarch64`. Specifically, compiling certain dependencies (like `aws-lc-sys`) under `qemu`-based emulation does not seem to work.

As the container-based approach can be slow and makes us subject to these kinds of problems, with this PR I switch the build procedure used to generate AC binaries by natively compiling the sources to the appropriate targets and linking using the [Zig toolchain](https://ziglang.org/) instead, via a utility called [`cargo-zigbuild`](https://github.com/rust-cross/cargo-zigbuild). Building AC locally now needs:

- The required targets being added to the Rust toolchain (e.g. via `rustup target add <TARGET>`).
- Zig present on the machine.
- `cargo-zigbuild`, installable via `cargo install --locked cargo-zigbuild`.

The CI tooling for Kubernetes should work the same as this PR makes `tilt` use the Zig-based compilation by default.

It also changes all the builds (excluding the tests via `cargo test ...`) to use the actual targets that we distribute. This should help us with any false alarms in the CI. Previously, using the default Linux machine in the CI, our calls to `cargo build` were not specifying any target so they would target `x86_64-unknown-linux-gnu` (GNU libc), making it different to the build we actually distribute in our releases, which is the `x86_64-unknown-linux-musl` (musl libc).

This should speed up the compilation process overall by:
- Getting rid of containers and required infrastructure (i.e. a VM on macOS) to build natively. 
- Letting us recover the parallelism on the GoReleaser calls, which was disabled due to a race condition in the calls to `docker` implicit in `cross`.

Also, the fact that this approach is currently the default used to provide [Rust support with cross-compilation in the latest versions of GoReleaser](https://goreleaser.com/customization/builds/rust/?h=rust) should allow us to easily transition towards a [newer GoReleaser version](https://github.com/goreleaser/example-rust/blob/main/.goreleaser.yaml) and streamline the whole process. It gives me the confidence that it's just a better approach than `cross`.

## Checklist


docs/DEVELOPMENT.md
sigilioso, line 1: It is in the PR description but, should we mention the zig installation somewhere in this doc? Otherwise the steps to compile locally will fail with:

```
Error: Failed to find zig

Caused by:
    cannot find binary path
```

sigilioso (dismissed):
Thanks!

docs/DEVELOPMENT.md
DavSanchez: Added in https://github.com/newrelic/newrelic-agent-control/pull/1511/commits/b1f67eb4c209902a905795d823ccc027b0dd7739!

sigilioso:
When trying locally I run into this zig issue: https://github.com/ziglang/zig/issues/23273

```
❯ zig version
0.14.1

❯ ulimit -n
256

❯ cargo zigbuild --bin newrelic-agent-control-onhost --target aarch64-unknown-linux-musl
[...]
  = note: some arguments are omitted. use `--verbose` to show all linker arguments
  = note: error: unable to search for static library /Users/calvarez/repos/github.com/newrelic/newrelic-agent-control/target/aarch64-unknown-linux-musl/debug/deps/libpotential_utf-dc5d986049886d53.rlib: ProcessFdQuotaExceeded
```

DavSanchez:
> When trying locally I run into this zig issue: [ziglang/zig#23273](https://github.com/ziglang/zig/issues/23273)
> 
> ```
> ❯ zig version
> 0.14.1
> 
> ❯ ulimit -n
> 256
> 
> ❯ cargo zigbuild --bin newrelic-agent-control-onhost --target aarch64-unknown-linux-musl
> [...]
>   = note: some arguments are omitted. use `--verbose` to show all linker arguments
>   = note: error: unable to search for static library /Users/calvarez/repos/github.com/newrelic/newrelic-agent-control/target/aarch64-unknown-linux-musl/debug/deps/libpotential_utf-dc5d986049886d53.rlib: ProcessFdQuotaExceeded
> ```

Very interesting issue! Given here we use a script, I have added a small condition that should set it for macOS, but unfortunately this won't translate well into the follow-up GoReleaser PR as it gets rid of the script. I'll think on alternatives, but meanwhile, if you're able to reproduce this, can you try again with the latest state of this branch?

sigilioso:
I think `ulimit` is set at session level, if we are worried about local environments we can add some lines to the `DEVELOPMENT.md` file and assure everybody sets it up in their shell session file. WDYT?

danielorihuela:
> Good to know!
> 
> I have added [a similar check on the GoReleaser branch](https://github.com/newrelic/newrelic-agent-control/pull/1533/commits/11b40fb7da4a3622b627296c2bda4ec62d12b1b4), but I'm not sure if the `ulimit` is set for the whole session or just the current process and sub-processes. If the latter I think it should not work. When you have some time, can you check there as well?
> 
> A command like `goreleaser release --skip=publish --skip=sign --clean --verbose --timeout 2h --snapshot` should trigger the build.

Captura de pantalla 2025-08-21 a las 14 52 14

build/scripts/build_binary.sh
```diff
@@ -28,8 +34,12 @@ export GIT_COMMIT
 export AGENT_CONTROL_VERSION=${AGENT_CONTROL_VERSION}
 
 export RUSTFLAGS="-C target-feature=+crt-static"
-export CROSS_CONFIG=${CROSS_CONFIG:-"./Cross.toml"}
-cross build --target "${ARCH_NAME}-unknown-linux-musl" --profile "${BUILD_MODE}" --package "${PKG}" --bin "${BIN}"
+
+# if [[ "$OSTYPE" == "darwin"* ]]; then
+#   echo "macOS detected, increasing ulimit to 4096 for concurrent Zig linking"
+#   ulimit -n 4096
+# fi
```
sigilioso, line 41: nit: I'd remove it instead of commenting it out

docs/DEVELOPMENT.md
```diff
@@ -2,16 +2,33 @@
...
+        = note: some arguments are omitted. use `--verbose` to show all linker arguments
+        = note: error: unable to search for static library /<SOME_PATH_TO_RLIB_FILE>.rlib: ProcessFdQuotaExceeded
+      ```
+
+      This is a [known](https://github.com/ziglang/zig/issues/23273) [issue](https://github.com/rust-cross/cargo-zigbuild/issues/329). To address it, increase the number of file descriptors for the current shell session with:
+
+      ```sh
+      ulimit -n 4096
+      ```
+
```
sigilioso, line 29: Nice!

sigilioso (approved):
LGTM! Thanks!

Although I would try to make everybody aware of the change before merging (just to avoid issues in dev environments)

Tiltfile
```diff
@@ -15,10 +17,11 @@ cluster = os.getenv('CLUSTER', "")
 chartmuseum_basic_auth = os.getenv('CHARTMUSEUM_BASIC_AUTH', "")
 
 # build_with options:
-# cargo: No crosscompilation, faster than cross
-# cross: Supports crosscompilaton
-build_with = os.getenv('BUILD_WITH','cross')
+# cargo: No crosscompilation, faster than cargo-zigbuild
+# zig: Supports crosscompilaton
+build_with = os.getenv('BUILD_WITH','zig')
```
gsanchezgavier, line 22: perhaps we could go just with zig i guess and simplify this , but we can do it in another PR

Tiltfile
```diff
@@ -36,11 +39,16 @@ if build_with == 'cargo':
         './agent-control',
       ]
   )
-elif build_with == 'cross': 
+elif build_with == 'zig':
   local_resource(
       'build-binary',
-      cmd="make BUILD_MODE=debug ARCH=%s build-agent-control-cli" % arch +
-           "&& make BUILD_MODE=debug ARCH=%s build-agent-control-k8s" % arch ,
```
gsanchezgavier, line 43: would make sense to change the make targets instead?

gsanchezgavier (commented):
cool!! just left minor questions

Tiltfile
DavSanchez: Done in https://github.com/newrelic/newrelic-agent-control/pull/1511/commits/3011912f72ed6290360a7372530c8a88eebc95f7!

PR #1576, opened 2025-08-22
chore: make errors lowercase

danielorihuela:

Update style docs.

### DIFF
--- a/docs/style/errors.md
+++ b/docs/style/errors.md
@@ -1,0 +1,54 @@
+# Errors
+
+## Error messages
+
+Error messages **must begin with a lowercase letter and should not end with a period**.
+
+Examples:
+
+```rust
+// 👍 Good:
+#[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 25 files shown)


docs/style/errors.md
```diff
@@ -0,0 +1,54 @@
...
+// 👎 Bad:
+// It must start with lowercase letter
+#[derive(Debug, thiserror::Error)]
+#[error("Serializing string: {0}")]
+pub struct Serialize(String);
+// It should not end with a period
+#[derive(Debug, thiserror::Error)]
+#[error("serializing string: {0}.")]
+pub struct Serialize(String);
+```
```
sigilioso, line 24: Nice!

Would also make sense to include some examples using `map_error`? I know it is the same, but the code is pretty different so, maybe having both examples is useful!

sigilioso (dismissed):
Thank you! 🚀

sigilioso (approved):

PR #1577, opened 2025-08-22
feat: make deployment release name configurable

danielorihuela:

Makes `agent-control-deployment` release name configurable.

## Which issue this PR fixes

- fixes #NR-441446

## Checklist

### DIFF
--- a/agent-control/tests/k8s/self_update.rs
+++ b/agent-control/tests/k8s/self_update.rs
@@ -42,6 +42,7 @@ fn k8s_self_update_bump_chart_version_from_last_release_to_local_new_config() {
         &opamp_server,
         &namespace,
         CHART_VERSION_LATEST_RELEASE,
+        "agent-control-deployment",
     );
 
     let ac_config = format!(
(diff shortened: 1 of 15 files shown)


agent-control/tests/k8s/self_update.rs
```diff
@@ -42,6 +42,7 @@ fn k8s_self_update_bump_chart_version_from_last_release_to_local_new_config() {
         &opamp_server,
         &namespace,
         CHART_VERSION_LATEST_RELEASE,
+        "agent-control-deployment",
```
danielorihuela, line 45: Ideally, we want each test to have it's own release name, so that we avoid collisions between tests.
The problem is that this test uses the "latest" image of agent control, which doesn't take into account the configured name yet. We have two options:

* Change and merge with a test in red (I don't like it)
* Don't change it and merge with a test in green (potential issues in the future)

I think the second option is best. As soon as we make a release, we could update this test to avoid having issues in the future. We have to work a little bit more, but we avoid momentarily breaking a test and potentially hiding a bug or some other error.

If you like the idea, answer this comment or put a thumbs up emoji. Then, I'll create a task to handle that after the release and pass it to the hero.

PR #1580, opened 2025-08-25
fix(ac-updater): avoid fetching resources when update is not needed [NR-453803]

sigilioso:

This PR changes the Agent Control version updater in order to avoid fetching k8s resources when the version is not informed.

## Special notes for your reviewer

- This caused a bug when the chart weren't fully ready and the agent-control-cd wasn't created even if the remote updates for agent-control-cd were enabled.
- Unit tests have been extended to assure that the k8s client doesn't try to fetch helm-releases in the updater unless they are actually needed.
- Some fields in the AC's k8s updater have been renamed for the shake of readability.

## Checklist

### DIFF
--- a/agent-control/src/agent_control/version_updater/k8s.rs
+++ b/agent-control/src/agent_control/version_updater/k8s.rs
@@ -24,8 +24,8 @@ impl fmt::Display for Component {
     }
 }
 pub struct K8sACUpdater {
-    ac_remote_update: bool,
-    cd_remote_update: bool,
+    ac_remote_update_enabled: bool,
+    cd_remote_update_enabled: bool,
     k8s_client: Arc<SyncK8sClient>,
     namespace: String,
     // current_chart_version is the version of the agent control that is currently running.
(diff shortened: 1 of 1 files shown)


agent-control/src/agent_control/version_updater/k8s.rs
```diff
@@ -24,8 +24,8 @@ impl fmt::Display for Component {
     }
 }
 pub struct K8sACUpdater {
-    ac_remote_update: bool,
-    cd_remote_update: bool,
+    ac_remote_update_enabled: bool,
```
DavSanchez, line 27: 💯

DavSanchez (approved):
At first I wasn't sure what did we win with this change since it seems it only moves the version retrieval to after we know if we have a new version or not, but we already had the booleans that seemed to serve a similar purpose. Probably we could make this check earlier, merging with the boolean, and make `update_helm_release_version` not accept an `Option<&String>` for the version but the version directly, so we know we only call it when we have it, but that's for another PR :)

Thanks! 🚀

sigilioso:
> At first I wasn't sure what did we win with this change since it seems it only moves the version retrieval to after we know if we have a new version or not, but we already had the booleans that seemed to serve a similar purpose. Probably we could make this check earlier, merging with the boolean, and make `update_helm_release_version` not accept an `Option<&String>` for the version but the version directly, so we know we only call it when we have it, but that's for another PR :)
> 
> Thanks! 🚀

I actually tried something similar: a function checking if the update was needed or not for a component becaouse I thought that the `update_helm_release` function was already doing to much. It returned an enum as follows:

```rust
enum UpdateNeeded {
    No,
    Yes({target_version: String}
}
```
However, I realized that there were too many and I opted for this simple callback approach that already fixes the issue we were facing. If we ever need to extend this component we will likely consider something more structured (and avoid that ugly `Option<&String>`: if there is no version why trying to upgrade!). 🙂

PR #1594, opened 2025-08-28
test: improve k8s e2e troubleshooting logs

gsanchezgavier:
Add extra troubleshooting logs

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -143,19 +143,28 @@ if license_key != '':
 if cluster != '':
   ac_flags.append('--set=global.cluster='+cluster)
 
-
-local_resource(
-    'wait-install-job',
-    allow_parallel=True,
-    cmd="kubectl wait --for=create --timeout=200s job/ac-agent-control-install-job",
-    resource_deps=['local-child-chart-upload'],
-)
-
+#### Troubleshooting logs from installation Jobs
 local_resource(
     'log-install-job',
     allow_parallel=True,
-    serve_cmd="while true ; do kubectl logs -f job/ac-agent-control-install-job || sleep 1; continue ; done",
-    resource_deps=['wait-install-job'],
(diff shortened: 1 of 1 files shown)


Tiltfile
```diff
@@ -143,19 +143,28 @@ if license_key != '':
...
+)
+local_resource(
+    'log-system-identity-job',
     allow_parallel=True,
-    serve_cmd="while true ; do kubectl logs -f job/ac-agent-control-install-job || sleep 1; continue ; done",
-    resource_deps=['wait-install-job'],
+    # the SI job might not be triggered for some of the test but this resource must always be success for `tilt ci` execs in CI.
+    cmd="kubectl wait --for=create --timeout=200s job/agent-control-generate-system-identity || echo System Identity job not created",
+    serve_cmd="while true ; do kubectl logs -f job/agent-control-generate-system-identity 2>/dev/null || sleep 5; continue ; done",
+    resource_deps=['log-install-job'],
```
sigilioso, line 167: I wonder if we could avoid creating this resource if we know that the jobs won't be needed, after all in many test cases fleet is not enabled.

I was thinking of something like this (totally untested!):

```python
chart_values_content = read_file_stream(sa_chart_values_file)
is_fleet_control_enabled = (
  chart_values_content
  .get("agentControlDeployment", {})
  .get("chartValues", {})
  .get("config", {})
  .get("fleet_control", {})
  .get("enabled", true)
)
if is_fleet_control_enabled:
    # Create the local resource here
```

gsanchezgavier: I though about some conditional but then i decided to created anyway, they never fail, nor add extra times or load, they are just there for troubleshooting proposes. Do you see any issue on having them always? i think the code is a bit more simple this way

sigilioso: I was worried about the whole thing taking extra time because of the `wait`, but it doesn't take extra time thanks to the `allow_parallel` field, right?

sigilioso (approved):

PR #1617, opened 2025-09-05
chore(refactor): simplify agent-control command

sigilioso:
This PR simplifies how the Agent Control command is built and the main function executed. As we don't expect the commands to grow (since there is a separate binary to act as the cli) the command abstraction can be removed.

Summary of changes:

- Removes the previous Command abstraction
- Renames `Flag` to `Command`
- Unifies how the main function is called for k8s and on-host


DavSanchez (dismissed):
Thanks!! 🚀

DavSanchez (approved):

PR #1639, opened 2025-09-10
chore(hostname): align implementation

paologallinaharbur:
Unifies get_hostname functions, avoid useless structs
Follow up of past PR @sigilioso 

## Checklist

### DIFF
--- a/resource-detection/src/system/hostname.rs
+++ b/resource-detection/src/system/hostname.rs
@@ -1,22 +1,16 @@
 use super::detector::SystemDetectorError;
-#[cfg(unix)]
-use nix::unistd::gethostname;
-use std::ffi::OsString;
 
-/// wrapper for an hostname getter
-#[derive(Default)]
-pub struct HostnameGetter {}
+#[cfg(target_family = "unix")]
+/// hostname getter
+pub fn get_hostname() -> Result<String, SystemDetectorError> {
+    use nix::unistd::gethostname;
+    gethostname()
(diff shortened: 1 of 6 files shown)


resource-detection/src/system/hostname.rs
```diff
@@ -1,22 +1,16 @@
...
-    pub fn get(&self) -> Result<OsString, SystemDetectorError> {
-        gethostname().map_err(|e| SystemDetectorError::HostnameError(e.to_string()))
-    }
+#[cfg(target_family = "unix")]
+/// hostname getter
+pub fn get_hostname() -> Result<String, SystemDetectorError> {
+    use nix::unistd::gethostname;
+    gethostname()
+        .map_err(|e| SystemDetectorError::HostnameError(e.to_string()))
+        .map(|h| h.into_string().unwrap_or_default())
```
sigilioso, line 9: It was already like this, but would it make sense to use [to_string_lossy](https://doc.rust-lang.org/nightly/std/ffi/os_str/struct.OsString.html#method.to_string_lossy) instead of `into_string().unwrap_or_default()`?

sigilioso (approved):
Thanks for cleaning this up! I've left a comment regarding something we had already there, feel free to ignore it!

resource-detection/src/system/hostname.rs
paologallinaharbur: yes!

PR #1686, opened 2025-09-16
feat: support signature-algorithms in verifier

sigilioso:
This PR updates the `Verifier` trait to support different sets of algorithms depending on its implementation.

When removing support for the certificate approach we can simplify the SigningAlgorithm definition accordingly by removing the conversion to `&webpki::SignatureAlgorithm` and also removing algorithms that are not supported anymore (explained in the _TODOs_).


paologallinaharbur (approved):
Nice!

PR #1693, opened 2025-09-17
feat: update signature on OpAMP integration tests

sigilioso:
This PR extends the OpAMP's mock server in the integration tests to sign the remote configuration with a self-generated private key and expose the public key using JWKS. It also updates on-host and k8s integration tests to leverage this new feature.

Additionally, it includes an additional integration test to check that 'falling back to certificate validation' works as expected. In case that the configured JWKS serves a public key but the signatures are still signed using the legacy system.


gsanchezgavier (approved):
Awesome!

PR #1695, opened 2025-09-18
feat: add public key signature validation

gsanchezgavier:
- Adds the public key fetcher from the JWKS endpoint.
- Changes the current validator to use the public key as a first method and the cert as fallback to support transition.


agent-control/src/opamp/remote_config/validators/signature/validator.rs
```diff
@@ -191,6 +211,27 @@ impl RemoteConfigValidator for CertificateSignatureValidator {
             .map_err(|e| SignatureValidatorError::VerifySignature(e.to_string()))?
             .as_bytes();
 
+        // Until backend migrates to new signature platform, the validation starts with the public key based,
+        // and falls back to cert based in case of failure.
+        // This fallback mechanism makes errors misleading in case the platform is migrated and the validation fails
+        // since the showed error is from the cert validation.
```
sigilioso, line 217: As discussed _offline_, we can an additional log message (maybe in INFO level) in case we fall-back to certificate validation: the misleading error is still shown but we so a clarifying comment right after.

sigilioso (approved):
Just a small comment regarding the logs. LGTM already!

sigilioso (approved):

PR #1730, opened 2025-09-29
chore: remove certificate validator

gsanchezgavier:
- Removes the remote config validation based on the public key extracted from the certificate, which is deprecated by the JWKS one. 

NOTE: To be merged after FC deploys the new signature.


agent-control/src/agent_control/config.rs
```diff
@@ -181,7 +181,7 @@ impl<'de> Deserialize<'de> for OpAMPClientConfig {
             #[serde(default)]
             fleet_id: String,
             #[serde(default)]
-            pub signature_validation: SignatureValidatorConfig,
+            pub signature_validation: Option<SignatureValidatorConfig>,
```
gsanchezgavier, line 184: I changed the interface to disable the validation in case no config is added but realized that makes more dificult to manage overrides of this in the chart. I will revert back to the `enabled` flag approach, it's also more extensible

agent-control/src/opamp/remote_config/validators/signature/validator.rs
```diff
@@ -1,199 +1,117 @@
...
-/// can be removed.
-pub struct CompositeSignatureValidator {
-    certificate_store: VerifierStore<Certificate, CertificateFetcher>,
-    public_key_store: Option<VerifierStore<PublicKey, PublicKeyFetcher>>,
-}
+        let Some(public_key_server_url) = config.public_key_server_url else {
+            return Err(SignatureValidatorError::BuildingValidator(
+                "missing public_key_server_url configuration".to_string(),
+            ));
+        };
```
sigilioso, line 70: We fail if enabled and url is not set, nice!

agent-control/tests/common/opamp.rs
```diff
@@ -52,12 +45,10 @@ struct AgentState {
 }
 
 impl ServerState {
-    fn new(cert_key_pair: KeyPair, use_legacy_signatures: bool) -> Self {
+    fn new() -> Self {
```
sigilioso, line 48: clippy doesn't complain here because it isn't public, right?

agent-control/src/agent_control/run.rs
```diff
@@ -160,10 +158,10 @@ impl AgentControlRunner {
         let signature_validator = config
             .opamp
             .map(|fleet_config| {
-                build_signature_validator(fleet_config.signature_validation, config.proxy)
+                SignatureValidator::new(fleet_config.signature_validation, config.proxy)
             })
             .transpose()?
-            .unwrap_or(SignatureValidator::Noop);
+            .unwrap_or(SignatureValidator::new_noop());
```
sigilioso, line 164: We also need the the Noop thing here because we need to build a signature validator even if OpAMP is disabled, right? 🙃 

Someday we will take care of the whole `let Some(opamp) else` thing...

sigilioso (approved):
🧹 🧹 🧹 

Thanks for taking care of this! If left some relatively unrelated questions and you'll need a rebase, but LGTM 🚀

agent-control/src/opamp/remote_config/validators/signature/validator.rs
gsanchezgavier: yes , it shouldn't happen on any existing flow becase recipe and helm chart sets the corresponding url but just in case

agent-control/src/agent_control/run.rs
gsanchezgavier: yes and in this case the warn message doesn't make sense, it only appears whenver opamp is enabled and sig verificatino is disabled

sigilioso (approved):

PR #1741, opened 2025-10-01
chore(style): follow style guide

paologallinaharbur:
removed default/new implementations when not needed

## Checklist


agent-control/src/secrets_provider.rs
DavSanchez, line 1: No `Default` for this one? It would be an empty `HashMap` in all cases no? Doesn't `#[derive(Default)]` work for the `Registry<S>` type?

DavSanchez (approved):
Thanks! I have only a doubt.

PR #1742, opened 2025-10-02
fix: validate if agent-id collide with release-name

sigilioso:
This PR extends the `DynamicConfigValidator` for K8s in order to assure that the agent identifiers provided in the remote configuration for Agent Control do not collide with the `agent-control-deployment` or `agent-control-cd` release names.

If there was collisions the AC behavior was totally unexpected. For instance, the AC HelmRelease could be replaced by the HelmRelease corresponding to a subagent, leading to completely stopping Agent Control.

Additionally, the corresponding validations in the K8s garbage have been removed because the garbage collector checks the `agent-id` **labeled** in the K8s objects

```yaml
labels:
  app.kubernetes.io/managed-by: agent-control
  newrelic.io/agent-id: some-agent-id
``` 

and [already skips deletion](https://github.com/newrelic/newrelic-agent-control/blob/d983685db72bce6b9b0d0f445041f6f4a1da4ba4/agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs#L145-L146) if the corresponding agent-id is `agent-control` (the only reserved agent-id)

It is easy to check that `agent-control-deploment` and `agent-control-cd` [resources are labeled with agent-control as agent-id](https://github.com/newrelic/newrelic-agent-control/blob/d983685db72bce6b9b0d0f445041f6f4a1da4ba4/agent-control/src/cli/install.rs#L460).

### DIFF
--- a/agent-control/src/agent_control/config_validator.rs
+++ b/agent-control/src/agent_control/config_validator.rs
@@ -3,6 +3,8 @@ use crate::agent_type::agent_type_registry::AgentRegistry;
 use std::sync::Arc;
 use thiserror::Error;
 
+pub mod k8s;
+
 #[derive(Error, Debug)]
 #[error("config validation failed: {0}")]
 pub struct DynamicConfigValidatorError(String);
--- a/agent-control/src/agent_control/config_validator.rs
+++ b/agent-control/src/agent_control/config_validator.rs
@@ -1,13 +1,11 @@
 use crate::agent_control::config::AgentControlDynamicConfig;
-use crate::agent_type::agent_type_registry::{AgentRegistry, AgentRepositoryError};
+use crate::agent_type::agent_type_registry::AgentRegistry;
 use std::sync::Arc;
 use thiserror::Error;
 
 #[derive(Error, Debug)]
-pub enum DynamicConfigValidatorError {
-    #[error("{0}")]
-    AgentRepositoryError(#[from] AgentRepositoryError),
-}
+#[error("config validation failed: {0}")]
+pub struct DynamicConfigValidatorError(String);
 
 /// Represents a validator for dynamic config
 pub trait DynamicConfigValidator {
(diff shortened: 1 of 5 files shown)


agent-control/src/agent_control/config_validator.rs
```diff
@@ -1,13 +1,13 @@
...
 
+pub mod k8s;
+
 #[derive(Error, Debug)]
-pub enum DynamicConfigValidatorError {
-    #[error("{0}")]
-    AgentRepositoryError(#[from] AgentRepositoryError),
-}
+#[error("config validation failed: {0}")]
+pub struct DynamicConfigValidatorError(String);
```
DavSanchez, line 10: I guess a way to not have arbitrary `String`s here would be to have an associated type for `DynamicConfigValidator` that is `Into<DynamicConfigValidatorError>` or something so each implementation can decide how to create the error value. I still find it weird to see `String` as the inner type for an error, even if what we end up doing is logging it 😅

agent-control/src/agent_control/config_validator/k8s.rs
```diff
@@ -0,0 +1,146 @@
...
+            .agents
+            .keys()
+            .try_for_each(|agent_id| match agent_id.as_str() {
+                agent_id if agent_id == self.ac_release_name => {
+                    Err(validation_error(agent_id, "agent-control-deployment"))
+                }
+                agent_id if agent_id == self.cd_release_name => {
+                    Err(validation_error(agent_id, "agent-control-cd"))
+                }
+                _ => Ok(()),
```
DavSanchez, line 28: 👌 some `if-else` chain expression could have worked here but I think this one is shorter!

agent-control/src/agent_control/config_validator/k8s.rs
```diff
@@ -0,0 +1,146 @@
...
+        dynamic_config: &AgentControlDynamicConfig,
+    ) -> Result<(), DynamicConfigValidatorError> {
+        self.inner.validate(dynamic_config)?;
+
+        dynamic_config
+            .agents
+            .keys()
+            .try_for_each(|agent_id| match agent_id.as_str() {
+                agent_id if agent_id == self.ac_release_name => {
+                    Err(validation_error(agent_id, "agent-control-deployment"))
```
DavSanchez, line 23: Shouldn't we be using the actual `self.ac_release_name` instead of `"agent-control-deployment"` and same for CD?

DavSanchez (commented):
Thanks! I only have a doubt regarding the collision errors

agent-control/src/agent_control/config_validator.rs
sigilioso: I also considered the associated type but I opted for this approach in order to give a try to what we agreed recently. I actually don't have a strong opinion.

PR #1779, opened 2025-10-17
feat: add config generator to host cli [NR-475245]

sigilioso:
This PR adds a tool to generate the host configuration to the cli, this normally includes generating a System Identity.

## Implementation details

The system Identity creation can be configured as follows:

```
      --organization-id <ORGANIZATION_ID>
          Organization identifier
      --auth-parent-client-id <AUTH_PARENT_CLIENT_ID>
          Client ID corresponding to the parent system identity (requires `auth_client_secret`)
      --auth-parent-client-secret <AUTH_PARENT_CLIENT_SECRET>
          Client Secret corresponding to the parent system identity (requires `auth_client_id`)
      --auth-parent-token <AUTH_PARENT_TOKEN>
          Auth token corresponding to the parent system identity
      --auth-private-key-path <AUTH_PRIVATE_KEY_PATH>
          When (`auth_token` or `auth_client_id` + `auth_client_secret`) are set, this path is used to store the identity key. Otherwise, the path is expected to contain the already provided private key was already provided
      --auth-client-id <AUTH_CLIENT_ID>
          Client identifier corresponding to an already provisioned identity. No identity creation is performed, therefore setting this up also requires an existing private key pointed in `auth_private_key_path`
```

* **auth-private-key-path** is required and is the path of the private key of the Identity that will be used by Agent Control.
* If **auth-client** is provided, this is the Identity to be used by Agent Control (the corresponding key is expected to be in **auth-private-key-path**).
* If **auth-parent-token** is set, the token will be used to create a new Identity (whose key will be stored in **auth-private-key-path**).
* If **auth-parent-client-id** and **auth-parent-secret**, these arguments are used to first obtain a token and then create a new identity.

Once the identity is obtained, the configuration is generated as configured through the command arguments.

<details>

<summary>Full set of configuration options</summary>

```
❯ ./newrelic-agent-control-onhost-cli generate-config --help

Generates the Agent Control configuration for host environments

Usage: newrelic-agent-control-onhost-cli generate-config [OPTIONS] --output-path <OUTPUT_PATH> --region <REGION> --fleet-id <FLEET_ID> --organization-id <ORGANIZATION_ID> --agent-set <AGENT_SET> --auth-parent-client-id <AUTH_PARENT_CLIENT_ID> --auth-parent-client-secret <AUTH_PARENT_CLIENT_SECRET> --auth-parent-token <AUTH_PARENT_TOKEN> --auth-private-key-path <AUTH_PRIVATE_KEY_PATH> --auth-client-id <AUTH_CLIENT_ID> --proxy-url <URL> --proxy-ca-bundle-dir <CA_BUNDLE_DIR> --proxy-ca-bundle-file <CA_BUNDLE_FILE>

Options:
      --output-path <OUTPUT_PATH>
          Sets where the generated configuration should be written to
      --cli-log-level <CLI_LOG_LEVEL>
          Log level for the cli command [default: info]
      --fleet-enabled
          Defines if Fleet Control is enabled
      --region <REGION>
          New Relic region
      --fleet-id <FLEET_ID>
          Fleet identifier
      --organization-id <ORGANIZATION_ID>
          Organization identifier
      --agent-set <AGENT_SET>
          Set of agents to be used as local configuration [possible values: infra-agent, otel, none]
      --auth-parent-client-id <AUTH_PARENT_CLIENT_ID>
          Client ID corresponding to the parent system identity (requires `auth_client_secret`)
      --auth-parent-client-secret <AUTH_PARENT_CLIENT_SECRET>
          Client Secret corresponding to the parent system identity (requires `auth_client_id`)
      --auth-parent-token <AUTH_PARENT_TOKEN>
          Auth token corresponding to the parent system identity
      --auth-private-key-path <AUTH_PRIVATE_KEY_PATH>
          When (`auth_token` or `auth_client_id` + `auth_client_secret`) are set, this path is used to store the identity key. Otherwise, the path is expected to contain the already provided private key was already provided
      --auth-client-id <AUTH_CLIENT_ID>
          Client identifier corresponding to an already provisioned identity. No identity creation is performed, therefore setting this up also requires an existing private key pointed in `auth_private_key_path`
      --proxy-url <URL>
          Proxy URL proxy: <protocol>://<user>:<password>@<host>:<port> (All parts except host are optional)
      --proxy-ca-bundle-dir <CA_BUNDLE_DIR>
          System path with the CA certificates in PEM format. All `.pem` files in the directory are read
      --proxy-ca-bundle-file <CA_BUNDLE_FILE>
          System path with the CA certificate in PEM format
      --ignore-system-proxy
          When set to true, the HTTPS_PROXY and HTTP_PROXY environment variables are ignored, defaults to false
  -h, --help
          Print help
```

</details>

## Out of scope

- Packaging the new command (it will come in a follow-up PR)

### DIFF
--- a/agent-control/src/cli/on_host/config_gen.rs
+++ b/agent-control/src/cli/on_host/config_gen.rs
@@ -36,7 +36,7 @@ pub struct Args {
     region: Region,
 
     /// Fleet identifier
-    #[arg(long, default_value = "")]
+    #[arg(long, default_value_t)]
     fleet_id: String,
 
     /// Set of agents to be used as local configuration.
@@ -215,13 +215,13 @@ mod tests {
         || String::from("--fleet-disabled --output-path /some/path --agent-set otel --region us")
     )]
     #[case::identity_already_provided(
-        || format!("--output-path /some/path --agent-set otel --region us --auth-private-key-path {} --auth-client-id some-client-id", pwd())
+        || format!("--output-path /some/path --agent-set otel --region us --fleet-id some-id --auth-private-key-path {} --auth-client-id some-client-id", pwd())
     )]
     #[case::token_based_identity(
-        || format!("--output-path /some/path --agent-set otel --region us --auth-private-key-path {} --auth-parent-token TOKEN --auth-parent-client-id id --organization-id org-id", pwd())
+        || format!("--output-path /some/path --agent-set otel --region us --fleet-id some-id --auth-private-key-path {} --auth-parent-token TOKEN --auth-parent-client-id id --organization-id org-id", pwd())
     )]
     #[case::client_id_and_secret_based_identity(
-        || format!("--output-path /some/path --agent-set otel --region us --auth-private-key-path {} --auth-parent-client-secret SECRET --auth-parent-client-id id --organization-id org-id", pwd())
+        || format!("--output-path /some/path --agent-set otel --region us --fleet-id some-id --auth-private-key-path {} --auth-parent-client-secret SECRET --auth-parent-client-id id --organization-id org-id", pwd())
     )]
     fn test_args_validation(#[case] args: fn() -> String) {
         let cmd = Args::command().no_binary_name(true);
--- a/agent-control/tests/on_host/agent_control_cli.rs
+++ b/agent-control/tests/on_host/agent_control_cli.rs
@@ -1,0 +1,78 @@
+//! Integration tests for newrelic-agent-control-onhost-cli
+
+use assert_cmd::Command;
+use tempfile::TempDir;
+
+#[test]
+fn test_config_generator_fleet_disabled_proxy() {
+    let tmp = TempDir::new().unwrap();
+    let path = tmp.path().join("output.yaml").to_string_lossy().to_string();
+
+    let mut cmd = Command::cargo_bin("newrelic-agent-control-onhost-cli").unwrap();
+    let args = format!(
+        "generate-config --fleet-disabled --agent-set infra-agent --region us --proxy-url https://some.proxy.url/ --proxy-ca-bundle-dir /test/bundle/dir --proxy-ca-bundle-file /test/bundle/file --ignore-system-proxy --output-path {path}",
+    );
+    cmd.args(args.split(" "));
+    cmd.assert().success();
+
+    let expected_value: serde_yaml::Value = serde_yaml::from_str(
+        r#"
+server:
+  enabled: true
+agents:
+  nr-infra:
+    agent_type: "newrelic/com.newrelic.infrastructure:0.1.0"
+proxy:
+  url: https://some.proxy.url/
+  ca_bundle_dir: /test/bundle/dir
+  ca_bundle_file: /test/bundle/file
+  ignore_system_proxy: true
+    "#,
+    )
+    .unwrap();
+    let actual_content = std::fs::read_to_string(&path).unwrap();
+    let actual_value: serde_yaml::Value = serde_yaml::from_str(&actual_content).unwrap();
+    assert_eq!(actual_value, expected_value);
+}
+
+#[test]
+fn test_config_generator_fleet_enabled_identity_provisioned() {
+    let tmp = TempDir::new().unwrap();
+    let path = tmp.path().join("output.yaml").to_string_lossy().to_string();
+    let key_path = tmp.path().join("key");
+    std::fs::write(&key_path, "fake-key").unwrap();
+    let key_path = key_path.to_string_lossy().to_string();
+
+    let mut cmd = Command::cargo_bin("newrelic-agent-control-onhost-cli").unwrap();
+    let args = format!(
+        "generate-config --agent-set infra-agent --region us --fleet-id FLEET-ID --auth-client-id CLIENT-ID --auth-private-key-path {key_path} --output-path {path}",
+    );
+    cmd.args(args.split(" "));
+    cmd.assert().success();
+
+    let expected_value: serde_yaml::Value = serde_yaml::from_str(
+        &format!(r#"
+fleet_control:
+  endpoint: https://opamp.service.newrelic.com/v1/opamp
+  signature_validation:
+    public_key_server_url: https://publickeys.newrelic.com/r/blob-management/global/agentconfiguration/jwks.json
+  fleet_id: FLEET-ID
+  auth_config:
+    token_url: https://system-identity-oauth.service.newrelic.com/oauth2/token
+    client_id: CLIENT-ID
+    provider: local
+    private_key_path: {key_path}
+
+server:
+  enabled: true
+
+agents:
+  nr-infra:
+    agent_type: "newrelic/com.newrelic.infrastructure:0.1.0"
+    "#,
+        ))
+    .unwrap();
+    let actual_content = std::fs::read_to_string(&path).unwrap();
+    let actual_value: serde_yaml::Value = serde_yaml::from_str(&actual_content).unwrap();
+    assert_eq!(actual_value, expected_value);
+}
--- a/agent-control/tests/on_host/mod.rs
+++ b/agent-control/tests/on_host/mod.rs
@@ -1,3 +1,4 @@
+mod agent_control_cli;
 mod cli;
 mod command;
 mod config_repository;


sigilioso:
Depending on the other arguments it will be generated or it is assumed to exist. We can use different arguments but both would be representing the private-key path. What do you think?

agent-control/src/cli/on_host/config_gen.rs
```diff
@@ -1,14 +1,311 @@
...
+            },
+            fleet_id: args.fleet_id.to_string(),
+            auth_config: AuthConfig {
+                token_url: args.region.token_renewal_endpoint().to_string(),
+                client_id,
+                provider: "local".to_string(),
+                private_key_path: private_key_path.to_string_lossy().to_string(),
+            },
+        })
+    };
```
DavSanchez: Can be an usage for [`bool::then`](https://doc.rust-lang.org/std/primitive.bool.html#method.then):

```suggestion
    let fleet_control = args.fleet_enabled.then(|| {
        let Identity {
            client_id,
            private_key_path,
        } = provide_identity_fn(args)?;

        FleetControl {
            endpoint: args.region.opamp_endpoint().to_string(),
            signature_validation: SignatureValidation {
                public_key_server_url: args.region.public_key_endpoint().to_string(),
            },
            fleet_id: args.fleet_id.to_string(),
            auth_config: AuthConfig {
                token_url: args.region.token_renewal_endpoint().to_string(),
                client_id,
                provider: "local".to_string(),
                private_key_path: private_key_path.to_string_lossy().to_string(),
            },
        }
    });
```

agent-control/src/cli/on_host/config_gen.rs
```diff
@@ -1,14 +1,311 @@
...
+
+    let config = Config {
+        fleet_control,
+        server: Server { enabled: true },
+        proxy: args.proxy_config.clone(),
+        agents: args.agent_set.into(),
+    };
+
+    serde_yaml::to_string(&config)
```
DavSanchez: Total nit but if we are not using the `String` for anything else than writing the file I'd serialize it to binary data `&[u8]` instead.

agent-control/src/cli/on_host/config_gen.rs
```diff
@@ -1,14 +1,311 @@
...
+};
+
+pub mod config;
+pub mod identity;
+pub mod region;
+
+/// Generates the Agent Control configuration for host environments.
 #[derive(Debug, clap::Parser)]
-pub struct ConfigInputs {
+pub struct Args {
```
DavSanchez: I think there's some way to make arguments mutually exclusive to guarantee having either (parent ID + secret) or auth client ID but don't know for sure, did you check?

agent-control/src/cli/on_host/config_gen/config.rs
```diff
@@ -0,0 +1,112 @@
...
+            )]
+            .into(),
+            AgentSet::None => HashMap::new(),
+        }
+    }
+}
+
+/// Configuration to be written as result of the corresponding command.
+#[derive(Debug, PartialEq, Serialize)]
+pub struct Config {
```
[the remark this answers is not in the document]
sigilioso: I had the very same though but I opted for a new type because of the visibility thing and also to avoid a cascade `Serialize` implementation in all types used by `AgentControlConfig`.

There is a [check in the unit tests of the config generator](https://github.com/newrelic/newrelic-agent-control/pull/1779/files#diff-adb90468a9f7e5edea98eea5902442017bb0e3333819b8c169da306aa61390f0R169-R171) to prevent issues if types drift apart. Do you thing it is enough?

agent-control/src/cli/on_host/config_gen.rs
sigilioso: Considering that `provide_identity_fn(args)?` is inside the closure (note the `?`), it would also involve some _transposing_ to deal with the error, wouldn't it? Isn't it simpler this way?

agent-control/src/cli/on_host/config_gen.rs
sigilioso: Thanks for pointing it out! I though about it and then I totally forgot 😅 I'll dig into it!

agent-control/src/cli/on_host/config_gen/config.rs
```diff
@@ -0,0 +1,112 @@
...
+
+use crate::http::config::ProxyConfig;
+
+/// Represents the set of agents to be included in the AC configuration.
+#[derive(Debug, Copy, Clone, PartialEq, clap::ValueEnum)]
+pub enum AgentSet {
+    InfraAgent,
+    Otel,
+    None,
+}
```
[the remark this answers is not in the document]
DavSanchez: I feel more aligned to leveraging `Option` as I think its API allows for more expressive manipulation but no strong opinion 👌.

What I'd really consider then is to perhaps add a `Default` implementation that selects the `NoAgent` variant and also rename the type to something different than `AgentSet` (as it's doesn't represent a collection but an element of some later collection).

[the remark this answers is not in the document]
danielorihuela: Another alternative

```
type AgentSet = HashMap<String, Agent>
enum AgentSetType { ... }
```

agent-control/src/cli/on_host/config_gen/identity.rs
```diff
@@ -0,0 +1,338 @@
...
+/// Provides a key-based identity considering the supplied args.
+pub fn provide_identity(args: &Args) -> Result<Identity, CliError> {
+    let environment = NewRelicEnvironment::from(args.region);
+    build_identity(args, environment)
+}
+
+/// Helper to allow injecting testing urls when building the identity.
+fn build_identity(args: &Args, environment: NewRelicEnvironment) -> Result<Identity, CliError> {
+    // Already provided identity
+    if !args.auth_client_id.is_empty() {
```
danielorihuela: Should we check that the private key path is not empty. Here or in "clap". Maybe in both places.

agent-control/src/cli/on_host/config_gen/identity.rs
```diff
@@ -0,0 +1,338 @@
...
+        return Ok(Identity {
```
danielorihuela: Maybe a log would be nice. It's hard for me to know what it's too much and what it's too little. I see you added `info!("Generating System Identity");`. Maybe a log in this other path would be nice.

```
info!(Using passed identity);
```

sigilioso: I wanted to keep logs simple but I probably went too far! I'm adding some details about how the identity is generated.

agent-control/src/cli/on_host/config_gen/region.rs
```diff
@@ -0,0 +1,181 @@
...
+        let matches = cmd.try_get_matches_from(["cmd", "--flag", input]).unwrap();
+        let value = matches.get_one::<Region>("flag").cloned();
+        assert_eq!(value, Some(expected));
+    }
+
+    #[rstest]
+    #[case("invalid")]
+    #[case("")]
+    //#[case("invalid", Err("invalid is not a supported region".to_string()))]
+    //#[case("", Err(" is not a supported region".to_string()))]
```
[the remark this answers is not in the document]
sigilioso: Yes! Thanks!

build/examples/agent-control-config-nr-otel-collector.yaml
```diff
@@ -1,15 +1,15 @@
...
-#     public_key_server_url: https://publickeys.newrelic.com/r/blob-management/global/agentconfiguration/jwks.json
-#   headers:
-#     api-key: API_KEY_HERE
-#   fleet_id: FLEET_ID_HERE
-#   auth_config:
-#     token_url: PLACEHOLDER
-#     client_id: PLACEHOLDER
-#     provider: PLACEHOLDER
-#     private_key_path: PLACEHOLDER
+fleet_control:
```
danielorihuela: Do we still need the examples after adding the cli? Should we remove them?
Why did you only uncomment stuff in this example, but not the others?

agent-control/src/cli/on_host/config_gen/config.rs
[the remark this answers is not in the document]
sigilioso: If we used this last alternative we would need different arguments, wouldn't we? `--agents infra-agent` wouldn't be enough 🤔

build/examples/agent-control-config-nr-otel-collector.yaml
sigilioso: Great catch! The example should be untouched for now.

The examples will be removed but they can't  be removed until the recipe is updated. I was thinking following these steps:

1. Finish the implementation (this PR)
2. Include the new cli in the package (along with nr-auth cli and the examples)
3. Release it.
4. Update the recipe to use the new cli (and stop 
5. Clean everything up: stop packaging the nr-auth cli and remove the examples.

agent-control/src/cli/on_host/config_gen/config.rs
```diff
@@ -0,0 +1,112 @@
...
+
+impl From<AgentSet> for HashMap<String, Agent> {
+    fn from(value: AgentSet) -> Self {
+        match value {
+            AgentSet::InfraAgent => [(
+                "nr-infra".to_string(),
```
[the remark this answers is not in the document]
DavSanchez: > Why is it out of scope? Wouldn't it be easy to have an optional cli parameter with a default value? That way we have `nr-infra` hardcoded but we can modify it rapidly. That would avoid a re-release I think, if we call this directly from the recipe.

I think it's out of scope because it's kind of an overarching issue. IIRC FC sends us some default names for agents as well given AC config is not explicit in the UI anymore, so it's not a matter of e.g. making the names here customizable or anything but to know what exactly we want to do with these keys.

agent-control/src/cli/on_host/config_gen/config.rs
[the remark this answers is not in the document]
sigilioso: It is currently covered in unit tests, instead of having an explicit conversion it uses the same conversion that will be used in the expected scenarios: `Cli config --> yaml --> AC config`.

agent-control/src/cli/on_host/config_gen/identity.rs
sigilioso: I added more logs here: [1f4d4da](https://github.com/newrelic/newrelic-agent-control/pull/1779/commits/1f4d4dac5e9342324e519238fb64e6b9dcaad1a8)

PR #1782, opened 2025-10-17
feat: onhost supervisor remove condvar approach

danielorihuela:

Reworks the process management for onhost supervisors.

## Which issue this PR fixes

- fixes #NR-476366

## Checklist

### DIFF
--- a/agent-control/src/sub_agent/on_host/command/command_os.rs
+++ b/agent-control/src/sub_agent/on_host/command/command_os.rs
@@ -84,7 +84,7 @@ impl CommandOSStarted {
         self.process.id()
     }
 
-    pub(crate) fn is_running(&mut self) -> bool {
+    pub fn is_running(&mut self) -> bool {
         self.process.try_wait().is_ok_and(|v| v.is_none())
     }
 
--- a/agent-control/tests/on_host/command/shutdown.rs
+++ b/agent-control/tests/on_host/command/shutdown.rs
@@ -45,7 +45,7 @@ fn command_shutdown_when_sigterm_is_ignored() {
             args: Default::default(),
             env: Default::default(),
             restart_policy: Default::default(),
-            shutdown_timeout: Duration::from_secs(1),
+            shutdown_timeout: Duration::from_millis(100),
         },
         false,
         Default::default(),
@@ -55,5 +55,7 @@ fn command_shutdown_when_sigterm_is_ignored() {
     sleep(Duration::from_millis(500)); // Give the process some room to start
 
     let terminated = cmd.shutdown();
-    assert!(terminated.is_ok())
+    sleep(Duration::from_millis(200));
+    assert!(!cmd.is_running());
+    assert!(terminated.is_ok());
 }


agent-control/src/context.rs
```diff
@@ -1,77 +0,0 @@
-use std::sync::{Arc, Condvar, Mutex, MutexGuard, PoisonError};
```
paologallinaharbur: 🥲 finally

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -208,159 +193,171 @@ impl NotStartedSupervisorOnHost {
         )
     }
 
-    fn start_process_thread(
+    fn start_process_threads(
```
paologallinaharbur: this method is something like 200 lines 😕 
Could we "break it down"? I feel like it is doing way too much

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -208,159 +193,171 @@ impl NotStartedSupervisorOnHost {
...
+                        _ = ProcessTerminator::new(pid)
+                            .shutdown(|| process_finished_consumer.is_cancelled(Duration::new(10, 0)));
+                    } else {
+                        info!(msg = "Executable not running");
+                    }
+                },
+                recv(process_error_consumer.as_ref()) -> _ => info!(msg = "Executable not running"),
+            }
+            span.exit();
+        };
```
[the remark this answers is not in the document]
paologallinaharbur: Moreover, if we can stop this thread via `recv(stop_consumer.as_ref())` why can't we stop also the other?

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -87,24 +78,20 @@ impl SupervisorStarter for NotStartedSupervisorOnHost {
             .chain(thread_contexts)
             .collect();
 
-        Ok(StartedSupervisorOnHost {
-            ctx,
-            thread_contexts,
-        })
+        Ok(StartedSupervisorOnHost { thread_contexts })
```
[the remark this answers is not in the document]
paologallinaharbur: IMO if we go for this approach we should not mix the two kinds of threads, and it should be instead something like 
```
thread_contexts: Vec<(StartedThreadContext, StartedThreadContext)>,
```
to differentiate the pourpose of the two

agent-control/src/event/cancellation.rs
```diff
@@ -5,15 +5,23 @@ use std::time::Duration;
...
+            Ok(_) => true,
+            Err(RecvTimeoutError::Timeout) => false,
+            Err(RecvTimeoutError::Disconnected) => true,
+        }
+    }
+
+    /// Checks whether the consumer is cancelled immediately.
+    ///
+    /// Calls [`Self::is_cancelled`] with a timeout of zero.
+    pub fn is_cancelled_immediately(&self) -> bool {
```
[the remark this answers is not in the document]
danielorihuela: https://github.com/newrelic/newrelic-agent-control/pull/1782/commits/331b53a16141725369efce301833db5ffdcc4f33

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -208,159 +193,171 @@ impl NotStartedSupervisorOnHost {
...
-                wait_exit_timeout(ctx.clone(), duration);
-            });
+                span.exit();
+            }
         };
 
-        NotStartedThreadContext::new(executable_data.bin.clone(), callback).start()
+        vec![
+            NotStartedThreadContext::new(executable_data.bin.clone(), terminator_callback).start(),
+            NotStartedThreadContext::new(executable_data.bin.clone(), executor_callback).start(),
```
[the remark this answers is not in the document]
paologallinaharbur: I think we can either return a tuple or even better avoid the second thread. I still have doubts regarding really needed it

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -212,152 +208,141 @@ impl NotStartedSupervisorOnHost {
...
-                    "Error publishing health status for {}: {}",
-                    executable_data_clone.id, err
+                let (process_finished_publisher, process_finished_consumer) = pub_sub();
+                let command_result = start_command(
+                    not_started_command,
+                    processes.clone(),
+                    process_finished_consumer,
                 );
-            }
+                span.exit();
```
DavSanchez: I remember spans causing resource leakage. This one will be closed right after exiting a process which could potentially have run for a long time. I guess it was also like this before, but if not it would be great to have some memory profiling.

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -212,152 +208,141 @@ impl NotStartedSupervisorOnHost {
...
+                }
 
-            // A context cancelled means that the supervisor has been gracefully stopped and is the
-            // most probably reason why process has been exited.
-            if *Context::get_lock_cvar(&ctx).0.lock().unwrap() {
                 info!(
-                    supervisor = executable_data_clone.bin,
-                    msg = "supervisor has been stopped and process terminated"
+                    "Restarting supervisor ({}/{})",
+                    i + 1,
```
DavSanchez: I know this is a very stupid comment, but I'd prefer to have this counter logged after it has been actually incremented, to save us the typical troubleshooting of the numbers not matching until we notice it was not updated or updated two times or something like that xD

agent-control/src/event/cancellation.rs
```diff
@@ -5,15 +5,23 @@ use std::time::Duration;
...
-        let timed_out = matches!(
-            self.as_ref().recv_timeout(timeout),
-            Err(RecvTimeoutError::Timeout)
-        );
-        !timed_out
+    pub fn is_cancelled_with_timeout(&self, timeout: Duration) -> bool {
+        match self.as_ref().recv_timeout(timeout) {
+            Ok(_) => true,
+            Err(RecvTimeoutError::Timeout) => false,
+            Err(RecvTimeoutError::Disconnected) => true,
```
DavSanchez: I think this might work

```suggestion
        match self.as_ref().recv_timeout(timeout) {
            Ok(_) | Err(RecvTimeoutError::Disconnected) => true,
            Err(RecvTimeoutError::Timeout) => false,
            
```

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -212,152 +208,141 @@ impl NotStartedSupervisorOnHost {
...
-
-            // check if restart policy needs to be applied
-            // As the exit code comes inside a Result but we don't care about the Err,
-            // we just unwrap or take the default value (0)
-            if !restart_policy.should_retry(exit_code.unwrap_or_default()) {
-                // Log if we are not restarting anymore due to the restart policy being broken
-                warn!("supervisor won't restart anymore due to having exceeded its restart policy");
+                let mut should_stop = false;
+                restart_policy.backoff(|duration| {
+                    // early exit if supervisor timeout is canceled
```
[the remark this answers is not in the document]
danielorihuela: Not sure we can reduce them.

The first exit makes sure the command is not executed if the stop signal is received before. I assume we want that. Otherwise, the command will always execute at least once. The user may not want this, because it could modify the system.

The second exit is required to avoid reaching the line of code were we execute the backoff. The stop signal was already consumed to kill the command. Hence, if we don't exit before reaching the backoff, we will 

The third exit is required to abort the restart.

I'll try to simplify it though.

agent-control/src/sub_agent/on_host/supervisor.rs
danielorihuela: Spans where there, but they were not working properly when I checked the tests.
They were not being added to the logs for some reason.
So... we could have the problem now yes. Anyways, the logs from the process won't have the span because those logs are "created" somewhere else.

In the next snippet we can see that behaviour.

```
2025-10-21T13:41:34.226637Z DEBUG newrelic_agent_control::sub_agent::on_host::command::logging::logger: NR-command agent_id=wrong-command
2025-10-21T13:41:34.226668Z  INFO executable process{id=wrong-command exec_id="echo"}: newrelic_agent_control::sub_agent::on_host::supervisor: msg="Executable not running"
2025-10-21T13:41:34.226703Z  INFO executable process{id=wrong-command exec_id="echo"}: newrelic_agent_control::sub_agent::on_host::supervisor: Restarting supervisor (1/3)
2025-10-21T13:41:34.226748Z  INFO executable process{id=wrong-command exec_id="echo"}: newrelic_agent_control::sub_agent::on_host::supervisor: Starting executable
2025-10-21T13:41:34.226760Z DEBUG executable process{id=wrong-command exec_id="echo"}: newrelic_agent_control::sub_agent::on_host::supervisor: Informing executable as healthy
2025-10-21T13:41:34.484664Z DEBUG newrelic_agent_control::sub_agent::on_host::command::logging::logger: NR-command agent_id=wrong-command
```

I think we should be okay. It could be a problem with huge "restart" number.

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -427,53 +409,27 @@ fn compute_exit_code(exit_status: ExitStatus) -> i32 {
...
-    *pid = Some(streaming.get_pid());
-    // free the lock so the wait_for_termination can lock it on graceful shutdown
-    drop(pid);
-
-    drop(span_guard);
+    let process_pid = streaming.get_pid();
+    processes
+        .lock()
+        .unwrap()
+        .insert(process_pid, process_finished_consumer);
```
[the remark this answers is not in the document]
danielorihuela: This doesn't happen with the new approach that Paolo suggested. I'll keep this comment unresolved here in case we decide the current approach is prefered.

agent-control/src/sub_agent/on_host/command/command_os.rs
```diff
@@ -75,14 +76,47 @@ impl CommandOSNotStarted {
...
+    pub fn shutdown(&mut self) -> Result<(), CommandError> {
+        let pid = self.get_pid() as i32;
+
+        use nix::{sys::signal, unistd::Pid};
+        signal::kill(Pid::from_raw(pid), signal::SIGTERM)
+            .map_err(|err| CommandError::NixError(err.to_string()))?;
+
+        let mut callback = || {
+            let timeout = Duration::from_secs(10); // 10s total
+            let poll_interval = Duration::from_millis(10); // Check every 10ms
```
paologallinaharbur: I believe we could also check each 100 millis or even more

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -214,150 +194,123 @@ impl NotStartedSupervisorOnHost {
...
+                            info!(supervisor = bin, msg = "Executable terminated");
+                            should_stop = true;
+                        }
                     }
-                })
-                .map(|exit_status| {
-                    handle_termination(
-                        &executable_data_clone,
+
+                    // At this point, the command is already dead. However, we call `wait` to
```
paologallinaharbur: nice and useful comment!

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -208,159 +193,171 @@ impl NotStartedSupervisorOnHost {
...
+                    logging_path.clone(),
+                );
 
-            let supervisor_start_time = SystemTime::now();
+                let supervisor_start_time = SystemTime::now();
+                let bin = executable_data_clone.bin.clone();
 
-            let init_health = Healthy::new();
+                // TODO: when the executable fails, and max-retries are not configured in the backoff policy, this
+                // can lead to false positives (reporting healthy when the executable is actually not working)
```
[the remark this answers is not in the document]
paologallinaharbur: by the way I do not remember why we are reporting healthy BEFORE the binary starts 😅  @sigilioso do you have any context on this?

agent-control/src/sub_agent/on_host/command/command_os.rs
```diff
@@ -75,14 +76,47 @@ impl CommandOSNotStarted {
...
     pub fn get_pid(&self) -> u32 {
         self.process.id()
     }
 
+    pub fn shutdown(&mut self) -> Result<(), CommandError> {
+        let pid = self.get_pid() as i32;
+
+        use nix::{sys::signal, unistd::Pid};
+        signal::kill(Pid::from_raw(pid), signal::SIGTERM)
+            .map_err(|err| CommandError::NixError(err.to_string()))?;
```
sigilioso: Do you know if `self.process.kill()` would work in the same way? I was wondering if using it would easy the future Windows support.

agent-control/src/sub_agent/on_host/command/command_os.rs
```diff
@@ -75,14 +76,47 @@ impl CommandOSNotStarted {
...
+
+        if self.is_running_after_timeout(Duration::from_secs(10)) {
+            signal::kill(Pid::from_raw(pid), signal::SIGKILL)
+                .map_err(|err| CommandError::NixError(err.to_string()))?;
+        }
+        Ok(())
```
sigilioso: After the timeout we send the kill signal again, but we actually don't know if the process is actually stopped. Is it OK to return `Ok(())`?

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -214,150 +194,123 @@ impl NotStartedSupervisorOnHost {
...
+                            .inspect_err(|err| error!("Publishing health status: {err}"));
+
+                        0 // Default exit code
+                    }
+                };
+
+                info!(msg = "Executable not running");
+
+                // check if restart policy needs to be applied
+                if !restart_policy.should_retry(exit_code) {
```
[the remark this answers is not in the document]
danielorihuela: I double checked tests, and it's not used. Which made me think... If the the binary was successful, we don't want to restart it, right? The backoff is only for when it went wrong.

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -214,150 +194,123 @@ impl NotStartedSupervisorOnHost {
...
+                    // Listen for the cancellation signal while the command is running.
+                    // Upon receiving the signal, we kill the process. That way, we can ensure
+                    // that the thread stops in time.
+                    while command.is_running() {
+                        if stop_consumer.is_cancelled() {
+                            info!(supervisor = bin, "Stopping executable");
+                            let _ = command.shutdown();
+                            info!(supervisor = bin, msg = "Executable terminated");
+                            should_stop = true;
+                        }
```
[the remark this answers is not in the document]
sigilioso: We already had some functions such `handle_termination` (it has many arguments but it still helps), I wonder if we could add some others. If having too many arguments is an issue we could also consider a helper struct for common arguments.

agent-control/src/sub_agent/on_host/supervisor.rs
DavSanchez: That's enough for me!

agent-control/src/sub_agent/on_host/command/command_os.rs
danielorihuela: We can discuss it tomorrow. It's true that we are being very aggressive, but to be honest, I don't know how much of this is relevant in real use.

danielorihuela: After researching for a bit and talking with an AI, I decided to put the linux implementation we had back. Now we have both a linux and windows implementation. https://github.com/newrelic/newrelic-agent-control/pull/1782/commits/9c3d85311708c93410550e6b7c1ffb399779dc9a.

I added a TODO to properly design the unix/windows shutdown once we implement windows. I expect this to happen shortly. In https://new-relic.atlassian.net/jira/software/c/projects/NR/boards/61?quickFilter=7808&selectedIssue=NR-476370.

danielorihuela:
I checked the CPU consumption locally. This are the results.

Captura de pantalla 2025-10-24 a las 11 01 46
Captura de pantalla 2025-10-24 a las 11 03 06

0.0% of CPU usage in the different threads. I don't know how to identify the specific "process thread", but seems like they all consume low CPU. We should we okay to merge that. If anything slipped, we will notice in the canaries.

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -214,244 +199,222 @@ impl NotStartedSupervisorOnHost {
...
-        error!(
-            %agent_id,
-            supervisor = bin,
-            exit_code = ?exit_status.code(),
-            "supervisor process exited unsuccessfully"
-        )
+        // Publish healthy status once after the process has been running
+        // for an arbitrary long time without issues.
+        if !already_published && Instant::now() > deadline {
+            debug!("Informing executable as healthy");
```
sigilioso: Nit: maybe adding some information regarding `healthy_publish_delay` could be useful in the debug message.

danielorihuela: done

PR #1792, opened 2025-10-22
fix(config_migrate): logs were disabled

DavSanchez:

We use a filtering mechanism in which we disabled all logs not coming from `newrelic_agent_control` crate, which excluded `newrelic_config_migrate`

the filters used make it tricky to configure logging properly for `config_migrate`. For example, `debug` log level cannot be activated, and previously the logs would be emitted depending on if the 'callsite' is inside the AC crate or not.

the error handling of `config_migrate` needs to be improved in any case.

## Which issue this PR fixes

The logs seemed unpredictable, sometimes they are seen and sometimes they're not, depending if some error was found when migrating.

## Checklist

### DIFF
--- a/agent-control/src/instrumentation/config/logs/config.rs
+++ b/agent-control/src/instrumentation/config/logs/config.rs
@@ -206,7 +206,7 @@ mod tests {
             TestCase {
                 name: "everything default",
                 config: Default::default(),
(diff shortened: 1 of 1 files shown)


sigilioso:
Let's also check the config for on-host and k8s `cli`, in case there are issues there as well.

PR #1794, opened 2025-10-22
test: local repo onhost e2e

gsanchezgavier:
Introduce the following changes in the onhost e2e:
- The installation of AC is performed with the recipe
- A specific branch of the recipe can be set for the repo
- Packages are installed from a local created repo in the host using the built .deb package from the current branch.
- E2E now can be triggered on each PR without the publish package dance.

Pending:
- remove temporal branch: @sigilioso perhaps we could merge it with this and then change it whenever is the recipe pr is merged?

### DIFF
--- a/test/ac-ansible-roles/install_ac_recipe/tasks/debian.yaml
+++ b/test/ac-ansible-roles/install_ac_recipe/tasks/debian.yaml
@@ -4,7 +4,11 @@
     state: directory
   register: tmp_repo
   when: local_package_path is defined
-# Local repository https://askubuntu.com/questions/170348/how-to-create-a-local-apt-repository
+# Adds a local folder to the trusted repo list. The folder contains the local .deb packages that will be 
+# scanned and added to the repo (building the required metadata). After that is done these packages are 
+# available to installed with apt. 
+# The recipe is still adding the apt upstream production repo so both interoperates, and because of that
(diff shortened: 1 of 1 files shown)


sigilioso:
> remove temporal branch: @sigilioso perhaps we could merge it with this and then change it whenever is the recipe pr is merged?

I think it makes sense, besides the PR is ready to review and we should merge soon.

test/ac-ansible-roles/install_ac_recipe/tasks/debian.yaml
```diff
@@ -0,0 +1,67 @@
...
+    if [ -z "$(ls -A "{{ tmp_repo.path }}")" ]; then
+      echo "No packages where found"
+      exit 1
+    fi
+    
+    cd {{ tmp_repo.path }}
+    dpkg-scanpackages -m . > Packages
+    
+    apt-get update
+  when: local_package_path is defined
```
sigilioso: This is nice! At first I din't understand how we were _telling to the recipe_ that it needs to use this repository but I guess we are not: since the version we are trying to install is **not** published in the repository that the recipe is adding, we are using this to install the package.

Would it make sense to document it somehow? I fear we face weird issues if we ever change the version name we are generating here and it collided with an already published package.

PR #1805, opened 2025-10-28
feat(on-host-cli): support empty proxy on config generation

sigilioso:
The changes extend the support of the `generate-config` on-hots command for allowing empty values. Basically we can move from:

```
    # Proxy configuration
          if [ ! -z "{{.NEW_RELIC_AGENT_CONTROL_PROXY_URL}}" ] ; then
            AC_PROXY_ARGS="--proxy-url {{.NEW_RELIC_AGENT_CONTROL_PROXY_URL}}"
            if [ ! -z "{{.NEW_RELIC_AGENT_CONTROL_PROXY_CA_BUNDLE_FILE}}" ] ; then
              AC_PROXY_ARGS="$AC_PROXY_ARGS --proxy-ca-bundle-file {{.NEW_RELIC_AGENT_CONTROL_PROXY_CA_BUNDLE_FILE}}"
            fi
            if [ ! -z "{{.NEW_RELIC_AGENT_CONTROL_PROXY_CA_BUNDLE_DIR}}" ] ; then
              AC_PROXY_ARGS="$AC_PROXY_ARGS --proxy-ca-bundle-dir {{.NEW_RELIC_AGENT_CONTROL_PROXY_CA_BUNDLE_DIR}}"
            fi
            if [ "{{.NEW_RELIC_AGENT_CONTROL_PROXY_IGNORE_SYSTEM}}" == "true" ] ; then
              AC_PROXY_ARGS="$AC_PROXY_ARGS --ignore-system-proxy"
            fi
          else
            AC_PROXY_ARGS=""
          fi
          # Config generation
          newrelic-agent-control-cli generate-config \
            <OTHER-ARGS> \
            $AC_PROXY_ARGS
```

to:

```
newrelic-agent-control-cli generate-config \
  <OTHER-ARGS> \
  --proxy-url "{{.NEW_RELIC_AGENT_CONTROL_PROXY_URL}}" \
  --proxy-ca-bundle-file "{{.NEW_RELIC_AGENT_CONTROL_PROXY_CA_BUNDLE_FILE}}" \
  --proxy-ca-bundle-dir "{{.NEW_RELIC_AGENT_CONTROL_PROXY_CA_BUNDLE_DIR}}" \
  --ignore-system-proxy "{{.NEW_RELIC_AGENT_CONTROL_PROXY_IGNORE_SYSTEM}}" 
```

### DIFF
--- a/agent-control/src/cli/on_host/config_gen/config.rs
+++ b/agent-control/src/cli/on_host/config_gen/config.rs
@@ -1,10 +1,8 @@
 //! Contains the definition of the configuration to be generated
 
-use std::collections::HashMap;
+use std::{collections::HashMap, convert::Infallible};
 
 use serde::Serialize;
-
-use crate::http::config::ProxyConfig;
 
 /// Represents the set of agents to be included in the AC configuration.
 #[derive(Debug, Copy, Clone, PartialEq, clap::ValueEnum)]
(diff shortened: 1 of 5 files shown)


paologallinaharbur:
With an env var you could do something like:
```
> echo ${env:+prefix_$env}
prefix_test
```

agent-control/src/cli/on_host/config_gen/config.rs
```diff
@@ -36,6 +34,46 @@ impl From<AgentSet> for HashMap<String, Agent> {
...
+    pub proxy_ca_bundle_file: Option<String>,
+
+    #[arg(long, default_value_t = false, value_parser = ignore_system_proxy_parser, action = clap::ArgAction::Set)]
+    pub ignore_system_proxy: bool,
+}
+
+// Helper to avoid serializing empty values
+fn is_none_or_empty_string(v: &Option<String>) -> bool {
+    v.as_ref().map(|s| s.is_empty()).unwrap_or(true)
+}
```
alvarocabanas: i understand this is to set it as None if the string is empty, isn't it? If it's not present, I guess it's None and you would not need this helper, am I correct?

sigilioso:
> I am a bit scared that we could end-up hiding an issue with this approach, could the variable instead contain both the flag and the value?
> 
> the command could become:
> 
> ```shell
> newrelic-agent-control-cli generate-config \
>   <OTHER-ARGS> \
>   "{{.NEW_RELIC_AGENT_CONTROL_PROXY_URL}}" \
>   "{{.NEW_RELIC_AGENT_CONTROL_PROXY_CA_BUNDLE_FILE}}" \
>   "{{.NEW_RELIC_AGENT_CONTROL_PROXY_CA_BUNDLE_DIR}}" \
>   "{{.NEW_RELIC_AGENT_CONTROL_PROXY_IGNORE_SYSTEM}}"
> 
> and we would not need to have empty defaults.
> ```

Well, tha's pretty similar to the current approach where the args are compounded in the recipe. I'm fine with it but then we need to keep that _logic_ in the recipe and/or guided install.

agent-control/src/cli/on_host/config_gen/config.rs
```diff
@@ -36,6 +34,46 @@ impl From<AgentSet> for HashMap<String, Agent> {
...
+
+// Custom parser to allow empty values as false booleans
+fn ignore_system_proxy_parser(s: &str) -> Result<bool, Infallible> {
+    match s.to_lowercase().as_str() {
+        "" | "false" => Ok(false),
+        _ => Ok(true),
+    }
+}
```
DavSanchez: 👌 using `Infallible`.

I guess this is because `value_parser` is required to return a `Result`?

PR #1843, opened 2025-11-06
feat: minimal `fs` support for Windows

DavSanchez:

This changes the minimal code to get the `fs` crate working for Windows. Able to create files and directories. It should work the same for UNIX systems **and** unblock many of the Windows tasks. Any shortcoming that arises should impact Windows only so we have room to fix and improve the crate.

It's done by gating the only UNIX-specific details (e.g. permissions) to the UNIX target family while leaving the rest out of the conditional compilation macro, so the functions are available to all targets.

## Checklist

### DIFF
--- a/fs/src/writer_file.rs
+++ b/fs/src/writer_file.rs
@@ -105,12 +105,8 @@ pub mod tests {
     use std::fs;
     use std::path::PathBuf;
 
-    #[cfg(target_family = "unix")]
     #[test]
     fn test_file_writer_content_and_permissions() {
-        use std::fs::metadata;
-        use std::os::unix::fs::PermissionsExt;
-
         // Prepare temp path and content for the file
         let file_name = "some_file";
         let content = "some content";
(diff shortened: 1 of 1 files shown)


fs/src/writer_file.rs
```diff
@@ -1,12 +1,9 @@
 use super::LocalFile;
 use super::directory_manager::DirectoryManagementError;
 use super::utils::{FsError, validate_path};
+#[cfg(target_family = "unix")]
```
sigilioso: Nit: should we move the import to the unix block? As we do in the other modules.

DavSanchez:
> I'm not sure if we still need to revisit something in the `fs` module regarding Windows implementation (I case we need to add some sort of TODO or something).

At the time I did not see any obvious follow-ups so I didn't add TODOs.

In any case, I expect the work for the process management on Windows will uncover any possible shortcomings that we can fix on the spot or split into different tasks.

sigilioso (approved):
🚀

PR #1857, opened 2025-11-12
chore: improve windows pr checks (NR-480762)

gsanchezgavier:
- Moves the release build to nightly
- Add clippy check for windows 

Note: I worked on some of the clippy warnings but most of them are in the integration test and command which are still pending to be reworked for windows. Because of this i disable the "treat warning as errors" for now


gsanchezgavier:
changes done in #1862

PR #1838, opened 2025-11-05
feat(windows): enable more testing surface

DavSanchez:
This PR follows up on https://github.com/newrelic/newrelic-agent-control/pull/1862 and enables more tests for Windows. Only those related to running as privileged (by the time work started on this, might have diverged from https://github.com/newrelic/newrelic-agent-control/pull/1861) and the ones for K8s are gated for UNIX only.

Can still add more tests for example on restarting processes, but this can be reviewed already.

## Checklist

- [x] Provided a meaningful title following conventional commit style.
- [x] Included a detailed description for the Pull Request.
- [ ] Documentation under `docs` is aligned with the change.
- [x] Follows guidelines for Pull Requests in [`CONTRIBUTING.md`](../blob/main/CONTRIBUTING.md).
- [ ] Follows [`log level guidelines`](../blob/main/docs/style/logs.md).

PR #1862, opened 2025-11-13
feat(windows): implement process management 1

DavSanchez:

First stage of the process management for Windows.

Essentially, I create the process the same way as done in the former implementation, but on Windows a [process creation flag](https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags) is added to the `Command`, `CREATE_NEW_PROCESS_GROUP`, which allows to send `CTRL+BREAK` events to it via [`GenerateConsoleCtrlEvent`](https://learn.microsoft.com/en-us/windows/console/generateconsolectrlevent) to perform a graceful shutdown. I mainly took inspiration from the current implementation of task termination logic for Windows in the [GitLab runners codebase](https://gitlab.com/gitlab-org/gitlab-runner/-/blob/397ba5dc2685e7b13feaccbfed4c242646955334/helpers/process/killer_windows.go#L75-108).

Not all the tests that can possibly run for Windows at this stage are enabled. Following up on PR https://github.com/newrelic/newrelic-agent-control/pull/1838.

## Checklist

### DIFF
--- a/agent-control/src/sub_agent/on_host/supervisor.rs
+++ b/agent-control/src/sub_agent/on_host/supervisor.rs
@@ -426,6 +426,8 @@ pub mod tests {
     use crate::sub_agent::on_host::command::restart_policy::BackoffStrategy;
     use crate::sub_agent::on_host::command::restart_policy::{Backoff, RestartPolicy};
     use rstest::*;
+    use serde::Deserialize;
+    use std::collections::HashMap;
     use std::thread;
     use std::time::{Duration, Instant};
     use tracing_test::internal::logs_with_scope_contain;
@@ -433,21 +435,45 @@ pub mod tests {
 
     const DURATION_DELTA: std::time::Duration = Duration::from_millis(100);
 
+    #[derive(Clone, Deserialize)]
+    struct TextExecutableData {
+        id: String,
+        path: String,
+        #[serde(default)]
+        args: Vec<String>,
+        #[serde(default)]
+        env: HashMap<String, String>,
+    }
+
+    impl From<TextExecutableData> for ExecutableData {
+        fn from(text_exec_data: TextExecutableData) -> Self {
+            ExecutableData::new(text_exec_data.id, text_exec_data.path)
+                .with_args(text_exec_data.args)
+                .with_env(text_exec_data.env)
+        }
+    }
+
+    fn build_test_exec_data(json_str: &str) -> ExecutableData {
+        serde_json::from_str::<TextExecutableData>(json_str)
+            .expect("Test input should be deserializable to an ExecutableData")
+            .into()
+    }
+
     #[traced_test]
     #[rstest]
     #[cfg_attr(target_family = "unix",case::long_running_process_shutdown_after_start(
         "long-running",
-        ExecutableData::new("sleep".to_owned(), "sleep".to_owned()).with_args(vec!["10".to_owned()]),
+        build_test_exec_data(r#"{"id":"sleep","path":"sleep","args":["10"]}"#),
         Some(Duration::from_secs(1)),
         vec!["Stopping executable", "Executable terminated"]))]
     #[cfg_attr(target_family = "windows",case::long_running_process_shutdown_after_start(
         "long-running",
-        ExecutableData::new("cmd".to_owned(), "cmd".to_owned()).with_args(vec!["/C".to_owned(), "timeout".to_owned(), "/T".to_owned(), "10".to_owned(), "/NOBREAK".to_owned()]),
+        build_test_exec_data(r#"{"id":"cmd","path":"cmd","args":["/C","timeout","/T","10","/NOBREAK"]}"#),
         Some(Duration::from_secs(1)),
         vec!["Stopping executable", "Executable terminated"]))]
     #[case::fail_process_shutdown_after_start(
         "wrong-command",
-        ExecutableData::new("wrong-command".to_owned(), "wrong-command".to_owned()),
+        build_test_exec_data(r#"{"id":"wrong-command","path":"wrong-command"}"#),
         Some(Duration::from_secs(1)),
         vec!["Executable not running"])]
     fn test_supervisor_gracefully_shutdown(
@@ -543,8 +569,7 @@ pub mod tests {
             .with_last_retry_interval(Duration::new(30, 0));
 
         let executables = vec![
-            ExecutableData::new("wrong-command".to_owned(), "wrong-command".to_owned())
-                .with_args(vec!["x".to_owned()])
+            build_test_exec_data(r#"{"id":"wrong-command","path":"wrong-command","args":["x"]}"#)
                 .with_restart_policy(RestartPolicy::new(BackoffStrategy::Fixed(backoff))),
         ];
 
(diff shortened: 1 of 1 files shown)


agent-control/src/agent_control/uptime_report.rs
```diff
@@ -227,11 +227,11 @@ mod tests {
             rest.iter()
                 .try_fold(first, |previous, current| {
                     // check that the uptime is increasing by interval with a tolerance of 16ms
-                    if Duration::abs_diff(*previous + Duration::from(config.interval), *current) < Duration::from_millis(16) {
+                    if Duration::abs_diff(*previous + Duration::from(config.interval), *current) < Duration::from_millis(25) {
                         Ok(current)
                     } else {
                         Err(format!(
-                            "uptime diff exceeding 16ms toleration. Previous: {previous:?}. Current: {current:?}"
+                            "uptime diff exceeding 25ms toleration. Previous: {previous:?}. Current: {current:?}"
```
sigilioso: Nit: We also need to update the comment above


```suggestion
                    // check that the uptime is increasing by interval with a tolerance
                    let tolerance = Duration::from_milis(25);
                    if Duration::abs_diff(*previous + Duration::from(config.interval), *current) < Duration::from_millis(25) {
                        Ok(current)
                    } else {
                         Err(format!(
                            "uptime diff exceeding {tolerance} toleration. Previous: {previous:?}. Current: {current:?}"
```

agent-control/src/sub_agent/on_host/command/command_os.rs
```diff
@@ -122,58 +137,56 @@ impl CommandOSStarted {
...
-            if graceful_shutdown_result.is_err()
-                || self.is_running_after_timeout(self.shutdown_timeout)
-            {
-                self.process.kill().map_err(CommandError::from)?;
+        while Instant::now() < deadline {
+            if self.is_running() {
+                std::thread::sleep(POLL_INTERVAL);
+            } else {
+                return false;
             }
```
sigilioso: Nit:(although I believe this hasn't actually changed)


```suggestion
            if !self.is_running() {
                return false;
            }
            std::thread::sleep(POLL_INTERVAL);
}
```

I think it is a bit more readable, is it just me?

agent-control/src/sub_agent/on_host/command/command_os.rs
```diff
@@ -122,58 +137,56 @@ impl CommandOSStarted {
...
-                    return false;
-                }
-            }
+    pub fn shutdown(&mut self) -> Result<(), CommandError> {
+        let pid = self.get_pid();
+
+        // Attempt a graceful shutdown (platform-dependent).
+        #[cfg(not(any(target_family = "windows", target_family = "unix")))]
+        let graceful_shutdown_result =
+            Err(io::Error::other("Unsupported platform for graceful shutdown").into());
```
sigilioso: Are we generally taking care of these targets? Would it make sense to use [compile_error](https://doc.rust-lang.org/std/macro.compile_error.html) so we assure we are actively taking care of it if ever supporting such targets?

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -433,14 +433,18 @@ pub mod tests {
...
         "long-running",
         ExecutableData::new("sleep".to_owned(), "sleep".to_owned()).with_args(vec!["10".to_owned()]),
         Some(Duration::from_secs(1)),
-        vec!["Stopping executable", "Executable terminated"])]
+        vec!["Stopping executable", "Executable terminated"]))]
+    #[cfg_attr(target_family = "windows",case::long_running_process_shutdown_after_start(
+        "long-running",
+        ExecutableData::new("cmd".to_owned(), "cmd".to_owned()).with_args(vec!["/C".to_owned(), "timeout".to_owned(), "/T".to_owned(), "10".to_owned(), "/NOBREAK".to_owned()]),
+        Some(Duration::from_secs(1)),
+        vec!["Stopping executable", "Executable terminated"]))]
```
sigilioso: I guess the `target_family = "unix"` was added because the use case doesn't work yet. If that wasn't the reason why adding it, I'd also remove the windows part.

.github/workflows/push_pr_checks_tests.yml
gsanchezgavier: I'm following a slightly differnt approach in the pr for this https://github.com/newrelic/newrelic-agent-control/pull/1857/files#diff-b8aabbcf7b30fee3905e919e4f12a21d52537a34a66bc77c5710e6853bae6118R358, lets sync offline

sigilioso (commented):
Thanks for taking care of this and so many integration tests!

It is kind of surprising how little the `command_os` module needs to change, maybe it is a good sign 🙂 

I've left some comments, mostly _nits_

gsanchezgavier (commented):
thanks ! looks great i left just some minor comments

agent-control/tests/integration_tests.rs
```diff
@@ -1,6 +1,6 @@
 //! Entry point for the integration tests.
 #![warn(missing_docs)]
-
+#![cfg_attr(target_family = "windows", allow(unused_imports, dead_code))] // to remove once all code paths are covered for windows
 mod common;
 mod k8s;
 mod on_host;
```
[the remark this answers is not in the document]
DavSanchez: We can have some kind of [`pgrep`-like check for processes](https://github.com/newrelic/newrelic-agent-control/blob/5033bdb26af0e41a1a81f342896872bbc9d9ec5f/agent-control/tests/on_host/id.rs#L255-L258), so should be doable. Given the integration tests are being adapted in https://github.com/newrelic/newrelic-agent-control/pull/1838 I can add that test there if you don't mind!

agent-control/src/sub_agent/on_host/command/command_os.rs
```diff
@@ -122,58 +137,56 @@ impl CommandOSStarted {
...
-                    std::thread::sleep(POLL_INTERVAL);
-                } else {
-                    return false;
-                }
-            }
+    pub fn shutdown(&mut self) -> Result<(), CommandError> {
+        let pid = self.get_pid();
+
+        // Attempt a graceful shutdown (platform-dependent).
+        #[cfg(not(any(target_family = "windows", target_family = "unix")))]
```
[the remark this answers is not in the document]
DavSanchez: I can extract it to a different function as suggested by @sigilioso and put an `unimplemented!` for uncovered targets. I think that would be cleaner.

Cargo.toml
```diff
@@ -43,6 +43,7 @@ reqwest = { version = "0.12.24", default-features = false, features = [
   "socks",
 ] }
 rstest = "0.26.1"
+windows = "0.62.2"
```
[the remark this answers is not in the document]
DavSanchez: Nop, these are different but part of the same ecosystem: https://github.com/microsoft/windows-rs

agent-control/src/sub_agent/on_host/command/command_os.rs
DavSanchez: https://github.com/newrelic/newrelic-agent-control/pull/1862/commits/d77c6bd6da4a27cac5bfe36c472929b071d5369a

agent-control/src/sub_agent/on_host/command/command_os.rs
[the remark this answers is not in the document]
DavSanchez: https://github.com/newrelic/newrelic-agent-control/pull/1862/commits/4b1a7924c7fd7b46981c31e2079d1b76683bee98

agent-control/src/sub_agent/on_host/supervisor.rs
sigilioso: I totally misread it! Thanks for the explanation!

sigilioso (approved):
Considering that the integration tests will be extended in #1838 this LGTM already. Thanks!

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -582,9 +586,18 @@ pub mod tests {
...
                 .with_args(vec!["NR-command".to_owned()])
                 .with_restart_policy(RestartPolicy::new(BackoffStrategy::Fixed(backoff))),
+            #[cfg(target_family = "windows")]
+            ExecutableData::new("cmd".to_owned(), "cmd".to_owned())
+                .with_args(vec![
+                    "/C".to_owned(),
+                    "echo".to_owned(),
+                    "NR-command".to_owned(),
+                ])
+                .with_restart_policy(RestartPolicy::new(BackoffStrategy::Fixed(backoff))),
```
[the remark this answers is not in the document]
DavSanchez: WDYT? https://github.com/newrelic/newrelic-agent-control/pull/1862/commits/200892e27d2f4ffbadf1791be0a64ed96dfee8c9

sigilioso (approved):

PR #1872, opened 2025-11-18
test: proxy e2e onhost (NR-480755)

gsanchezgavier:
- Add e2e proxy for linux onhost. The test forces the environment to actually reject net comms outside the proxy so it can help to catch new features that are adding to AC without proxy support. 
- Do some general improvement in other test
- Remove a (manually executed) proxy integration test

### DIFF
--- a/test/onhost-e2e/ansible/proxy.yaml
+++ b/test/onhost-e2e/ansible/proxy.yaml
@@ -5,6 +5,10 @@
   become: true
   gather_facts: yes
 
+  vars:
+    # to be enable on local vm tests as it makes GH runner to hang
+    enable_firewall: false
+
   tasks:
     - name: Cleanup
       include_tasks: ./tasks/clean_all.yaml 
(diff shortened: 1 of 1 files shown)


alvarocabanas (dismissed):
LGTM, good job.

test/onhost-e2e/ansible/proxy.yaml
```diff
@@ -0,0 +1,171 @@
...
+            
+            # Test that proxied traffic works
+            curl -s --max-time 5 --proxy http://localhost:8080 https://www.newrelic.com
+            proxy=$?
+            if [ $proxy -ne 0 ]; then
+              echo "ERROR: Traffic blocked with proxy or proxy not working"
+              exit 1
+            fi
+
+        - name: Install Agent Control
```
danielorihuela: I see this `Install Agent Control` is a step inside `Set firewall rules`. Should it be outside? They are two different things.

test/onhost-e2e/ansible/proxy.yaml
```diff
@@ -0,0 +1,171 @@
...
+              copy:
+                dest: "/etc/newrelic-agent-control/priv.key"
+                # Read private key from an env var directly because of issues passing multiline --extra-vars to ansible-playbook.
+                content: "{{ lookup('ansible.builtin.env', 'NR_SYSTEM_IDENTITY_PRIVATE_KEY') }}"
+                
+            - name: Install Agent Control
+              include_role:
+                name: install_ac_recipe
+              vars:
+                ac_proxy_url: "http://localhost:8080"
```
danielorihuela: In `debian.yaml` this variable is passed to `HTTPS_PROXY`, shouldn't we use `http://localhost:8080` then?

test/onhost-e2e/ansible/proxy.yaml
gsanchezgavier: well the idea is to catch issues on the installation recipe to. For instance if tomorrow we add some other cli in there as part of the installation and we miss the proxy this test will fail. We have some good experience about this with the k8s one :)

danielorihuela (approved):

PR #1875, opened 2025-11-19
feat: restore similar cache when exact match is missing

danielorihuela:
Improve rust compile times by restoring a "similar" cache.

> Disclaimer: Some pipelines are failing due to unused dependencies. Those were added for testing out the behaviour and they will be removed if we decide to merge the PR.

### Implementation

Until now, we only restored a cache that matched the unique key configured.

```
- uses: actions/cache@v4
  with:
    path: ...
    key: ${{ runner.os }}-release-cargo-${{ hashFiles('**/Cargo.lock') }}
```

Whenever the cargo lock changes, this triggers a full compilation cycle. Wasting time. The [action/cache](https://github.com/actions/cache) github action includes a `restore-keys` input to restore keys based on substring match.

For example,

```
- uses: actions/cache@v4
  with:
    path: ...
    key: ${{ runner.os }}-release-cargo-${{ hashFiles('**/Cargo.lock') }}
    restore-keys: ${{ runner.os }}-release-cargo-
```

This allows restoring an old cache even if the dependencies changed. Thus, we restore a "similar" cache. This, from my point of view, is equivalent to what we locally do. We add new dependencies or remove and rebuild without starting from scratch.

### Cache load order

1. Search cache with exact `key` (PR branch)
2. Search cache with `restore-keys` (PR branch)
3. Search cache with exact `key` (default branch)
4. Search cache with `restore-keys` (default branch)

https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#cache-hits-and-misses
https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#cache-key-matching

## How was this tested?

Added new dependencies and checked that cache was restored even after the `Cargo.lock` changed. For example, https://github.com/newrelic/newrelic-agent-control/pull/1875/commits/ec35fa4b0160daff78ed26de10aa6739b2496bae added a new dependency but could restore a cache.

## Checklist

### DIFF
--- a/.github/workflows/push_pr_checks_tests.yml
+++ b/.github/workflows/push_pr_checks_tests.yml
@@ -63,7 +63,7 @@ jobs:
           toolchain: ${{ env.RUST_VERSION }}
           components: clippy
 
-      - uses: actions/cache/restore@v4
+      - uses: actions/cache@v4
         with:
           path: |
             ~/.cargo/bin/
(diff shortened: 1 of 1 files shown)


gsanchezgavier (approved):
thanks

paologallinaharbur (approved):

PR #1884, opened 2025-11-21
chore: pin gha to commit hashes (NR-487580)

gsanchezgavier:
- pins all gh actions (except from the newrelic owned ones) to a commit hash
- add renovate rule to weekly bump them
- splits renovate config to be reused by other rust repos that the team owns

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,20 +1,7 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
-    "config:recommended",
-    "helpers:pinGitHubActionDigests"
-  ],
-  "labels": [
-    "dependencies"
-  ],
-  "automerge": false,
-  "lockFileMaintenance": {
-    "enabled": true
-  },
-  "enabledManagers": [
-    "cargo",
-    "custom.regex",
-    "gomod"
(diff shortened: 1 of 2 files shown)


.github/renovate.json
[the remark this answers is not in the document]
gsanchezgavier: because is being deprecated by renovate

.github/workflows/ansible_install_test.yml
```diff
@@ -15,7 +15,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
```
danielorihuela: I don't have the context. Why do we need to pin by commit now?

sigilioso (approved):
Thanks 🚀

.github/workflows/ansible_install_test.yml
```diff
@@ -15,7 +15,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
```
paologallinaharbur: why is this useful?
 > pins all gh actions (except from the newrelic owned ones) to a commit hash

I mean if the intent is to improve security then we should not bump automatically hashes, no?

PR #1893, opened 2025-11-24
fix: validate AC remote config signature

gsanchezgavier:
Issue:
We were currently skipping the validation of AC remote config signature due to capabilities check. 
Context:
This capabilities check was done on the first iteration when no AC signature was being sent by the server. Then the serverver start sending it and we did #1744 but without removing the capability check.
Fix:
This PR is removing the capability check as is not used anymore and not capabilities are not fully implemented yet.


alvarocabanas (approved):
LGTM, good find

PR #1896, opened 2025-11-25
feat: validates all remote configs (NR-490635)

gsanchezgavier:
- Validates signature and regex for all configurations coming in the ConfigurationMap from RemoteConfig

### DIFF
--- a/agent-control/src/opamp/remote_config/validators/regexes.rs
+++ b/agent-control/src/opamp/remote_config/validators/regexes.rs
@@ -36,9 +36,13 @@ impl RemoteConfigValidator for RegexValidator {
         opamp_remote_config: &OpampRemoteConfig,
     ) -> Result<(), RegexValidatorError> {
         // This config will fail further on the event processor.
-        if let Ok(raw_config) = opamp_remote_config.get_default() {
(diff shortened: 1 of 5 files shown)


agent-control/src/opamp/remote_config/validators/regexes.rs
```diff
@@ -36,9 +36,13 @@ impl RemoteConfigValidator for RegexValidator {
         opamp_remote_config: &OpampRemoteConfig,
     ) -> Result<(), RegexValidatorError> {
         // This config will fail further on the event processor.
-        if let Ok(raw_config) = opamp_remote_config.get_default() {
-            self.validate_regex_rules(&agent_identity.agent_type_id, raw_config)?;
-        }
+        opamp_remote_config
+            .configs()
+            .0
+            .iter()
```
sigilioso: Instead of making the attribute public as I suggested in the previous comment, we could also implement some `iter_configs()` helper, considering that every validator will need to iterate over the config map.

sigilioso (approved):
Thanks! 🚀 

I've left a single _nit_ in two comments. Feel free to ignore it.

PR #1897, opened 2025-11-25
feat: avoid mounting secret from volumne and get it with the k8s api

vjripoll:
Following the security request to avoid mounting the secret as a volume in K8s, I've updated the TokenRetriever to require the private key itself instead of the file path. 

Now, both the on-host and K8s components function the same way. However, to accommodate this change, the OpAMP client initialization has been moved to the respective run functions for K8s and on-host, rather than in the common run logic.

These changes also require updates to the chart included in this PR: https://github.com/newrelic/helm-charts/pull/2020

## Checklist

### DIFF
--- a/agent-control/src/secret_retriever/k8s/retrieve.rs
+++ b/agent-control/src/secret_retriever/k8s/retrieve.rs
@@ -46,37 +46,17 @@ where
 mod tests {
     use super::*;
     use crate::agent_control::config::AuthSecret;
-    use std::fmt;
+    use assert_matches::assert_matches;
+    use mockall::predicate::*;
+    use mockall::*;
 
-    struct MockProvider {
-        expected_path: String,
-        should_fail: bool,
-        return_value: String,
-    }
+    mock! {
+        pub SecretsProvider {}
 
-    #[derive(Debug)]
-    struct MockError(String);
-    impl fmt::Display for MockError {
-        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-            write!(f, "{}", self.0)
-        }
-    }
-    impl std::error::Error for MockError {}
+        impl SecretsProvider for SecretsProvider {
+            type Error = std::io::Error;
 
-    impl SecretsProvider for MockProvider {
-        type Error = MockError;
-
-        fn get_secret(&self, secret_path: &str) -> Result<String, Self::Error> {
-            assert_eq!(
-                secret_path, self.expected_path,
-                "The retriever constructed the wrong secret path!"
-            );
-
-            if self.should_fail {
-                Err(MockError("Simulated K8s failure".to_string()))
-            } else {
-                Ok(self.return_value.clone())
-            }
+            fn get_secret(&self, secret_path: &str) -> Result<String, std::io::Error>;
         }
     }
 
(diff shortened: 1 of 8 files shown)


agent-control/src/opamp/builder.rs
```diff
@@ -0,0 +1,30 @@
...
+) -> Result<OpAMPHttpClientBuilder<TokenRetrieverImpl>, RunError> {
+    debug!("OpAMP configuration found, creating an OpAMP client builder");
+
+    let token_retriever = Arc::new(
+        TokenRetrieverImpl::try_build(
+            opamp_config.clone().auth_config,
+            private_key,
+            proxy_config.clone(),
+        )
+        .inspect_err(|err| error!(error_msg=%err, "Building token retriever failed"))
```
sigilioso: I'd take the opportunity to align to the logging guidelines:


```suggestion
        .inspect_err(|err| error!("Could not build OpAMP's token retriever: {err}"))
```

agent-control/src/agent_control/run/k8s.rs
```diff
@@ -66,9 +68,28 @@ impl AgentControlRunner {
...
 
+        let opamp_http_builder = if let Some(opamp_config) = &self.opamp {
+            let k8s_secret_provider = K8sSecretProvider::new(k8s_client.clone());
+
+            let secret_path = format!(
+                "{}:{}:{}",
+                self.k8s_config.namespace, self.k8s_config.secret_private_key_name, K8S_KEY_SECRET
+            );
+
+            let secret = k8s_secret_provider.get_secret(&secret_path)?;
```
sigilioso: I'd consider moving this to a helper function, also improve naming to explicitly say that we are getting the `private_key`.

agent-control/src/secrets_provider/file.rs
```diff
@@ -0,0 +1,82 @@
+use crate::secrets_provider::SecretsProvider;
+use std::fs;
+use std::path::PathBuf;
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+#[error("reading file secret: {0}")]
+pub struct FileSecretProviderError(String);
```
DavSanchez: I'd use [`std::io::Error`](https://doc.rust-lang.org/std/io/struct.Error.html) directly for this instead of a custom error type. After all we don't do anything outside opening a file.

agent-control/src/secret_retriever/k8s/retrieve.rs
```diff
@@ -0,0 +1,138 @@
...
+    #[derive(Debug)]
+    struct MockError(String);
+    impl fmt::Display for MockError {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            write!(f, "{}", self.0)
+        }
+    }
+    impl std::error::Error for MockError {}
+
+    impl SecretsProvider for MockProvider {
```
sigilioso: Any particular reason to use a "custom mock" instead of [mockall::mock!](https://docs.rs/mockall/latest/mockall/macro.mock.html) here?

agent-control/src/secret_retriever/k8s/retrieve.rs
```diff
@@ -0,0 +1,138 @@
...
+        };
+
+        let retriever = K8sSecretRetriever::new(mock_provider, config);
+
+        let opamp_config = OpAMPClientConfig::default();
+
+        let result = retriever.retrieve(&opamp_config);
+
+        assert!(result.is_ok());
+        assert_eq!(result.unwrap(), "SUPER_SECRET_TOKEN");
```
sigilioso: Nit:
```suggestion
        let result = retriever.retrieve(&opamp_config).expect("retrieve should not fail");

        assert_eq!(result, "SUPER_SECRET_TOKEN");
```

agent-control/src/secret_retriever/on_host/retrieve.rs
```diff
@@ -0,0 +1,196 @@
...
+    #[derive(Debug)]
+    struct MockError;
+    impl fmt::Display for MockError {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            write!(f, "MockError")
+        }
+    }
+    impl std::error::Error for MockError {}
+
+    impl SecretsProvider for MockProvider {
```
sigilioso: Same here, why not [mockall::mock!](https://docs.rs/mockall/latest/mockall/macro.mock.html)?

sigilioso (approved):
Last _nit_, but it LGTM already

PR #1912, opened 2025-12-01
feat: support multi-config parsing for Agent Control (NR-490637)

gsanchezgavier:
on top of #1902 

- Adds the feature to split AC remote config in multiple files . Keys are appended and error on collisions except for the `agents` key. All keys inside `agents` are appended into a single one erroring on collisions. 

Other changes:
- Due to how merging is handle now, unrecognised keys are no longer stored. 

Note:
- Working on integration tests on a different PR.

PR #1913, opened 2025-12-02
test: add multiconfig integration tests (NR-490637)

gsanchezgavier:
on top of #1912 

- Adds integration tests covering split remote configs


agent-control/tests/on_host/scenarios/multiple_remote_config.rs
sigilioso: Would checking that AC sets the remote config as failed when there are conflicting values be too much for an integration test?

PR #1919, opened 2025-12-03
feat(cli): support proxy on infra-agent config [NR-490718]

sigilioso:
This PR add support for including the AC's proxy configuration in the generated configuration for the infrastructure agent regardless of if the configuration is _migrated_ or new. There are also some minor changes regarding how the proxy configuration is handled in on-host cli.

## Summary of changes
* Add support for proxy configuration when migrating/generating the infra-agent configuration.
* Move the `ProxyConfig` struct to a common module so the same configuration is used for the AC's and infra-agent config generation.
* Avoids deserializing empty proxy configuration as `{"ignore_system_proxy": false}`. Such configuration is now rendered as `{}`, which is the default and keeps the very same behavior.

## Checklist

### DIFF
--- a/agent-control/src/cli/on_host/config_gen/config.rs
+++ b/agent-control/src/cli/on_host/config_gen/config.rs
@@ -83,7 +83,7 @@ pub struct Agent {
 
 /// Helper to avoid deserializing proxy values when empty or default
 fn is_none_or_empty(v: &Option<ProxyConfig>) -> bool {
-    v.as_ref().map(|v| v.is_empty()).unwrap_or(true)
+    v.as_ref().is_none_or(|v| v.is_empty())
 }
 
 #[cfg(test)]
@@ -116,7 +116,7 @@ mod tests {
     }
 
     #[rstest]
-    #[case(
+    #[case::no_proxy_config(
         Config {
             fleet_control: None,
             server: Server { enabled: true },
@@ -125,7 +125,7 @@ mod tests {
         },
         r#"{"server":{"enabled":true},"agents":{}}"#
     )]
-    #[case(
+    #[case::empty_proxy_config(
         Config {
             fleet_control: None,
             server: Server { enabled: false },
@@ -134,7 +134,7 @@ mod tests {
         },
         r#"{"server":{"enabled":false},"agents":{}}"#
     )]
-    #[case(
+    #[case::some_proxy_config(
         Config {
             fleet_control: None,
             server: Server { enabled: true },
--- a/agent-control/src/cli/on_host/proxy_config.rs
+++ b/agent-control/src/cli/on_host/proxy_config.rs
@@ -40,14 +40,14 @@ impl ProxyConfig {
             self.proxy_ca_bundle_file.as_ref(),
         ]
         .iter()
-        .all(|v| v.map(|v| v.is_empty()).unwrap_or(true))
+        .all(|v| v.is_none_or(|s| s.is_empty()))
             && !self.ignore_system_proxy
     }
 }
 
 // Helper to avoid serializing empty values
 fn is_none_or_empty_string(v: &Option<String>) -> bool {
-    v.as_ref().map(|s| s.is_empty()).unwrap_or(true)
+    v.as_ref().is_none_or(|s| s.is_empty())
 }
 
 // Helper to avoid serializing false values


agent-control/src/cli/on_host/host_monitoring_gen/infra_config.rs
```diff
@@ -119,14 +120,42 @@ impl InfraConfig {
         self
     }
 
-    pub fn with_proxy(mut self, proxy: &str) -> Self {
-        if !proxy.trim().is_empty() {
+    pub fn setup_proxy(mut self, proxy: Option<ProxyConfig>) -> Self {
+        // Delete previous proxy configuration
```
gsanchezgavier: have you check which are the configs that are set whenever the infra=agent is installed used the guided install? just curious about if we are supporting that use case.

agent-control/src/cli/on_host/config_gen/config.rs
```diff
@@ -149,53 +115,36 @@ mod tests {
...
-        ProxyConfig{proxy_url: Some("https://proxy.url".into()), proxy_ca_bundle_dir: Some("/bundle/dir".into()), proxy_ca_bundle_file: Some("/bundle/file".into()), ignore_system_proxy: true},
+        Config {
+            fleet_control: None,
+            server: Server { enabled: true },
+            proxy: None,
+            agents: HashMap::new(),
+        },
+        r#"{"server":{"enabled":true},"agents":{}}"#
+    )]
+    #[case(
```
gsanchezgavier: nit:  add the case name so is easy to understand what is being tested

agent-control/src/cli/on_host/proxy_config.rs
```diff
@@ -0,0 +1,120 @@
...
+    #[arg(long, required = false)]
+    pub proxy_ca_bundle_file: Option<String>,
+
+    #[serde(skip_serializing_if = "is_false")]
+    #[arg(long, default_value_t = false, value_parser = ignore_system_proxy_parser, action = clap::ArgAction::Set)]
+    pub ignore_system_proxy: bool,
+}
+
+impl ProxyConfig {
+    pub fn is_empty(&self) -> bool {
```
gsanchezgavier: nit: would be possible to skip serializing the whole proxy config in case it's equal to the default value, would make sense? perhaps it cannot be done because is flattened

agent-control/src/cli/on_host/host_monitoring_gen/infra_config.rs
sigilioso: I've checked it and we are not supporting the same now (we were supporting it before!).

[Ref](https://github.com/newrelic/open-install-library/blob/0f9fe836dfd93519ce7760293acad1db48526323/recipes/newrelic/infrastructure/debian.yml#L184-L187)

```yaml
    setup_proxy:
      cmds:
        - |
          if [ ! -z "$HTTPS_PROXY" ]; then
            sed -i "/^proxy/d" /etc/newrelic-infra.yml
            echo 'proxy: {{.HTTPS_PROXY}}' >> /etc/newrelic-infra.yml
          fi

```

The same proxy configuration used by the cli itself, which is exactly what we are doing on AC at the moment (and wasn't coherent with the guided install proxy setup).

I'd say that customers with a proper proxy setup didn't use the infra-agent's guided install or they needed to modify the configuration manually after installation.

agent-control/src/cli/on_host/proxy_config.rs
sigilioso: I tried to, but I didn't find a way to do it because of the flattened thing 🙁

agent-control/src/cli/on_host/config_gen/config.rs
```diff
@@ -119,10 +81,14 @@ pub struct Agent {
     pub agent_type: String,
 }
 
+/// Helper to avoid deserializing proxy values when empty or default
+fn is_none_or_empty(v: &Option<ProxyConfig>) -> bool {
+    v.as_ref().map(|v| v.is_empty()).unwrap_or(true)
```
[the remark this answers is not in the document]
gsanchezgavier: cool is functional and readable :)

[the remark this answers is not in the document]
sigilioso: Changed here [17aaa2a](https://github.com/newrelic/newrelic-agent-control/pull/1919/commits/17aaa2a0f596b7c472f55787b8bf4b9e2f0a93f8)

agent-control/src/cli/on_host/config_gen/config.rs
sigilioso: Names added here: [17aaa2a](https://github.com/newrelic/newrelic-agent-control/pull/1919/commits/17aaa2a0f596b7c472f55787b8bf4b9e2f0a93f8)

gsanchezgavier (approved):
🚀

PR #1931, opened 2025-12-09
fix: cache size in main

danielorihuela:
This should solve the issue with the ever increasing cache size. The different use cases are:

* Only restore -> Use "nearest" for fastest pipelines.
* Restore + save (branch) -> Use "nearest" for fastest pipelines at the expense of bigger caches (only in branches, not sure if this could be a problem too).
* Restore + save (main) -> Use "exact" to avoid the increasing cache size problem.

## Checklist


gsanchezgavier (approved):
awesome we need to use this actions pattern more in our pipelines,

gsanchezgavier (approved):

PR #1933, opened 2025-12-09
refactor: install rust toolchain composite action

danielorihuela:
Improved pipelines a bit. Created composite action to install rust toolchain.

## Context

I noticed we don't create caches in the `nightly`. I wanted to fix it. It's not that "easy". We have to install different components in different OSes. I also noticed that Linux and Windows don't work the same. For example, in Windows we have a single `test-dev` cache for every step. In Linux we don't. I remember adding that distinction because it made it faster on Linux.

Anyway, instead of trying to fix all at once. I decided to start small and improve one thing at a time whenever I have a bit of time.

The idea is to end up with smaller more manageable units that we can easily mix and compose in the pipelines. Otherwise, it's "hard" to make modifications.

## Checklist

### DIFF
--- a/.github/actions/install-rust-toolchain/action.yml
+++ b/.github/actions/install-rust-toolchain/action.yml
@@ -1,0 +1,24 @@
+name: 'Install Rust Toolchain'
+description: 'Installs the Rust toolchain version specified in Cargo.toml'
+inputs:
+  targets:
+    description: 'Comma-separated list of additional targets to install'
+    required: false
+  components:
+    description: 'Comma-separated list of additional components to install'
+    required: false
+runs:
+  using: "composite"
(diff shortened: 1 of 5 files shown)


.github/actions/install-rust-toolchain/action.yml
```diff
@@ -0,0 +1,24 @@
...
+inputs:
+  targets:
+    description: 'Comma-separated list of additional targets to install'
+    required: false
+  components:
+    description: 'Comma-separated list of additional components to install'
+    required: false
+runs:
+  using: "composite"
+  steps:
```
gsanchezgavier: would make sense to also include the zig installation?

PR #1949, opened 2025-12-15
fix: rust-cache not saving cache as post action

danielorihuela:
Fixes the rust-cache composite action. It was not working properly. As a result, caches were not being saved as a post step.

Seems like actions/cache/save creates the cache when it's called as opposed to actions/cache which creates a "post step" at the end of the workflow to save the cache.

Because of that, I had to change the composite action. Notice that now we call actions/cache/restore when we want to just restore data and actions/cache when we want to restore and save.

PD: Even though the save-cache input is a boolean, we need to treat it as a string for some reason.

## Checklist

### DIFF
--- a/.github/actions/rust-cache/action.yml
+++ b/.github/actions/rust-cache/action.yml
@@ -5,6 +5,51 @@ inputs:
     description: 'A unique identifier that will be used as part of the cache key'
     required: true
     type: string
+  # Why do we need a "restore-strategy" input field?
+  #
+  # Before going straight to the explanation, please read how Github Action Caches work.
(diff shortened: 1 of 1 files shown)


.github/actions/rust-cache/action.yml
```diff
@@ -5,6 +5,51 @@ inputs:
     description: 'A unique identifier that will be used as part of the cache key'
     required: true
     type: string
+  # Why do we need a "restore-strategy" input field?
```
gsanchezgavier: o yeah is log, but cool , i remember it wasn't easy to explain , thanks for adding it

PR #1956, opened 2025-12-16
feat: generic opamp attributes reporter

danielorihuela:
Builds the machinery required to add arbitrary identifiiyng and non-identifying opamp attributes.
We also refactored the version checker to use the new approach.

## Checklist

### DIFF
--- a/agent-control/src/opamp/attributes.rs
+++ b/agent-control/src/opamp/attributes.rs
@@ -81,7 +81,7 @@ where
 }
 
 fn update_agent_description_attributes(
-    mut agent_description: AgentDescription,
+    old_agent_description: AgentDescription,
     new_attributes: Vec<Attribute>,
 ) -> AgentDescription {
     let (new_identifying_attributes, new_non_identifying_attributes): (Vec<_>, Vec<_>) =
@@ -89,6 +89,7 @@ fn update_agent_description_attributes(
             .into_iter()
             .partition(|attribute| attribute.attribute_type == AttributeType::Identifying);
 
+    let mut agent_description = old_agent_description;
     let key_value_iter = |attrs: Vec<Attribute>| attrs.into_iter().map(|attr| attr.key_value);
     agent_description.identifying_attributes = merge_attributes(
         agent_description.identifying_attributes,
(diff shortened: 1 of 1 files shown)


agent-control/src/opamp/attributes.rs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+/// Updates the attributes of the OpAMP agent
+///
+/// If an attribute already exists, it will be updated. If it doesn't, it will be added.
+/// If the `new_attributes` contain duplicated keys, the last occurrence will be kept.
+pub fn update_opamp_attributes<C>(
+    opamp_client: &C,
+    new_attributes: Vec<Attribute>,
+) -> Result<(), Box<dyn Error>>
```
sigilioso: Nit: should we return `opamp_client::client::ClientError` here?

agent-control/src/agent_control/run/k8s.rs
```diff
@@ -285,7 +285,7 @@ fn start_cd_version_checker(
         // where T is the "event" sendable by the above publisher.
         // Using an enum variant that wraps a type is the same as a function taking the type.
         // Same as passing "|x| AgentControlInternalEvent::AgentControlCdVersionUpdated(x)"
-        AgentControlInternalEvent::AgentControlCdVersionUpdated,
+        AgentControlInternalEvent::AgentControlAttributesUpdated,
```
paologallinaharbur: the message above is outdated!

paologallinaharbur: would make sense to have the concept of "extra attributes"?

agent-control/src/opamp/attributes.rs
danielorihuela: I don't know. It doesn't matter. We just print it in the callers.

https://github.com/newrelic/newrelic-agent-control/pull/1956/changes/42b95450a501c3102c09c01b7874b305229632d4

paologallinaharbur (dismissed):
I think that I would implement it in the same way. I do not like that much that you cannot "unset" attributes, but it is not a use case we have for now.

If we notice that with the GUID we need that we can add it!

agent-control/src/opamp/attributes.rs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+/// Merges new attributes into old attributes
+///
+/// If an attribute already exists, it will be updated. If it doesn't, it will be added.
+/// If the `new_attributes` contain duplicated keys, the last occurrence will be kept.
+fn merge_attributes(
+    mut old_attributes: Vec<KeyValue>,
+    new_attributes: Vec<KeyValue>,
+) -> Vec<KeyValue> {
```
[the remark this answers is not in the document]
danielorihuela: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=c17e4a6473b5a19af8a55eeea0dde8f7

PR #1968, opened 2025-12-18
feat: load environment variables from file

gsanchezgavier:
During exploration task on how to pass secrets from envs on windows we discover that setting them as service environment variables makes them easy to access to non-admin user. So achiving what is currently done with systemd envfile config is not possible. 

# feat
Loads env files from the file `/<local-path>/environment_variables.yaml` and set them into the AC running process. The file is located in a protected directory only administrators user have access.

# other
- modify the cli so it creates the env file.
- modify the win install script so it uses the feature.

# pending work out of the scope of this PR
- Remove the current systemd `.conf` file approach so win and unix uses the same mechanism

### DIFF
--- a/agent-control/src/utils/env_var.rs
+++ b/agent-control/src/utils/env_var.rs
@@ -2,19 +2,18 @@ use fs::file_reader::FileReader;
 use std::collections::HashMap;
 use std::error::Error;
 use std::path::Path;
-// Key validation regex, start with letter/underscore, then letters/digits/underscore.
+/// Key validation regex, start with letter/underscore, then letters/digits/underscore.
 const KEY_VALIDATION_REGEX: &str = r"^[A-Za-z_][A-Za-z0-9_]*$";
 
-/// Load environment variables into the current process from a YAML file located at `path`.
-/// The YAML file should contain key-value pairs where keys are the environment variable names
-/// and values are their corresponding values.
+/// Loads environment variables into the current process from a YAML file at `path`.
+///
+/// The YAML document must be a mapping of string keys to string values. Keys must match
+/// `^[A-Za-z_][A-Za-z0-9_]*$`; values must not contain the NUL (`\0`) character.
+///
 /// # Errors
-/// Returns an error if the file cannot be read, if the YAML is malformed,
-/// or if any key doesn't match `^[A-Za-z_][A-Za-z0-9_]*$` or if any value contains NUL characters.
-/// # Safety
-/// This function uses `std::env::set_var` which is not thread-safe on all platforms.
-/// It is the caller's responsibility to ensure that no other threads are concurrently
-/// accessing environment variables while this function is executing.
+///
+/// Returns an error if the file cannot be read, if the YAML content cannot be parsed as a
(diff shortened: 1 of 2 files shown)


agent-control/src/utils/env_var.rs
```diff
@@ -0,0 +1,103 @@
...
+/// Load environment variables into the current process from a YAML file located at `path`.
+/// The YAML file should contain key-value pairs where keys are the environment variable names
+/// and values are their corresponding values.
+/// # Errors
+/// Returns an error if the file cannot be read, if the YAML is malformed,
+/// or if any key doesn't match `^[A-Za-z_][A-Za-z0-9_]*$` or if any value contains NUL characters.
+/// # Safety
+/// This function uses `std::env::set_var` which is not thread-safe on all platforms.
+/// It is the caller's responsibility to ensure that no other threads are concurrently
+/// accessing environment variables while this function is executing.
```
danielorihuela: Can we put some newlines to separate sections like in other parts of the code?

```
/// A trait for types that support templating using a set of variables.
///
/// Implementers replace placeholders in their content with values from a
/// provided `Variables` map. Placeholders follow a specific format, such as
/// `${nr-var:key}`.
///
/// This trait allows to specify the output type with the `Output` associated type.
///
/// # Errors
///
/// Returns an `AgentTypeError` if a placeholder references an undefined or
/// missing variable.
pub trait Templateable {
    type Output;
```

You can see https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#appendix-a-full-conventions-text the RFC of rust docstrings. They have other stuff we are not being strict about like using the "third person singular present indicative" on the first sentence.

agent-control/src/utils/env_var.rs
```diff
@@ -0,0 +1,103 @@
...
+    let file = fs::LocalFile {};
+    let content = file.read(path)?;
+
+    let env_vars: HashMap<String, String> = serde_yaml::from_str(&content)?;
+
+    load_env_from_hashmap(env_vars)
+}
+
+fn load_env_from_hashmap(env_vars: HashMap<String, String>) -> Result<(), Box<dyn Error>> {
+    let re = regex::Regex::new(KEY_VALIDATION_REGEX).unwrap();
```
danielorihuela: Nit: Use expect.

agent-control/src/utils/env_var.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rstest::rstest;
+    use serial_test::serial;
+
+    #[test]
+    #[serial]
```
danielorihuela: Nit: Instead of serial, we could use the test name as a prefix or something like that.

gsanchezgavier: IIRC the test name determines the order but doesn't prevent the test is executed in parallel with other . I might be wrong but i think the idea was to tag all the test that involves env vars with #serial so they don't collide. Anyway is a safety precussion just in case the collisions happens

PR #1969, opened 2025-12-18
chore: take oci tests into account for check

danielorihuela:
Take oci tests into account for all green check


sigilioso (approved):
Thank you! 💯

PR #1971, opened 2025-12-19
chore: remove systemd env var config generation

gsanchezgavier:
- Removes the .conf file generation.
- Keeps the systemd service env file pointing to the .conf and the migration for backward compatibility.

Note:
- Test are pointing to this [branch](https://github.com/newrelic/open-install-library/pull/1289)
- on top of #1968 

Pending: 
- [ ] some manual test to assure the backward compatibility


.goreleaser.yml
```diff
@@ -147,7 +147,7 @@ nfpms: # deb and rpm are only built for linux targets
           mode: 0600
       - src: build/package/newrelic-agent-control.service
         dst: /lib/systemd/system/newrelic-agent-control.service
-      - src: build/package/newrelic-agent-control.conf
+      - src: build/package/systemd-env.conf
```
danielorihuela: I'm lost with the `systemd-env.conf` and `Commands::FilesBackwardsCompatibilityMigrationFromV120`. I don't think the migration behaviour is correct if we introduce those changes. We are not taking into account the new file. We are not copying the env vars that were used in the old systemd env conf file.

What should we do with that piece we implemented in the past? We implemented as a "training" for us, but I think we should maintain it. At the same time, we already decided that this should go away after some time, but it's vague when that should happen.

So, should we do a new backwards compatible migration step? To support this new change? That would be the most correct path forward. Not sure if we want to pay the price. We can also modify it, but it's less correct. We can also not do it and remove that already, but I don't know we want to do that already.

danielorihuela (commented):
The PR looks great to me. I only have one concern regarding the migration support. Also, I won't approve this PR because it depends on https://github.com/newrelic/open-install-library/pull/1289

.goreleaser.yml
danielorihuela: Ah wait, is this what you meant with

> Pending:
> some manual test to assure the backward compatibility

?

gsanchezgavier: I think i mentally covered them and there were no breaking changes but i might be wrong , i will double check it , thanks for having a deep look at it !

gsanchezgavier: you were right! i missed the part of the post install script! just considered the recipe code. I removed this change to avoid the breaking change , at the end we will probably remove all this by GA to have a clean start 🤞

docs/DEVELOPMENT.md
```diff
@@ -124,11 +124,12 @@ Notice that in windows the paths will be different, since everything will be und
 #### AC Service Windows vs Linux
 
 On Linux, Agent Control is expected to run as a system service, managed by `systemd`. 
-There is a service file at `package/newrelic-agent-control.service` and `package/newrelic-agent-control.conf` that can be used to template when installing AC on a system by the recipe that calls the AC cli command `systemd-config` to inject region, licenseKey and OTEL endpoint.
```
danielorihuela: This comment mentions injecting the region. We are not adding that to the env var. Should we?

agent-control/src/bin/main_agent_control_onhost_cli.rs
```diff
@@ -27,9 +27,7 @@ enum Commands {
     /// It generates either infra-agent or otel configuration.
     /// in case of infra-agent, the proxy configuration is included if provided, and we try to migrate from the old configuration.
     HostMonitoring(host_monitoring_gen::Args),
-    /// Generate systemd configuration according to the provided configuration data.
-    SystemdConfig(systemd_gen::Args),
-    /// Migrate from the older on host folders to the new ones. DON'T USE IT
+    /// Migrates legacy on-host directories (>v1.4.0) to the new layout. Intended to be run by post-installation package scripts only.
```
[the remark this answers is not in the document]
gsanchezgavier: the breaking change was added on 1.4.0 , don't know about the comment but changing that would be hard

sigilioso (approved):
LGTM

PR #1972, opened 2025-12-19
feat(test): e2e-runner linux support - [NR-480764]

sigilioso:
# e2e runner linux support

This PR extends the e2e-runner test framework to support Linux in addition to Windows.

##  Key Changes

- **Linux platform support tools**:  (bash command execution, installing Agent Control via recipe with local deb repository setup, systemd service management, ...
- **Refactor the e2e-runner package**: it supports Windows and LInux, it executes specific test scenarios based on the target OS. Note that `cfg!` instead of `#[cfg]` was used in order to easy local development.
- **Add test scenario for Linux**: infra-agent.
- **Common tools extension**: module to validate NRQL queries, common retry mechanism, utility to automatically display logs on test completion.
- **Use panic! insetead of error**: to provide more information in case of failure.
- **Pipelines integration**: the infra-agent test is already executed in the pipelines although it doesn't replace the Ansible one yet.

## TODO

- [x] ~Check if the `ShowLogsOnDrop` mechanism works well when the execution is interrupted by a failure~ (`panic!`) (works as expected)
- [x] Docs to execute the tests locally

## Out of scope

This will be addressed in a follow-up PR

- [ ] Move the rest of ansible tests implementation
- [ ] Remove previous code and simplify the pipelines

## Checklist

### DIFF
--- a/test/e2e-runner/src/linux/scenarios/infra_agent.rs
+++ b/test/e2e-runner/src/linux/scenarios/infra_agent.rs
@@ -1,11 +1,11 @@
-use std::{thread, time::Duration};
+use std::time::Duration;
 
 use tracing::info;
 
 use crate::{
     linux::{
         self,
-        install::{Args, RecipeData},
+        install::{Args, RecipeData, install_agent_control_from_recipe},
     },
     tools::{config, logs::ShowLogsOnDrop, nrql, test::retry},
 };
@@ -16,7 +16,7 @@ pub fn test_installation_with_infra_agent(args: Args) {
         monitoring_source: "infra-agent".to_string(),
         ..Default::default()
     };
-    linux::install::install_agent_control_from_recipe(&recipe_data);
+    install_agent_control_from_recipe(&recipe_data);
 
     let test_id = format!(
         "onhost-e2e-infra-agent_{}",
@@ -33,8 +33,6 @@ host_id: {test_id}
     );
     config::update_config(linux::DEFAULT_CONFIG_PATH, &config);
 
-    thread::sleep(Duration::from_secs(5));
-
     linux::service::restart_service(linux::SERVICE_NAME);
     let _show_logs = ShowLogsOnDrop::from(linux::DEFAULT_LOG_PATH);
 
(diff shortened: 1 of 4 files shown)


gsanchezgavier (commented):
looks great, haven't check all but i think it will really be easier to maintain than the ansible ones

test/e2e-runner/src/linux/scenarios/infra_agent.rs
```diff
@@ -0,0 +1,53 @@
...
+    config::update_config(linux::DEFAULT_CONFIG_PATH, &config)?;
+
+    thread::sleep(Duration::from_secs(5));
+
+    linux::service::restart_service(linux::SERVICE_NAME)?;
+
+    // TODO: PROPER VALUES
+    let api_endpoint = "";
+    let api_key = "";
+    let account_id = 0;
```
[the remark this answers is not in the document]
sigilioso: I gave it a try but having the globals isn't that simple (it requires something similar to `thread_local!` + `Refcell`) so I don't think it is worth it.

test/e2e-runner/src/linux/scenarios/infra_agent.rs
```diff
@@ -0,0 +1,55 @@
...
+    let debug_log_config = config::debug_logging_config(linux::DEFAULT_LOG_PATH);
+    let config = format!(
+        r#"
+host_id: {test_id}
+{debug_log_config}
+"#
+    );
+    config::update_config(linux::DEFAULT_CONFIG_PATH, &config);
+
+    thread::sleep(Duration::from_secs(5));
```
gsanchezgavier: could we change this for a condition instead? I fear that this will be flaky

test/e2e-runner/src/linux/scenarios/infra_agent.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use crate::{
+    linux::{
+        self,
+        install::{Args, RecipeData},
+    },
+    tools::{config, logs::ShowLogsOnDrop, nrql, test::retry},
+};
+
+pub fn test_installation_with_infra_agent(args: Args) {
```
gsanchezgavier: nice super clear, perhps add some `use` to avoid some ::

gsanchezgavier (approved):
looks great! just left some nits

gsanchezgavier:
oh should we remove the duplicated scenario from ansible??

test/e2e-runner/src/linux/scenarios/infra_agent.rs
sigilioso: I finally moved it inside `config::update_config` as it is just a matter of assuring we are not doing anything else before the config file gets written (Eg: restarting the service). Changed [570724f](https://github.com/newrelic/newrelic-agent-control/pull/1972/commits/570724fdb284829aa6828565048c6b3cc8c0fcd7)

test/e2e-runner/src/linux/scenarios/infra_agent.rs
sigilioso: I've just added some `use` although I kept some `::` when it helps clarifying.

PR #2000, opened 2025-12-31
feat: implement base package manager

DavSanchez:
Implements the minimal package manager functionality. Currently the downloader does not abstract filesystem operations so I did not change it in this implementation, what I did was abstract out the downloader itself behind a trait. In any case I had to modify the downloader. Now it returns the path where the artifact is downloaded so I can manipulate it as part of the package manager operations.

Not sure how feasible it is to generalize the downloader for network and filesystem operations, this might allow us to change the type parameters of the package manager so they are common.

## Checklist

### DIFF
--- a/agent-control/src/package/oci/downloader.rs
+++ b/agent-control/src/package/oci/downloader.rs
@@ -126,7 +126,11 @@ impl OCIRefDownloader {
         package_dir: &Path,
     ) -> Result<Vec<PathBuf>, OCIDownloaderError> {
         for attempt in 0u64..=self.retries {
-            debug!("Downloading artifact, attempt {attempt}/{}", self.retries);
+            debug!(
+                "Downloading artifact, attempt {}/{}",
+                attempt + 1,
+                self.retries + 1
+            );
             let result = self
                 .runtime
                 .block_on(self.oci_download_file(reference, package_dir));
--- a/agent-control/src/package/oci/package_manager.rs
+++ b/agent-control/src/package/oci/package_manager.rs
@@ -128,8 +128,8 @@ where
 /// The sanitization process:
 /// 1. Prepends "oci_" to the filename to avoid reserved filenames (e.g. "CON" on Windows) and
 ///    to prevent the filename from being exactly "." or "..".
-/// 2. Replaces directory separators (`/`, `\`) and the tag separator (`:`) with `__`.
-/// 3. Replaces any other character that is not alphanumeric, `.`, `-`, `_`, or `@` with `_`.
+/// 2. Replaces directory separators (`/`, `\`) with `__`.
+/// 3. Replaces any other character that is not alphanumeric, `.`, `-`, `_`, `@`, etc with `_`.
 pub fn compute_path_suffix(package: &Reference) -> Result<PathBuf, OCIPackageManagerError> {
     let package_full_reference = package.whole();
     let mut safe_name = String::with_capacity(package_full_reference.len() + 4);


agent-control/src/package/oci/package_manager.rs
```diff
@@ -0,0 +1,656 @@
...
+    use std::str::FromStr;
+
+    #[test]
+    fn test_install_success() {
+        let mut downloader = MockOCIDownloader::new();
+        let mut directory_manager = MockDirectoryManager::new();
+        let mut file_renamer = MockLocalFile::new();
+
+        let agent_id = AgentID::try_from("agent-id").unwrap();
+        let reference = Reference::from_str("docker.io/library/busybox:latest@sha256:1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef").unwrap();
```
sigilioso: Even if we are using mocks we need a valid reference and that's why we have this, right? As a nit: could we clarify that the busybox reference isn't actually used? (nothing is downloaded)

agent-control/src/package/oci/downloader.rs
```diff
@@ -19,23 +21,45 @@ use url::Url;
...
+    Io(std::io::Error),
+    #[error("OCI error: {0}")]
+    OciDistribution(OciDistributionError),
     #[error("certificate error: {0}")]
-    CertificateError(String),
+    Certificate(String),
 }
 
-pub struct OCIDownloader {
+pub trait OCIDownloader {
```
danielorihuela: > ... what I did was abstract out the downloader itself behind a trait. ...

> Not sure how feasible it is to generalize the downloader for network and filesystem operations, this might allow us to change the type parameters of the package manager so they are common.

For my understanding. The trait right now is only needed for testing, but you believe that we could need it in the future for different behaviours or is there something I'm not understanding? I'm not sure I understand why should we need a downloader for filesystem operations. Would this be a shortcut or workaround for embedding some stuff?

agent-control/src/package/oci/package_manager.rs
```diff
@@ -0,0 +1,712 @@
...
+            let paths_len = paths.len();
+            Err(OCIPackageManagerError::Install(io::Error::new(
+                io::ErrorKind::InvalidData,
+                format!("expected a single file in the OCI artifact, found {paths_len} files",),
+            )))
+        } else {
+            Ok(paths
+                .into_iter()
+                .next()
+                .expect("checked vector for length above >= 1"))
```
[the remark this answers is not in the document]
DavSanchez: Using `first` returns a reference so would need cloning, and using `paths[0]` kind of hides that is a panicking operation so I wanted to use `expect` directly. Maybe there's another straightforward and efficient `Vec<T> -> T` but besides using `VecDeque` instead I don't know it.

agent-control/src/package/oci/package_manager.rs
```diff
@@ -0,0 +1,723 @@
...
+        package: &Reference,
+    ) -> Result<PathBuf, OCIPackageManagerError> {
+        let digest = package.digest().ok_or_else(|| {
+            OCIPackageManagerError::Install(io::Error::new(
+                io::ErrorKind::InvalidData,
+                "OCI reference missing digest".to_string(),
+            ))
+        })?;
+
+        todo!("return correct type")
```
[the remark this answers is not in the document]
DavSanchez: Yeah this was a TODO given that I had stopped for a bit and resumed work later xD It's done now.

agent-control/src/package/oci/package_manager.rs
```diff
@@ -0,0 +1,723 @@
...
+        &self,
+        agent_id: &AgentID,
+        package: &Reference,
+        download_filepath: &Path,
+    ) -> Result<PathBuf, OCIPackageManagerError> {
+        // Build artifact name
+        let file_name = format!(
+            "{}_{}",
+            package.repository(),
+            package.tag().unwrap_or("latest")
```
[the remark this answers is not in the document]
danielorihuela: Oh, didn't notice they were external. It's okay then. Thanks!

agent-control/src/package/oci/package_manager.rs
```diff
@@ -0,0 +1,656 @@
...
+            .and(installed_package)
+    }
+
+    fn uninstall(
+        &self,
+        _agent_id: &AgentID,
+        _package: Self::InstalledPackage,
+    ) -> Result<(), Self::Error> {
+        todo!("uninstall not implemented yet")
+    }
```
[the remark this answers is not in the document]
DavSanchez: Added

agent-control/src/package/oci/package_manager.rs
```diff
@@ -0,0 +1,712 @@
...
+        match c {
+            '/' | ':' | '\\' => safe.push_str("__"),
+            c if c.is_alphanumeric() || c == '.' || c == '-' || c == '_' || c == '@' => {
+                safe.push(c)
+            }
+            _ => safe.push('_'),
+        }
+    }
+    PathBuf::from(safe)
+}
```
DavSanchez: I have made this decision but I'm not sure if this is what we want. Let's discuss it.

agent-control/src/package/oci/downloader.rs
```diff
@@ -18,24 +20,46 @@ use url::Url;
...
+        reference: &Reference,
+        package_dir: &Path,
+    ) -> Result<Vec<PathBuf>, OCIDownloaderError> {
+        self.download_artifact(reference, package_dir)
+    }
+}
+
 const DEFAULT_RETRIES: u64 = 0;
 
 #[allow(dead_code, reason = "still unused")]
```
danielorihuela: Is the `allow` still needed?

agent-control/src/package/oci/package_manager.rs
```diff
@@ -0,0 +1,722 @@
...
+
+impl<D, DM, FR> PackageManager for OCIPackageManager<D, DM, FR>
+where
+    D: OCIDownloader,
+    DM: DirectoryManager,
+    FR: FileRenamer + FileDeleter,
+{
+    type Error = OCIPackageManagerError;
```
paologallinaharbur: why making the error generic if we are not leveraging this feature more than once?

agent-control/src/package/oci/downloader.rs
```diff
@@ -106,63 +123,55 @@ impl OCIDownloader {
         &self,
         reference: &Reference,
         package_dir: &Path,
-    ) -> Result<(), OCIDownloaderError> {
-        let mut download_error =
-            OCIDownloaderError::DownloadingArtifactError("initial value".into());
+    ) -> Result<Vec<PathBuf>, OCIDownloaderError> {
+        let mut result = Err(OCIDownloaderError::DownloadingArtifact(
+            "no tries attempted".into(),
```
paologallinaharbur: the error IMO is misleading, I am not even sure if sohuld be an error if we do have `self.retries=0`

agent-control/src/package/oci/downloader.rs
```diff
@@ -106,63 +123,55 @@ impl OCIDownloader {
...
-                    return Ok(result);
-                }
-                Err(err) => {
-                    download_error = OCIDownloaderError::DownloadingArtifactError(err.to_string());
-                }
+                .block_on(self.oci_download_file(reference, package_dir));
+
+            if result.is_ok() {
+                break;
+            } else {
```
paologallinaharbur: nit: else clause is not needed since we are breaking in the other case:

```            
if result.is_ok() {
    break;
 } 
sleep(self.retry_interval);
```

agent-control/src/package/oci/package_manager.rs
```diff
@@ -0,0 +1,722 @@
...
+    ///
+    /// Returns the single [`PathBuf`] if validation passes, otherwise returns an error.
+    fn validate_single_path(paths: Vec<PathBuf>) -> Result<PathBuf, OCIPackageManagerError> {
+        if paths.len() != 1 {
+            let paths_len = paths.len();
+            Err(OCIPackageManagerError::Install(io::Error::new(
+                io::ErrorKind::InvalidData,
+                format!("expected a single file in the OCI artifact, found {paths_len} files",),
+            )))
+        } else {
```
paologallinaharbur: the else clause is not needed, since we are doing an early exit

agent-control/src/package/oci/downloader.rs
```diff
@@ -18,24 +20,46 @@ use url::Url;
...
-pub struct OCIDownloader {
+pub trait OCIDownloader {
+    fn download(
+        &self,
+        reference: &Reference,
+        package_dir: &Path,
+    ) -> Result<Vec<PathBuf>, OCIDownloaderError>;
+}
+
+pub struct OCIRefDownloader {
```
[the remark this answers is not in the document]
DavSanchez: This is the same downloader that existed already, but since I added a trait to abstract the downloader from the package manager I changed the names. I can document it anyways.

agent-control/src/package/oci/downloader.rs
DavSanchez: The implementation relied on having a "default error" declared that is mutated before entering the loop. If the loop does not have any iterations should this even return `Ok(_)`?

agent-control/src/package/oci/package_manager.rs
```diff
@@ -0,0 +1,722 @@
...
+pub struct OCIPackageManager<D, DM, FR>
+where
+    D: OCIDownloader,
+    DM: DirectoryManager,
+    FR: FileRenamer + FileDeleter,
+{
+    pub downloader: D,
+    pub directory_manager: DM,
+    pub file_manager: FR,
+    pub base_path: PathBuf, // this would be the `auto-generated` directory
```
paologallinaharbur: can we improve the name? the fact that you had to add a name is a smell IMO 😄

agent-control/src/package/oci/package_manager.rs
```diff
@@ -0,0 +1,722 @@
...
+                .next()
+                .expect("checked vector for length above >= 1"))
+        }
+    }
+
+    /// Moves the downloaded package file from `download_filepath` to its final install location.
+    ///
+    /// This final location is determined from the package [`Reference`]. If the move fails the
+    /// destination directory is deleted.
+    fn install_package(
```
paologallinaharbur: is install_package going to do something more than just moving a file? Otherwise we could rename this

agent-control/src/package/oci/package_manager.rs
```diff
@@ -0,0 +1,722 @@
...
+
+        // The "install" action itself. Moves the downloaded file to its final location.
+        match self
+            .file_manager
+            .rename(download_filepath, &final_file_path)
+        {
+            Err(e) => {
+                warn!("Package intallation failed: {e}");
+                // Install failed, delete the created directory
+                self.directory_manager
```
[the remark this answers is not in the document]
paologallinaharbur: like an older version

agent-control/src/package/oci/package_manager.rs
```diff
@@ -0,0 +1,722 @@
...
+        let temp_download_dir = self
+            .base_path
+            .join(agent_id)
+            .join(DOWNLOADED_PACKAGES_LOCATION)
+            .join(download_path_suffix);
+
+        let download_dir = self
+            .directory_manager
+            .create(&temp_download_dir)
+            .map_err(OCIPackageManagerError::Directory);
```
paologallinaharbur: this is confusing IMO. Why temp_download_dir is used to get download_dir?

agent-control/src/package/oci/package_manager.rs
```diff
@@ -0,0 +1,722 @@
...
+                self.downloader
+                    .download(&package, &temp_download_dir)
+                    .map_err(OCIPackageManagerError::Download)
+            })
+            .and_then(Self::validate_single_path);
+
+        let installed_package = downloaded_pkg
+            .and_then(|file_path| self.install_package(agent_id, &package, &file_path))
+            .inspect(|p| debug!("OCI package installed at {}", p.display()))
+            .inspect_err(|e| warn!("OCI package installation failed: {}", e));
```
paologallinaharbur: why do we need to pass via a temp location? Is it to be future proof? I guess we imagine in the future to do something else to install

agent-control/src/package/oci/downloader.rs
paologallinaharbur: IMO something like this is better (warning PSEUDOCODE)
```rust
    pub fn download_artifact(
        &self,
        reference: &Reference,
        package_dir: &Path,
    ) -> Result<Vec<PathBuf>, OCIDownloaderError> {
        let mut attempt = 0;
        while   while attempt <= self.retries {
            attempt += 1;
            
            debug!("Downloading artifact, attempt {attempt}/{}", self.retries);
            let result = self
                .runtime
                .block_on(self.oci_download_file(reference, package_dir));

            if result.is_ok() || attempt > self.retries{
                return result;
            }
            sleep(self.retry_interval);
        }
            
        Err(OCIDownloaderError::DownloadingArtifact( "Exceeded maximum retry attempts".to_string()))
    }
```

agent-control/src/package/oci/package_manager.rs
DavSanchez: You mean the comment? This was intended for the future integration task to know what to pass, but I guess I can think of a better one.

agent-control/src/package/oci/package_manager.rs
DavSanchez: If we are not planning to have the package manager be something different from OCI I can hardcode the types, but I feel that the package management ability does not have to be tied to the underlying package representation or "manifest" (the OCI reference).

I can see the output type being fixed given this is on-host (the filesystem path or something more elaborate that still contains the path), but I don't quite see the OCI reference being coupled to a generic "I can install things" structure.

agent-control/src/package/oci/package_manager.rs
```diff
@@ -0,0 +1,722 @@
...
+where
+    D: OCIDownloader,
+    DM: DirectoryManager,
+    FR: FileRenamer + FileDeleter,
+{
+    /// Validates that the provided vector of paths, contains exactly one path (i.e. a single file
+    /// was downloaded from the [`OCIDownloader`]).
+    ///
+    /// Returns the single [`PathBuf`] if validation passes, otherwise returns an error.
+    fn validate_single_path(paths: Vec<PathBuf>) -> Result<PathBuf, OCIPackageManagerError> {
```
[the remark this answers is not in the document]
DavSanchez: What about this one? https://github.com/newrelic/newrelic-agent-control/pull/2000/commits/508e84f55467d9230ca9d184059c615ee5c7a1ab

agent-control/src/package/oci/package_manager.rs
DavSanchez: I used the `install` word inspired by this, which mostly copies files (with the possibility of changing attributes like ownership etc): https://www.gnu.org/software/coreutils/manual/html_node/install-invocation.html#install_003a-Copy-files-and-set-attributes

agent-control/src/package/oci/package_manager.rs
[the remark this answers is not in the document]
DavSanchez: You're right, this initially contained a more unique directory containing the file but I removed it. I'll delete the deletion (maybe we can attempt to delete the full file only in case the failure in rename implies the file might be there but in a corrupted state or something...)

Changed in https://github.com/newrelic/newrelic-agent-control/pull/2000/commits/f801da6e93c8f325976c726b0f7f99471556ceaf

agent-control/src/package/oci/package_manager.rs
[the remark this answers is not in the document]
paologallinaharbur: way better!

agent-control/src/package/oci/downloader.rs
```diff
@@ -106,63 +123,55 @@ impl OCIDownloader {
         &self,
         reference: &Reference,
         package_dir: &Path,
-    ) -> Result<(), OCIDownloaderError> {
-        let mut download_error =
-            OCIDownloaderError::DownloadingArtifactError("initial value".into());
+    ) -> Result<Vec<PathBuf>, OCIDownloaderError> {
```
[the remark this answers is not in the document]
DavSanchez: https://github.com/newrelic/newrelic-agent-control/pull/2000/commits/17a007b4d0442e8ed15cd85feee25fbae2c6dad9

agent-control/src/package/oci/package_manager.rs
```diff
@@ -0,0 +1,705 @@
...
+pub struct OCIPackageManager<D, DM, FR>
+where
+    D: OCIDownloader,
+    DM: DirectoryManager,
+    FR: FileRenamer + FileDeleter,
+{
+    pub downloader: D,
+    pub directory_manager: DM,
+    pub file_manager: FR,
+    pub base_path: PathBuf, // this would be the `auto-generated` directory
```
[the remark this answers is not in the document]
DavSanchez: I'm confirming what's the state of the whole `auto-generated` directory on AC and supervisor startup, restarts and so on.

This also applies, but is less critical, on filesystem items (such as the rendered config files for the sub-agents.)

agent-control/src/package/oci/package_manager.rs
```diff
@@ -0,0 +1,705 @@
...
+use tracing::{debug, warn};
+
+use crate::{
+    agent_control::agent_id::AgentID,
+    package::{manager::PackageManager, oci::downloader::OCIRefDownloader},
+};
+
+use super::downloader::{OCIDownloader, OCIDownloaderError};
+
+pub type DefaultOCIPackageManager =
```
danielorihuela: This type is not used. It's supposed to be used later?

DavSanchez: Yeah it was added as a convenient alias for the integration task that @paologallinaharbur is doing. If it's redundant I can remove it!

agent-control/src/package/oci/downloader.rs
```diff
@@ -97,77 +113,75 @@ impl OCIDownloader {
...
-            OCIDownloaderError::DownloadingArtifactError("initial value".into());
-        for attempt in 0..=self.retries {
-            debug!("Downloading artifact, attempt {attempt}/{}", self.retries);
-            match self
+    ) -> Result<Vec<PathBuf>, OCIDownloaderError> {
+        for attempt in 0u64..=self.retries {
+            debug!(
+                "Downloading artifact, attempt {}/{}",
+                attempt + 1,
+                self.retries + 1
```
danielorihuela: But the `self.retries` was okay, because in the loop we go until `=self.retries`. Am I missing something?

PR #2024, opened 2026-01-08
ci: fix k8s canaries

danielorihuela:
We were using the default `secretName` for k8s canaries. That is `agent-control-auth`. However, we use `sys-identity` in the makefile.

```
@kubectl get secret sys-identity --namespace newrelic || \
```

This PR updates the values file.


sigilioso (approved):
Thanks

PR #2028, opened 2026-01-09
ci: fix slack notification

danielorihuela:
Fixes the slack notifications in the pipeline.

https://newrelic.slack.com/archives/C076N1HK7NE/p1767951191259689
https://newrelic.slack.com/archives/C097R8K6GV7/p1767951191586499

Composite actions can't use secrets since these could be in a different repo. They don't inherit secrets. We could pass them as inputs, but I'm not sure they are equivalent to passing them as secrets to a workflow. I went with the workflow option.

## How was it tested?

I modified one of the pipelines in the second commit and executed the workflow manually. Then, I reverted the commit back.

## Checklist

### DIFF
--- a/.github/workflows/puppet_install_test.yml
+++ b/.github/workflows/puppet_install_test.yml
@@ -10,86 +10,86 @@ on:
     - cron: "0 6 * * 1"
 
 jobs:
-  # puppet-install:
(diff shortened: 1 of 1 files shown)


sigilioso (approved):
LGTM

PR #2048, opened 2026-01-12
feat: introduce new traits to represent the supervisor [NR-497494]

sigilioso:
## Summary

This PR introduces new traits for `Supervisor` and introduce the `apply` operation.

## Important notes for reviewers

* The new traits **are not used yet** in order to keep the changes as small as possible. As discussed with @DavSanchez, we will be introducing the new implementations next to the current ones and we will replace the current implementation in a final PR.
* The error definition is as generic as possible on purpose. When including the implementations we can opt for a single definition for errors as it is an easy change. 

## Checklist

- [x] Provided a meaningful title following conventional commit style.
- [x] Included a detailed description for the Pull Request.
- [x] Documentation under `docs` is aligned with the change.
- [x] Follows guidelines for Pull Requests in [`CONTRIBUTING.md`](../blob/main/CONTRIBUTING.md).
- [x] Follows [`log level guidelines`](../blob/main/docs/style/logs.md).

PR #2055, opened 2026-01-14
feat: impl new supervisor traits k8s

DavSanchez:
Implements the new traits defined on https://github.com/newrelic/newrelic-agent-control/pull/2048 for the Kubernetes running mode of AC. The implementation actually stops and destroys the running supervisor and starts a new supervisor under the `apply` method (making it a bit of a misnomer). This is mainly a matter of thread management ergonomics, but this could be reworked if desired (this PR or later iterations) by adding some data sharing.

## Checklist

### DIFF
--- a/agent-control/src/sub_agent/k8s/new_builder.rs
+++ b/agent-control/src/sub_agent/k8s/new_builder.rs
@@ -40,7 +40,6 @@ impl SupervisorBuilder for SupervisorBuilderK8s {
             }
         }
 
-        // Clone the k8s_client on each build.
         Ok(NotStartedSupervisorK8s::new(
             agent_identity.clone(),
             self.k8s_client.clone(),


agent-control/src/utils/thread_context.rs
```diff
@@ -145,6 +146,31 @@ where
...
+    fn stop(self) -> Result<(), ThreadContextStopperError>;
+}
+
+// Blanket implementation for all structures convertible to iterators of started threads
+// without a meaningful result type
+impl<I> ThreadCollectionStopperExt for I
+where
+    I: IntoIterator<Item = StartedThreadContext<()>>,
+{
+    fn stop(self) -> Result<(), ThreadContextStopperError> {
```
danielorihuela: Cool! I think it reads better, but we can improve it a little bit more I think. What about adding the logs directly to the `stop_blocking` function?

```rust
/// It sends a stop signal and waits until the thread handle is joined.
pub fn stop_blocking(self) -> Result<T, ThreadContextStopperError> {
    ...
    self.join_thread()
        .inspect(|_| debug!("Thread {thread_name} stopped"))
        .inspect_err(|e| error!("Error stopping thread {thread_name}: {e}"))
}

fn stop(self) -> Result<(), ThreadContextStopperError> {
    let stopped_threads = self
        .into_iter()
        .map(|ctx| ctx.stop_blocking())
        .collect::<Vec<_>>();

    stopped_threads
        .into_iter()
        // Return first err
        .try_for_each(identity)
}
```

Also, using some variables to differentiate steps might help. But this is purely subjective.

agent-control/src/sub_agent/k8s/supervisor.rs
```diff
@@ -146,7 +154,7 @@ impl NotStartedSupervisorK8s {
             let _guard = span.enter();
 
             // Check and apply k8s objects
-            if let Err(err) = Self::apply_resources(resources.iter(), k8s_client.clone()) {
+            if let Err(err) = StartedSupervisorK8s::apply_resources(resources.iter(), &k8s_client) {
```
danielorihuela: `apply_resources` is in `StartedSupervisorK8s`, but we use it in the non started supervisor. It's a bit confusing.

agent-control/src/sub_agent/k8s/supervisor.rs
```diff
@@ -230,15 +238,25 @@ impl NotStartedSupervisorK8s {
...
         ))
     }
+}
 
+pub struct StartedSupervisorK8s {
+    pub(super) thread_contexts: Vec<StartedThreadContext>,
+    pub(super) agent_identity: AgentIdentity,
+    pub(super) k8s_client: Arc<SyncK8sClient>,
+    pub(super) sub_agent_internal_publisher: EventPublisher<SubAgentInternalEvent>,
+    pub(super) k8s_config: K8s,
```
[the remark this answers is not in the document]
DavSanchez: Might be useful to group some data, yes, particularly the ones that are just passed around between `stopped -> started` and `apply` transitions.

agent-control/src/sub_agent/k8s/supervisor.rs
```diff
@@ -248,31 +266,37 @@ impl NotStartedSupervisorK8s {
...
+            self.agent_identity.clone(),
+            self.k8s_client.clone(),
+            new_k8s_config,
+        );
+
+        let resources = new_starter.build_dynamic_objects()?;
+
+        // I this `apply_resources` safe while the old supervisor is still running?
+        // Maybe this is a matter of moving this to after the old supervisor is stopped?
+        Self::apply_resources(resources.iter(), &self.k8s_client)?;
```
danielorihuela: Not sure I understand why we need a new supervisor in k8s. Won't the apply update the resources, which updated the `Deployment`, which updates the running pods? I'm a bit lost.

DavSanchez: We just had a review. The trait is getting modified so this PR will be modified as well 😅

agent-control/src/sub_agent/k8s/new_builder.rs
```diff
@@ -0,0 +1,153 @@
...
+            let obj_key = (k8s_obj.api_version.as_str(), k8s_obj.kind.as_str());
+            if !supported_set.contains(&obj_key) {
+                return Err(SubAgentBuilderError::UnsupportedK8sObject(format!(
+                    "Unsupported Kubernetes object with api_version '{}' and kind '{}'",
+                    k8s_obj.api_version, k8s_obj.kind
+                )));
+            }
+        }
+
+        // Clone the k8s_client on each build.
```
gsanchezgavier: it is an arc right? if so perhaps i would remove this comment that brings some uncertainty

agent-control/src/utils/thread_context.rs
```diff
@@ -134,14 +135,44 @@ where
...
+        let stopped_threads = self
+            .into_iter()
+            .map(StartedThreadContext::stop_blocking)
+            .collect::<Vec<_>>();
+
+        stopped_threads
+            .into_iter()
+            // Return first err (`identity` acts as a "pass-through" here,
+            // so the first Err interrupts the iteration)
```
gsanchezgavier: 🤔 is it ok to stop , wouldn't that leak threads ?

agent-control/src/sub_agent/k8s/new_builder.rs
DavSanchez: Remove misleading comment https://github.com/newrelic/newrelic-agent-control/pull/2055/commits/c976700967d4e15bf42613d2bd898f4a7fbcded9

PR #2056, opened 2026-01-14
feat: add logic to not download layers if already exists locally

vjripoll:
Add new check to see if the package exists and the metada have the same size, to avoid downloading existing packages.

## Checklist

### DIFF
--- a/agent-control/src/package/oci/package_manager.rs
+++ b/agent-control/src/package/oci/package_manager.rs
@@ -225,6 +225,19 @@ where
             &package_data.id,
             &package_data.oci_reference,
         )?;
+
+        if package_path.exists() {
+            debug!(
+                "Package already installed at {}. Skipping download and extraction.",
+                package_path.display()
+            );
+
+            return Ok(InstalledPackageData {
+                id: package_data.id,
+                installation_path: package_path,
+            });
+        }
+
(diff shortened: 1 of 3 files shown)


agent-control/src/package/oci/downloader.rs
```diff
@@ -147,6 +147,25 @@ impl OCIRefDownloader {
         ))
     }
 
+    async fn should_skip_download(layer_path: &Path, expected_size: i64) -> bool {
```
gsanchezgavier: aren't we deleting this paths after successfully "installed" ?

agent-control/src/package/oci/package_manager.rs
```diff
@@ -232,6 +232,18 @@ where
             &package_data.oci_reference,
         )?;
 
+        if package_path.exists() {
```
danielorihuela: I think that we should move that above `temp_package_path` to avoid leaving that behind. Otherwise, if the package exists we stop here and `.delete(&temp_package_path)` in line 254 is never called.

vjripoll: good catch :) avoid leaving shit everytime

gsanchezgavier (approved):

PR #2063, opened 2026-01-16
ci: test with larger runner

DavSanchez:
DO NOT MERGE


DavSanchez:
Larger runners work without issues. Closing to perform next testing phase.

PR #2070, opened 2026-01-19
feat: update OCI artifact downloader and media type handling (NR-510051)

gsanchezgavier:
- Add artifacts definition to have extra metadata available for processing downloaded artifacts.
- Refactor the downloader so it returns an artifact definition.
- Move extraction logic to the Package definition and remove PackageType definition.
- Avoid pulling blobs in case the manifest is not a valid Agent Pacakge

Out of scope:
- Move oci downloader out of Package module . We could work on that whenever we need to support Agent types packages

### DIFF
--- a/agent-control/src/package/oci/artifact_definitions.rs
+++ b/agent-control/src/package/oci/artifact_definitions.rs
@@ -2,6 +2,7 @@ use std::fmt::Display;
 use std::path::{Path, PathBuf};
 
 use crate::utils::extract::{extract_tar_gz, extract_zip};
+use oci_client::manifest::{OciDescriptor, OciImageManifest};
 
 #[derive(Debug, thiserror::Error)]
 #[error("{0}")]
@@ -9,7 +10,7 @@ pub struct DefinitionError(String);
 
 const AGENT_PACKAGE_ARTIFACT_TYPE: &str = "application/vnd.newrelic.agent.v1";
 const AGENT_TYPE_ARTIFACT_TYPE: &str = "application/vnd.newrelic.agent-type.v1";
-/// OCI manifestartifact types supported.
+/// OCI manifest artifact types supported.
 #[derive(Debug)]
 pub enum ManifestArtifactType {
     AgentPackage,
@@ -80,49 +81,35 @@ impl Display for LayerMediaType {
         }
     }
 }
-/// Represents a OCI artifact locally stored with its complete metadata.
-#[derive(Debug)]
-pub struct LocalArtifact {
-    artifact_type: ManifestArtifactType,
-    blobs: Vec<LocalBlob>,
-}
-impl LocalArtifact {
-    pub fn new(artifact_type: ManifestArtifactType, blobs: Vec<LocalBlob>) -> Self {
-        Self {
-            artifact_type,
-            blobs,
-        }
-    }
-}
 
 #[derive(Debug)]
 pub enum PackageMediaType {
     AgentPackageLayerTarGz,
     AgentPackageLayerZip,
 }
-/// Represents a OCI blob locally stored.
-#[derive(Debug)]
-pub struct LocalBlob {
-    media_type: LayerMediaType,
-    path: PathBuf,
-}
-
-impl LocalBlob {
-    pub fn new(media_type: LayerMediaType, path: PathBuf) -> Self {
-        Self { media_type, path }
+impl Display for PackageMediaType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            PackageMediaType::AgentPackageLayerTarGz => write!(f, "{}", AGENT_PACKAGE_LAYER_TAR_GZ),
+            PackageMediaType::AgentPackageLayerZip => write!(f, "{}", AGENT_PACKAGE_LAYER_ZIP),
+        }
     }
 }
 
 /// Represents a Agent Package OCI artifact locally stored.
-/// Agent Package Manifest requirements:
-/// - artifactType must be '[AGENT_PACKAGE_ARTIFACT_TYPE]'
-/// - exactly one layer with mediaType of '[PackageMediaType]'
 #[derive(Debug)]
 pub struct LocalAgentPackage {
     blob_path: PathBuf,
     blob_media_type: PackageMediaType,
 }
 impl LocalAgentPackage {
+    pub fn new(blob_media_type: PackageMediaType, blob_path: PathBuf) -> Self {
+        Self {
+            blob_media_type,
+            blob_path,
+        }
+    }
+
     /// Extracts the agent package to the specified destination path.
     pub fn extract(&self, dest_path: &Path) -> Result<(), DefinitionError> {
         match &self.blob_media_type {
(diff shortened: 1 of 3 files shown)


agent-control/src/package/oci/artifact_definitions.rs
sigilioso: Maybe we could use doc-comments in this module to document the supported artifact types / media types, it would be a great way of keeping it up to date. I'm not sure if we could use the rust-doc (published in GitHub pages) as doc for this 🤔

agent-control/src/package/oci/artifact_definitions.rs
```diff
@@ -0,0 +1,358 @@
...
+    /// Extracts the agent package to the specified destination path.
+    pub fn extract(&self, dest_path: &Path) -> Result<(), DefinitionError> {
+        let blob = self.package_blob()?;
+
+        match blob.media_type() {
+            MediaType::AgentPackageLayerTarGz => extract_tar_gz(&blob.path, dest_path),
+            MediaType::AgentPackageLayerZip => extract_zip(&blob.path, dest_path),
+            other => Err(DefinitionError(format!(
+                "unsupported media type '{other}' for agent package"
+            )))?,
```
sigilioso: Is this reachable? We are already checking the `MediaType`  in `self.package_blob()`

agent-control/src/package/oci/artifact_definitions.rs
```diff
@@ -0,0 +1,358 @@
...
+        }
+        .map_err(|e| DefinitionError(format!("failed extracting: {e}")))
+    }
+    /// Retrieves the blob that contains the agent package data.
+    fn package_blob(&self) -> Result<&LocalBlob, DefinitionError> {
+        let Some(blob) = self.0.blobs.iter().find(|blob| {
```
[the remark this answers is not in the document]
sigilioso: Oh! I have just seen that it is validated right below!

agent-control/src/package/oci/artifact_definitions.rs
```diff
@@ -0,0 +1,358 @@
...
+                MediaType::AgentPackageLayerTarGz | MediaType::AgentPackageLayerZip
+            )
+        }) else {
+            return Err(DefinitionError("agent package layer missing".to_string()));
+        };
+        Ok(blob)
+    }
+
+    /// Validates that the provided OCI artifact is a valid agent package.
+    fn valid_agent_package(artifact: &LocalArtifact) -> Result<(), DefinitionError> {
```
sigilioso: Shouldn't this be a method of `LocalArtifact`?

```rust
impl LocalArtifact {

    pub fn check(&self) -> Result<(), DefinitionError> { // ...
    }

}
```

Considering that it already iters the `blob`s and check media type, it could even be moved to the `try_from<LocalArtifact> for LocalAgentPackage` and keep a reference to the valid blob ready to be extracted in LocalPackagePackage.

agent-control/src/package/oci/artifact_definitions.rs
```diff
@@ -0,0 +1,358 @@
...
+            AGENT_PACKAGE_LAYER_ZIP => MediaType::AgentPackageLayerZip,
+            AGENT_TYPE_LAYER_TAR_GZ => MediaType::AgentTypeLayerTarGz,
+            other => MediaType::Other(other.to_string()),
+        }
+    }
+}
+
+/// Represents a downloaded OCI artifact with its complete metadata.
+#[derive(Debug)]
+pub struct LocalArtifact {
```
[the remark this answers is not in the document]
sigilioso: I suggested it because of the comment above, not a big deal anyway!

> /// Represents a downloaded OCI artifact with its complete metadata.

agent-control/src/package/oci/artifact_definitions.rs
```diff
@@ -0,0 +1,358 @@
...
+/// Represents a Agent Package OCI artifact that has been downloaded.
+/// Agent Package Manifest requirements:
+/// - artifactType must be '[AGENT_PACKAGE_ARTIFACT_TYPE]'
+/// - at least one layer with mediaType of either
+///   '[AGENT_PACKAGE_LAYER_TAR_GZ]' or '[AGENT_PACKAGE_LAYER_ZIP]'
+#[derive(Debug)]
+pub struct LocalAgentPackage(LocalArtifact);
+impl LocalAgentPackage {
+    /// Extracts the agent package to the specified destination path.
+    pub fn extract(&self, dest_path: &Path) -> Result<(), DefinitionError> {
```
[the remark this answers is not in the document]
sigilioso: I mean:

```rust
pub fn extract(&self, dest_path: &Path) -> Result<(), DefinitionError> {
    let Some(blob) = self.0.blobs.iter().find|blob| blob.can_be_extracted())  else {
        // ...
    };
    blob.extract(dest_path)
}
```

agent-control/src/package/oci/artifact_definitions.rs
gsanchezgavier: yes so the logic of specific packages remains outside LocalArtifact

agent-control/src/package/oci/artifact_definitions.rs
```diff
@@ -0,0 +1,327 @@
...
+        }
+        .map_err(|e| DefinitionError(format!("failed extracting: {e}")))
+    }
+
+    /// Retrieves the descriptor of the layer that contains the agent package data.
+    /// Validates that the manifest conforms to the agent package requirements.
+    /// - artifactType must be '[AGENT_PACKAGE_ARTIFACT_TYPE]'
+    /// - exactly one layer with mediaType of either
+    ///   '[AGENT_PACKAGE_LAYER_TAR_GZ]' or '[AGENT_PACKAGE_LAYER_ZIP]'
+    pub fn get_package_blob(manifest: &OciImageManifest) -> Result<OciDescriptor, DefinitionError> {
```
sigilioso: Would it make sense to receive a reference to the local artifact here and return the localBlob reference?

```rust
pub fn get_package_blob(artifact: &LocalArtifact) -> Result<LocalBlob, DefinitionError> {
    // ...
}
```

Then, when using it in the `try_from` implementation , we could directly use it:

```rust
    fn try_from(artifact: LocalArtifact) -> Result<Self, Self::Error> {
        let blob = Self::get_package_blob(&artifact.manifest)?;
        Ok(Self { blob_path: blob.path().clone, blob_media_type: blob.media_type() })
}
```

agent-control/src/package/oci/artifact_definitions.rs
```diff
@@ -0,0 +1,327 @@
...
+#[derive(Debug, thiserror::Error)]
+#[error("{0}")]
+pub struct DefinitionError(String);
+
+/// OCI artifact types
+pub enum ArtifactType {
+    AgentPackage,
+    AgentType,
+    Other(String),
+}
```
sigilioso: Would it make sense to remove `Other(String)`?

If we are not supporting "Unknown" ArtifactTypes we could fail earlier (as soon as we build the enum). We would need to change the `from<&str>` implementation to a `try_from<&str>`.

agent-control/src/package/oci/artifact_definitions.rs
```diff
@@ -0,0 +1,257 @@
...
+impl LocalBlob {
+    pub fn new(media_type: LayerMediaType, path: PathBuf) -> Self {
+        Self { media_type, path }
+    }
+}
+
+/// Represents a Agent Package OCI artifact locally stored.
+/// Agent Package Manifest requirements:
+/// - artifactType must be '[AGENT_PACKAGE_ARTIFACT_TYPE]'
+/// - exactly one layer with mediaType of '[PackageMediaType]'
```
alvarocabanas: Represents only packages?
How are the AgentType layers? Will they be raw files untar or unzip? 
We still don't have the logic on how to handle them or?

gsanchezgavier: The AgenType will have it's own type with validations, I just added the defined variants in the enums but we will add the full logic whenever we start supporting them. I guess that part of that we will move the downloader out of this module or iterate a bit in the abstractions

PR #2090, opened 2026-01-22
fix(test): avoid collisions when pushing testing artifacts

sigilioso:
The function to create an _unique_ tag when pushing artifacts in the testing OCI repositry now uses a hash of the backtrace information instead of a timestamps.

Collisions in the timestamp where detected when trying to reproduce the issue locally:

```bash
$ for i in `seq 1 50`; do cargo test --package newrelic_agent_control --test integration_tests -- on_host::oci_package_management --nocapture --ignored; done
```


paologallinaharbur (dismissed):
🚀

PR #2094, opened 2026-01-22
fix(oci): push test pipeline

paologallinaharbur:
Before the fix all packages had the wrong metadata attached

### DIFF
--- a/.github/workflows/push_oci_artifact.yml
+++ b/.github/workflows/push_oci_artifact.yml
@@ -85,11 +85,9 @@ jobs:
           if [ "${{ matrix.os }}" = "windows" ]; then
             FILENAME="newrelic-infra-${{ matrix.arch }}.zip"
             zip -r $FILENAME ${{ matrix.binary }} integrations/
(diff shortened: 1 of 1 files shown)


.github/workflows/push_oci_artifact.yml
```diff
@@ -114,10 +114,10 @@ jobs:
             --artifact-type application/vnd.newrelic.agent.v1 \
             --artifact-platform ${{ matrix.os }}/${{ matrix.arch }} \
             --annotation-file layer_annotations.json \
-            ${{ steps.package.outputs.filename }}:application/vnd.newrelic.agent.v1.${{ matrix.format }} | grep "Digest:" | awk '{print $2}')
+            ${{ steps.package.outputs.filename }}:application/vnd.newrelic.agent.content.v1.${{ matrix.format }} | grep "Digest:" | awk '{print $2}')
```
paologallinaharbur: I am not sure why a file lines above we were setting `mime=[...]` @alvarocabanas

PR #2108, opened 2026-01-27
fix(k8s): remove Instrumentation CRs before anything else

DavSanchez:
Provides a fix for the issue where an installation of AC that runs the K8s Operator creates Instrumentation CRs that cannot be deleted afterwards.

The fix consists on handling the Instrumentation CRs as special citizens that require their deletion first across namespaces. Only after the Instrumentation CRs have been deleted we proceed with the remaining resources.

## Before

1. Removed AC CRs.
2. Removed objects from AC's namespace.
    1. Removed the Operator HelmRelease.
    2. Removed the Instrumentation CRD.
    3. Removed everything else.
4. Removed the objects from the Agents namespace.
    1. Removed the Instrumentation CR. ⚠️ This was failing because the CRD was removed earlier and the resource API does not exist anymore. The `Deleter` entered the retry loop and expired the uninstallation time (we might want to fine-tune this as well).
    2. Removed everything else.

## After

1. Removes AC CRs.
2. Removed Instrumentation CRs from AC's namespace.
3. Removes Instrumentation CRs from the Agents namespace.
4. Removes remaining objects from AC's namespace.
    1. The Operator HelmRelease, along with the Instrumentation CRD it introduces, are removed here after all Instrumenatation CRs are gone 🟢.
5. Removes remaining objects from the Agents namespace.

### DIFF
--- a/agent-control/src/cli/k8s/uninstall/agent_control.rs
+++ b/agent-control/src/cli/k8s/uninstall/agent_control.rs
@@ -41,14 +41,10 @@ pub fn uninstall_agent_control(
     // We filter the static list of objects we want to delete against what is actually available in the cluster.
     let valid_objects_to_delete = objects_to_delete(&kinds_available);
 
-    // We need to handle the Instrumentation resources separately because we order the resource
-    // deletion like this:
-    // 1. Owned objects on the AC namespace. This deletes the Operator HelmRelease first, which
-    //    ends up removing the Instrumentation CRD.
-    // 2. Owned objects on the Agents namespace, which includes the Instrumentation CR that we
-    //    had listed before the deletions started.
-    // 3. The Instrumentation deletion attempts, coming after the Operator is out, will fail as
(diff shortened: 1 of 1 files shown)


agent-control/src/cli/k8s/uninstall/agent_control.rs
```diff
@@ -70,7 +71,26 @@ fn objects_to_delete(kinds_available: &HashSet<TypeMeta>) -> Vec<TypeMeta> {
     });
 
     tm_to_delete.retain(|tm| kinds_available.contains(tm));
-    tm_to_delete
+
+    // Introduce an ad-hoc ordering to ensure:
+    // 1. Deterministic results
+    // 2. Instrumentation CRs go first (i.e. before the K8s operator if present)
+    tm_to_delete.sort_by(instrumentations_first);
```
paologallinaharbur: ugly but effective 😅

PR #2114, opened 2026-01-28
fix: oci registry protocol with debug

paologallinaharbur:
This allows to do `cargo run ...` without specifying --release


sigilioso (approved):
Thanks!

Should we mention it in `docs/DEVELOPMENT.md` or `docs/oci_repository.md`?

It could be useful if e2e-debugging with a local oci repository... 🤔

PR #2116, opened 2026-01-28
Feat/environment variables windows

paologallinaharbur:
As discussed this PR introduces the environment variable identified by @alvarocabanas in order to make sure that we "control" where the infrastructure agents stores data and looks for binary

On top of that we are moving around the config file to mimic the default of the infrastructure agent


agent-control/agent-type-registry/newrelic/com.newrelic.infrastructure-0.1.0.yaml
```diff
@@ -175,18 +175,24 @@ deployment:
         - --version
       regex: \d+\.\d+\.\d+
     filesystem:
-      config:
-        newrelic-infra.yaml: |
-          ${nr-var:config_agent}
       integrations.d: ${nr-var:config_integrations}
       logging.d: ${nr-var:config_logging}
+      newrelic-infra:
+        newrelic-infra.yml: |
```
danielorihuela: Why change it to `yml`? I see we use `yaml` everywhere. I think sticking to one file extension is easier for the user.

If you prefer `yml`, we should do another PR changing every `yaml` file extension to `yml`

gsanchezgavier (approved):
LGTM

PR #2121, opened 2026-01-30
feat: modify teardown windows and panics

alvarocabanas:
Windows Service is not capable of stop or restart when the newrelic-agent-control process is crashing due to a panic, config error or something.
The issue is we were not handling the stop communication to Windows on time because the handler receives a stop and tells our producer to gracefull shutdown (or it has already done it) but windows finds the process irresponsive.
The modification communicates the stop and emits a StopPending that gives a graceful time of 10 seconds for the service to stop.
Panics are also captured and work properly.

### DIFF
--- a/agent-control/src/command/windows.rs
+++ b/agent-control/src/command/windows.rs
@@ -4,37 +4,86 @@
 use crate::event::ApplicationEvent;
 use crate::event::channel::EventPublisher;
 use std::error::Error;
+use std::sync::OnceLock;
 use tracing::error;
 use windows_service::{
     service::{
         ServiceControl, ServiceControlAccept, ServiceExitCode, ServiceState, ServiceStatus,
         ServiceType,
     },
-    service_control_handler::{self, ServiceControlHandlerResult},
+    service_control_handler::{self, ServiceControlHandlerResult, ServiceStatusHandle},
 };
+
+/// Global handle used by the event handler to signal state changes (like StopPending) to Windows.
+/// This allows the closure to access the handle, which is only available after registration.
+pub static WINDOWS_SERVICE_HANDLE: OnceLock<ServiceStatusHandle> = OnceLock::new();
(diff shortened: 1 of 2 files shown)


agent-control/src/command/windows.rs
```diff
@@ -4,36 +4,94 @@
...
+/// Type alias for the Windows service teardown closure to reduce type complexity.
+pub type WinServiceTeardown = Box<dyn Fn(Result<(), &Box<dyn Error>>) -> WinServiceResult>;
+
+/// WinServiceSetup contains the function to call teardown and the panic handler
+pub struct WinServiceSetup {
+    /// A guard that ensures the service reports a stopped state to Windows if the process panics.
+    pub panic_handler: PanicStatusHandler,
+    /// A closure called at the end of execution to transition the service to the stopped state.
+    pub teardown: WinServiceTeardown,
+}
```
DavSanchez: Not that it matters much, but couldn't we just be generic over `Fn(WinServiceResult) -> WinServiceResult` instead? It'd simplify things a bit, not sure what's the use case for boxing here. Also not sure why one of the types uses a `&Box<dyn Error>` and the other the owned value.

If the teardown function only _inspects_ the previous result without needing to consume anything, I'd take a reference to it instead.

```suggestion
/// WinServiceSetup contains the function to call teardown and the panic handler
pub struct WinServiceSetup<F>
  where F: Fn(&WinServiceResult) -> WinServiceResult,
  {
    /// A guard that ensures the service reports a stopped state to Windows if the process panics.
    pub panic_handler: PanicStatusHandler,
    /// A closure called at the end of execution to transition the service to the stopped state.
    pub teardown: F,
}
```

sigilioso (commented):
Thanks for taking care of this, it was an important bug concerning the Windows services. Some comments bellow:

PR #2122, opened 2026-01-30
fix: use TerminateJobObject

gsanchezgavier:
Even that closing the handle was making the JobObject to terminate we encounter errors 
`2026-01-30T09:19:29 ERROR process_fleet_event{id: nr-infra}: Fail to kill a JobObject, err: The handle is invalid. (0x80070006)`


DavSanchez (approved):
Thanks! 🚀

DavSanchez (approved):

PR #2129, opened 2026-02-03
feat(filesystem): support empty dirs

paologallinaharbur:
The main reason of the change is to support the creation of empty dirs.
To do so I slightly refactored the code in order to avoid a middle struct that was a list of files.
Moreover there was no need to keep the enum after templating

### DIFF
--- a/agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
+++ b/agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
@@ -1,6 +1,7 @@
 use crate::agent_type::runtime_config::on_host::filesystem::{DirEntriesMap, SafePath};
 use ::fs::{directory_manager::DirectoryManager, file::writer::FileWriter};
-use std::{collections::HashMap, io};
+use std::collections::HashMap;
+use std::path::PathBuf;
 use thiserror::Error;
 use tracing::trace;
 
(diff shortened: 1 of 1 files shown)


agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -1,63 +1,51 @@
...
+    file_writer: &impl FileWriter,
+    dir_manager: &impl DirectoryManager,
+    path: &SafePath,
+    dir_entries: &DirEntriesMap,
+) -> Result<(), FileSystemEntriesError> {
+    dir_entries
+        .0
+        .iter()
+        .try_for_each(|(sub_path, file_content)| {
+            let file_path = path.as_ref().join(sub_path);
```
DavSanchez: This is a very minor stylistic choice, but I'd make this function be `create_file` with only the logic that is inside the `try_for_each`, so the two iterations are in `write`, otherwise when creating an empty dir we call this function from `write` but the function does nothing.

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -69,34 +57,3 @@ impl From<io::Error> for FileSystemEntriesError {
         FileSystemEntriesError(value)
     }
 }
```
DavSanchez: I just remembered that you prefer to not have implicit conversions for errors! Just in case you want to get rid of this.

```suggestion
```

DavSanchez (approved):
Thanks! Nice redundancy cleanup as well 🚀

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -1,102 +1,54 @@
...
+    })?;
+    // Will overwrite files if they already exist!
+    file_writer
+        .write(file_path.as_path(), file_content.to_owned())
+        .map_err(|err| FileSystemEntriesError(format!("creating file {file_path:?}: {err}")))
+}
+
+#[derive(Debug, Error)]
+#[error("file system entries error: {0}")]
+pub struct FileSystemEntriesError(String);
```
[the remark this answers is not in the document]
paologallinaharbur: discussed online

PR #2152, opened 2026-02-05
feat: improve oci logging

alvarocabanas:
Give a better explanation for some of the logs when downloading the manifest or the blob from the oci repository.

We can't be more granular, because each error can act very differently on different repositories all depending on the server responses, for example, a wrong repository name in github gives unauthorized and a wrong registry or wrong headers both give the same error of not reachable.

We capture the error for wrong version though, that will probably the most frequent.

### DIFF
--- a/agent-control/src/package/oci/downloader.rs
+++ b/agent-control/src/package/oci/downloader.rs
@@ -25,7 +25,7 @@ pub enum OCIDownloaderError {
     DownloadingArtifact(String),
     #[error("I/O error: {0}")]
     Io(std::io::Error),
-    #[error("Registry error: {0}")]
+    #[error("registry error: {0}")]
     RegistryError(String),
     #[error("the registry and repository are not found or reachable: {0}")]
     ConnectionError(String),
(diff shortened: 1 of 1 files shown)


agent-control/src/package/oci/downloader.rs
```diff
@@ -25,12 +25,49 @@ pub enum OCIDownloaderError {
...
-    OciDistribution(OciDistributionError),
+    #[error("Registry error: {0}")]
+    RegistryError(String),
+    #[error("the registry and repository are not found or reachable: {0}")]
+    ConnectionError(String),
     #[error("certificate error: {0}")]
     Certificate(String),
 }
 
+impl OCIDownloaderError {
```
sigilioso: Is it possible to implement `From` directly?

```rust
impl From<OciDistributionError> for OciDownloaderError {
}
```

sigilioso (commented):
Looks good! Find some _nits_ below 🙂

agent-control/src/package/oci/downloader.rs
```diff
@@ -516,6 +548,49 @@ pub mod tests {
...
+        let oci_err = OciDistributionError::AuthenticationFailure("bad login".to_string());
+        let err = OCIDownloaderError::from(oci_err);
+
+        match err {
+            OCIDownloaderError::ConnectionError(msg) => {
+                assert!(msg.contains("bad login"));
+            }
+            _ => panic!("Expected ConnectionError variant for non-registry errors"),
+        }
+    }
```
sigilioso: Nit: this can be a case in the rstest-test above, right?

sigilioso:
BTW, we need to update the base branch to `freeze-develop` 🥶

agent-control/src/package/oci/downloader.rs
alvarocabanas: So i can't do the check for match err (I changed in last commit.)

danielorihuela (approved):

PR #2156, opened 2026-02-06
refactor: generalize oci-client [NR-515717]

sigilioso:
## Summary

This PR adds a new `oci` module containing a thing wrapper of the OCI client. This will help us reusing such client when implementing the AgentType download and will allow us implementing the signature verification features in that common client.

## Notes for reviewers

- The proxy setup is in its own private module (`oci::proxy`).
- Because `Client` can be too general (and we already have several `Client` structs) the references keep the module name: `let client = oci::Client` instead of `let client = Client`.
- The FakeOciServer is also moved to the `oci` module so it can also be reused.
- No additional abstractions are created for now (no new trait representing the `oci::Client`): the components using the client still need to interact with a lot of objects coming from the `oci_client` dependency (manifests, layers, media_types, ...) and an additional abstraction doesn't bring any value at this stage.

### DIFF
--- a/agent-control/src/oci/error.rs
+++ b/agent-control/src/oci/error.rs
@@ -47,10 +47,10 @@ impl From<OciDistributionError> for OciErrorMessage {
                     }
                     _ => format!("registry error ({:?}): {}", oci_err.code, oci_err.message),
                 };
-                format!("the registry and repository are not found or reachable: {err_msg}").into()
+                format!("registry error: {err_msg}").into()
             }
             // Use _ to catch all other variants like AuthenticationFailure, etc.
-            _ => format!("registry error: {err}").into(),
+            _ => format!("the registry and repository are not found or reachable: {err}").into(),
         }
     }
 }


agent-control/src/agent_control/run/on_host.rs
```diff
@@ -173,10 +174,10 @@ impl AgentControlRunner {
             ..Default::default()
         };
 
-        let packages_downloader =
-            OCIArtifactDownloader::try_new(self.proxy, self.runtime, oci_client_config).map_err(
-                |err| RunError(format!("failed to create OCIRefDownloader client: {err}")),
-            )?;
+        let oci_client = oci::Client::try_new(self.proxy, oci_client_config)
```
danielorihuela: > the references keep the module name: let client = oci::Client

Why not call it `OciClient`? Like with `HttpClient`?

danielorihuela: Nit: Change arguments order to client config and proxy config. Like in `setup_proxy`. Order of importance and consistency. It's irrelevant really.

agent-control/src/oci/proxy.rs
```diff
@@ -0,0 +1,248 @@
...
+
+use crate::http::{
+    client::{HttpBuildError, cert_paths_from_dir, certificate_error},
+    config::ProxyConfig,
+};
+
+use super::OciClientError;
+
+/// Returns a [ClientConfig] corresponding to `client_config` with the provided `proxy_config` applied.
+pub(super) fn setup_proxy(
```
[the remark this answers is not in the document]
sigilioso: Exactly! I kept what we had. We merely have a `try_new`

agent-control/src/oci.rs
```diff
@@ -0,0 +1,211 @@
...
+    ) -> Result<Self, OciClientError> {
+        let client_config = proxy::setup_proxy(client_config, proxy_config)?;
+        Ok(Self {
+            client: oci_client::Client::new(client_config),
+            auth: RegistryAuth::Anonymous,
+        })
+    }
+
+    /// Sets up the provided authentication.
+    pub fn with_auth(self, auth: RegistryAuth) -> Self {
```
gsanchezgavier: looks like dead code right?

agent-control/src/oci/error.rs
```diff
@@ -0,0 +1,102 @@
...
+                    }
+                    OciErrorCode::Toomanyrequests => {
+                        format!("rate limit exceeded: the registry is throttling requests: {err}")
+                    }
+                    _ => format!("registry error ({:?}): {}", oci_err.code, oci_err.message),
+                };
+                format!("the registry and repository are not found or reachable: {err_msg}").into()
+            }
+            // Use _ to catch all other variants like AuthenticationFailure, etc.
+            _ => format!("registry error: {err}").into(),
```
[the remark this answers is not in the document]
sigilioso: Great catch! I think I messed up in the refactor, I've just switched them back. Thanks!

PR #2157, opened 2026-02-06
refactor: move signature verification logic to its own module

gsanchezgavier:
- moves out common signature abstractions so they can be reused by the oci signature verifier.

### DIFF
--- a/agent-control/src/signature/public_key_fetcher.rs
+++ b/agent-control/src/signature/public_key_fetcher.rs
@@ -53,11 +53,16 @@ impl PublicKeyFetcher {
         }
 
         if keys.is_empty() {
-            return Err(PubKeyFetcherError(format!(
-                "no valid keys found ({} invalid keys: {})",
-                errors.len(),
-                errors.join(", ")
-            )));
+            let error_msg = if errors.is_empty() {
+                "no keys found".to_string()
+            } else {
+                format!(
+                    "no valid keys found ({} invalid keys: {})",
+                    errors.len(),
+                    errors.join(", ")
+                )
+            };
+            return Err(PubKeyFetcherError(error_msg));
         }
 
         Ok(keys)
@@ -344,7 +349,7 @@ pub mod tests {
         );
         let err = result.unwrap_err();
         assert!(
-            err.0.contains("no valid keys found"),
+            err.0.contains("no keys found"),
             "Error should indicate no valid keys were found, got: {}",
             err.0
         );


agent-control/src/opamp/remote_config/validators/signature/verifier.rs
```diff
@@ -18,6 +21,35 @@ pub trait Verifier {
     fn key_id(&self) -> &str;
 }
 
+impl Verifier for PublicKey {
```
danielorihuela: Have you thought about implementing this trait to the object you want to verify?

For example, `impl Verifier for RemoteConfig`. This way we could decouple verifying the signature itself from preparing the data to verify. We could put the preparation logic in the `verify_signature` implemented in `RemoteConfig` and pass the public key as input. We could the same later for `Package` or `AgentType`. In the `PublicKey` we still have the `verify_signature` method that only cares about checking the maths.

agent-control/src/signature/public_key.rs
```diff
@@ -0,0 +1,220 @@
...
+    }
+}
+
+impl AsRef<str> for SigningAlgorithm {
+    fn as_ref(&self) -> &str {
+        match self {
+            SigningAlgorithm::ED25519 => ED25519,
+        }
+    }
+}
```
danielorihuela: Nit: We can get rid of the constant and the `match` by using `Display` https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=5010f91ba31b2b108fe984bbe7a39e21.

```
impl Display for SigningAlgorithm {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(f, "{:?}", self)
    }
}
```

The debug implementation returns the variant as a string by default (if it doesn't have data stored in it). We avoid the extra constant and having to update that if we ever add more algorithms. It's not important. Just a nit.

danielorihuela (approved):
Great job!!! I like it. From my side, we can merge it already.

agent-control/src/signature/public_key_fetcher.rs
```diff
@@ -0,0 +1,399 @@
...
+        assert!(
+            result.is_err(),
+            "Expected an error when keys array is empty"
+        );
+        let err = result.unwrap_err();
+        assert!(
+            err.0.contains("no valid keys found"),
+            "Error should indicate no valid keys were found, got: {}",
+            err.0
+        );
```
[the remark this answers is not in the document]
gsanchezgavier: done

PR #2168, opened 2026-02-10
feat(windows): enable log to files by default

gsanchezgavier:
Enables logs to file by default in windows for AC and otel.
For infra-agent, it disables the feature to write the logs to a file as they are already being written by the infra agent itself (without the possibility to disable this)

### DIFF
--- a/agent-control/tests/on_host/agent_control_cli.rs
+++ b/agent-control/tests/on_host/agent_control_cli.rs
@@ -33,9 +33,8 @@ proxy:
   url: https://some.proxy.url/
   ca_bundle_dir: /test/bundle/dir
   ignore_system_proxy: true
-{}
+{LOG_SECTION}
     "#,
-        log_section()
     );
     let expected_value: serde_yaml::Value = serde_yaml::from_str(&expected_yaml).unwrap();
     let actual_content = std::fs::read_to_string(&path).unwrap();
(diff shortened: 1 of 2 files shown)


agent-control/tests/on_host/agent_control_cli.rs
```diff
@@ -154,3 +160,15 @@ NEW_RELIC_LICENSE_KEY: fake_license
...
+fn log_section() -> &'static str {
+    if cfg!(target_family = "windows") {
+        r#"
+log:
+  file:
+    enabled: true"#
+    } else {
+        ""
+    }
+}
```
DavSanchez: I'd use `const`s with the same names for these values, one for `target_family = "windows"` and the other for `target_family = "unix"`. Leaving the `const` undefined otherwise.

agent-control/agent-type-registry/newrelic/com.newrelic.infrastructure-0.1.0.yaml
```diff
@@ -23,11 +23,6 @@ variables:
       type: string
       required: false
       default: 20s
-    enable_file_logging:
```
paologallinaharbur: why disableing only for windows? isn't behaving in the same way in linux?

paologallinaharbur: I guess no, given the comments you left 😄

paologallinaharbur (approved):
Thanks for taking care of this. Are we going to change as well the location of AC logs? Currently it follows a different file structure than the subAgents and I am not sure if it makes sense to make that uniform 🤔

PR #2190, opened 2026-02-12
feat: add helpers for oci signatures validation

vjripoll:
This PR implements OCI image signature verification following the Cosign Simple Signing specification. It refactors oci::Client to inject a PublicKeyFetcher for retrieving keys via JWKS. 
The new logic is structured around key helper functions: triangulate() resolves the signature tag based on the manifest digest, trusted_signature_layers() retrieves the specific payload layers, and verify_signatures() performs the Ed25519 cryptographic check while enforcing replay attack protection. The high-level verify() method orchestrates this entire flow to return a secure, pinned reference.
## Checklist
 -->

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -312,9 +312,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.100"
+version = "1.0.101"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a23eb6b1614318a8071c9b2521f36b424b2c83db5eb3a0fead4a6c0809af6e61"
+checksum = "5f0e0fee31ef5ed1ba1316088939cea399010ed7731dba877ed44aeb407a75ea"
 
 [[package]]
 name = "arraydeque"
@@ -1189,9 +1189,9 @@ dependencies = [
 
 [[package]]
 name = "deranged"
-version = "0.5.5"
+version = "0.5.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ececcb659e7ba858fb4f10388c250a7252eb0a27373f1a72b8748afdd248e587"
+checksum = "cc3dc5ad92c2e2d1c193bbbbdf2ea477cb81331de4f3103f267ca18368b988c4"
 dependencies = [
  "powerfmt",
 ]
@@ -2060,14 +2060,13 @@ dependencies = [
 
 [[package]]
 name = "hyper-util"
-version = "0.1.19"
+version = "0.1.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "727805d60e7938b76b826a6ef209eb70eaa1812794f9424d4a4e2d740662df5f"
+checksum = "96547c2556ec9d12fb1578c4eaf448b04993e7fb79cbaad930a656880a6bdfa0"
 dependencies = [
  "base64 0.22.1",
  "bytes",
  "futures-channel",
- "futures-core",
  "futures-util",
  "http 1.4.0",
  "http-body",
@@ -2288,9 +2287,9 @@ checksum = "92ecc6618181def0457392ccd0ee51198e065e016d1d527a7ac1b6dc7c1f09d2"
 
 [[package]]
 name = "jiff"
-version = "0.2.18"
+version = "0.2.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e67e8da4c49d6d9909fe03361f9b620f58898859f5c7aded68351e85e71ecf50"
+checksum = "d89a5b5e10d5a9ad6e5d1f4bd58225f655d6fe9767575a5e8ac5a6fe64e04495"
 dependencies = [
  "jiff-static",
  "log",
@@ -2301,9 +2300,9 @@ dependencies = [
 
 [[package]]
 name = "jiff-static"
-version = "0.2.18"
+version = "0.2.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e0c84ee7f197eca9a86c6fd6cb771e55eb991632f15f2bc3ca6ec838929e6e78"
+checksum = "ff7a39c8862fc1369215ccf0a8f12dd4598c7f6484704359f0351bd617034dbf"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -2695,9 +2694,9 @@ checksum = "47e1ffaa40ddd1f3ed91f717a33c8c0ee23fff369e3aa8772b9605cc1d22f4c3"
 
 [[package]]
 name = "memchr"
-version = "2.7.6"
+version = "2.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f52b00d39961fc5b2736ea853c9cc86238e165017a493d1d5c8eac6bdc4cc273"
+checksum = "f8ca58f447f06ed17d5fc4043ce1b10dd205e060fb3ce5b979b8ed8e59ff3f79"
 
 [[package]]
 name = "mime"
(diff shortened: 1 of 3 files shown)


agent-control/src/oci.rs
```diff
@@ -65,134 +75,478 @@ impl Client {
...
+        reference: &Reference,
+        url: &Url,
+    ) -> Result<Reference, OciClientError> {
+        debug!("Fetching public keys from {}", url);
+
+        let fetcher = self.key_fetcher.clone();
+        let url = url.clone();
+
+        // Fetches the latest trusted public key from the JWKS endpoint.
+        let trusted_key = tokio::task::spawn_blocking(move || fetcher.fetch_latest_key(&url))
```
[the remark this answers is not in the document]
gsanchezgavier: I think we are going to verify with all keys

agent-control/src/oci.rs
```diff
@@ -65,134 +75,478 @@ impl Client {
...
+
+        let new_ref_str = format!(
+            "{}/{}:{}",
+            reference.registry(),
+            reference.repository(),
+            signature_tag
+        );
+
+        Reference::try_from(new_ref_str)
+            .map_err(|e| OciClientError::InvalidReference(e.to_string()))
```
danielorihuela: What about?

```rust
let signature_tag = format!("{}.sig", digest.replace(':', "-"));
Ok(Reference::with_tag(
    reference.registry().to_string(),
    reference.repository().to_string(),
    signature_tag,
))
```

agent-control/src/oci.rs
```diff
@@ -65,134 +75,478 @@ impl Client {
...
+        }
+
+        Err(OciClientError::Verify(format!(
+            "Verification failed. Checked {} candidates, but no valid signature found for digest {}",
+            checked_count, expected_image_digest
+        )))
+    }
+
+    /// Determines the reference for the signature image based on the target image digest.
+    fn triangulate(
```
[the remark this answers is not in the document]
vjripoll: added

agent-control/src/oci/error.rs
```diff
@@ -11,6 +11,12 @@ pub enum OciClientError {
     PullManifest(OciErrorMessage),
     #[error("failure pulling blob: {0}")]
     PullBlob(OciErrorMessage),
+    #[error("invalid reference format generated: {0}")]
+    InvalidReference(String),
+    #[error("image signature verification failed: {0}")]
+    Verify(String),
+    #[error("Failed to fetch public key: {0}")]
```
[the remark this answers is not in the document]
vjripoll: changed

agent-control/src/oci.rs
```diff
@@ -65,134 +60,282 @@ impl Client {
...
+        let result = client.verify(&image_ref, &[trusted_kp.public_key()]).await;
+
+        assert!(result.is_err(), "Verification should fail with wrong key");
+
+        let err_msg = result.unwrap_err().to_string();
+        assert!(
+            err_msg.contains("Verification failed"),
+            "Unexpected error: {}",
+            err_msg
+        );
```
danielorihuela: Nit: Maybe we can add the extra key and assertions to the `test_verify_full_flow_integration` test and avoid copying the setup again. Maybe it's easier to understand like that. I don't know.

agent-control/src/oci/signature_verification.rs
```diff
@@ -0,0 +1,294 @@
...
+                    debug!("Skipping layer with invalid base64 signature: {}", e);
+                    continue;
+                }
+            };
+
+        checked_count += 1;
+
+        for key in trusted_keys {
+            match key.verify_signature(
+                &SigningAlgorithm::ED25519,
```
danielorihuela: Nit: Not sure this should be hardcoded. We can get that from the `crv` field in the JWKS.

@gsanchezgavier, do you know how likely is that we need to support other types of keys?
The result will be an error anyways if we use the `SigningAlgorithm::try_from`. This would make it easier in the future to add new types of keys, if we ever need to.

agent-control/src/oci/signature_verification.rs
```diff
@@ -0,0 +1,312 @@
...
+                }
+            };
+
+        checked_count += 1;
+
+        for key in trusted_keys {
+            match key.verify_signature(&layer.raw_data, &signature_bytes) {
+                Ok(_) => {
+                    info!(
+                        "Valid signature found in layer {} and verified via OCI using key {}!",
```
gsanchezgavier: i would move this to debug (it has too much information IMO) and perhaps if make sese, print a similar log on the top layers, for instance whenever the main verify functions is called and just print that the package (reference) has a valid signature

agent-control/src/oci/signature_verification.rs
```diff
@@ -0,0 +1,312 @@
...
+                    debug!("Verification failed against key {}: {}", key.key_id(), e);
+                }
+            }
+        }
+    }
+
+    Err(OciClientError::Verify(format!(
+        "Verification failed. Checked {} candidates, but no valid signature found for digest {}",
+        checked_count, expected_image_digest
+    )))
```
sigilioso:
```suggestion
    Err(OciClientError::Verify(format!(
        "verification failed. Checked with {} public keys, but no valid signature found for digest {}",
        checked_count, expected_image_digest
    )))
```

agent-control/src/oci/signature_verification.rs
```diff
@@ -0,0 +1,312 @@
...
+                Ok(_) => {
+                    info!(
+                        "Valid signature found in layer {} and verified via OCI using key {}!",
+                        layer.oci_digest,
+                        key.key_id()
+                    );
+                    return Ok(());
+                }
+                Err(e) => {
+                    debug!("Verification failed against key {}: {}", key.key_id(), e);
```
sigilioso:
```suggestion
                    debug!(key_id = key.key_id(), "Verification failed against {e}");
```

agent-control/src/oci/signature_verification.rs
```diff
@@ -0,0 +1,312 @@
...
+        checked_count += 1;
+
+        for key in trusted_keys {
+            match key.verify_signature(&layer.raw_data, &signature_bytes) {
+                Ok(_) => {
+                    info!(
+                        "Valid signature found in layer {} and verified via OCI using key {}!",
+                        layer.oci_digest,
+                        key.key_id()
+                    );
```
sigilioso: I'd use log attributes and probably debug instead of info

```suggestion
                    debug!(layer=layer.oci_digest, key=key.key_id(), "Signature successfully verified");
```

agent-control/src/oci/signature_verification.rs
```diff
@@ -0,0 +1,312 @@
...
+/// * **Output Reference**: `registry.io/my-app:sha256-9f86....sig`
+pub fn triangulate(reference: &Reference, digest: &str) -> Result<Reference, OciClientError> {
+    let signature_tag = format!("{}.sig", digest.replace(':', "-"));
+
+    Ok(Reference::with_tag(
+        reference.registry().to_string(),
+        reference.repository().to_string(),
+        signature_tag,
+    ))
+}
```
sigilioso: The method doesn't fail, doesn't it? Could we return the `Reference` directly instead of the current `Result`?

Cargo.lock
sigilioso: Is this file expected to change or was committed by mistake?

agent-control/src/oci/signature_verification.rs
```diff
@@ -0,0 +1,312 @@
...
+    trusted_keys: &[PublicKey],
+) -> Result<(), OciClientError> {
+    let mut checked_count = 0;
+
+    for layer in layers {
+        if layer.simple_signing.critical.image.docker_reference != expected_image_digest {
+            debug!(
+                "Signature skipped: digest mismatch (claims {}, expected {})",
+                layer.simple_signing.critical.image.docker_reference, expected_image_digest
+            );
```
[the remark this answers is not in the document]
vjripoll: changed

agent-control/src/oci/signature_verification.rs
vjripoll: yes you are rigtht previously a check was executed here but since is not here not have sense to have the result

PR #2211, opened 2026-02-17
chore(embedded): packages should be no longer embedded

paologallinaharbur:
This change is deleting everything related to embedding binaries.
Also example folders are garbage collected, 

[x] Check external docs for references to the example folder
[x] Check for missed references 
[x] wait for agents to point to OCI real repositories


DavSanchez (approved):
Finally 🥲

PR #2218, opened 2026-02-18
refactor: move async runtime and signatures fetcher to oci client [NR-520796]

sigilioso:
## Summary

- **Move the tokio runtime from `OCIArtifactDownloader` into `oci::Client`**, making `Client` a synchronous-facing wrapper that internally `block_on`s all async OCI operations. This removes the need for callers to manage and pass the runtime separately.
- **Integrate `PublicKeyFetcher` into `oci::Client`** so that signature verification (`verify_signature`) fetches public keys from a JWKS URL directly, replacing the previous API that required callers to supply pre-fetched `PublicKey` slices.
- **Replace `pull_blob` with `pull_blob_to_file`**, encapsulating file creation and `sync_data` inside the client instead of requiring callers to manage `AsyncWrite` streams.

## Notes for reviewers and next steps

- Moving the runtime to the oci-client seems counter intuitive: we need to block the current thread and wait for more tokio tasks mor often. However it is consistent with the rest of the Agent Control code and allow us to reuse already _blocking_ components such as the public-keys fetcher.
- We need to use the client's `verify_signature` method in the downloader (this will come in a follow up PR also including the changes in #2212.

### DIFF
--- a/agent-control/src/oci.rs
+++ b/agent-control/src/oci.rs
@@ -136,13 +136,9 @@ pub mod tests {
         let jwks_key_pairs: Vec<TestKeyPair> = (0..num_jwks_keys).map(TestKeyPair::new).collect();
         // The signer key corresponds to the key position in `signer_position` if any, otherwise
         // it is public key that is not included in the list
-        let separate_signer;
         let kp_signer = match signer_position {
             Some(pos) => &jwks_key_pairs[pos],
-            None => {
-                separate_signer = TestKeyPair::new(num_jwks_keys + 1);
-                &separate_signer
-            }
+            None => &TestKeyPair::new(num_jwks_keys + 1),
         };
 
         let app_server = FakeOciServer::new("my-app", "v1")
--- a/agent-control/src/oci/signature_verification.rs
+++ b/agent-control/src/oci/signature_verification.rs
@@ -86,24 +86,35 @@ impl Client {
         reference: &Reference,
         public_keys: &[PublicKey],
     ) -> Result<Reference, OciClientError> {
-        // Resolve image digest (Client logic)
-        let (_, digest) = self
-            .client
-            .pull_image_manifest(reference, &self.auth)
-            .await
-            .map_err(|err| OciClientError::Verify(format!("could not fetch manifest: {err}")))?;
-        debug!("Image resolved to digest: {}", digest);
+        // Resolve image digest
+        let digest = match reference.digest() {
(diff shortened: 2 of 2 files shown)


agent-control/src/oci/signature_verification.rs
```diff
@@ -56,55 +64,117 @@ pub struct ImageIdentity {
...
-    let (manifest, _) = client.pull_image_manifest(cosign_image_ref).await?;
-    let mut signature_layers = Vec::new();
-
-    for layer in manifest.layers {
-        if layer.media_type != "application/vnd.dev.cosign.simplesigning.v1+json" {
-            continue;
-        }
+impl Client {
+    /// Helper to build the [PublicKeyFetcher] corresponding to the client.
+    pub(super) fn try_build_public_key_fetcher(
```
sigilioso: We are also building a fetcher for the config verifier that checks signatures: https://github.com/newrelic/newrelic-agent-control/blob/63710a647b801b8be56abc39090c18b5cd81c449/agent-control/src/opamp/remote_config/validators/signature/validator.rs#L81

We could consider building it outside and inject it in both (it could even be a shared reference). But I'd leave that refactor out of the scope of this PR

vjripoll (commented):
i think appart of this doubt everything is ok for me :)

agent-control/src/oci.rs
```diff
@@ -116,17 +122,28 @@ pub mod tests {
...
+        // The signer key corresponds to the key position in `signer_position` if any, otherwise
+        // it is public key that is not included in the list
+        let separate_signer;
+        let kp_signer = match signer_position {
+            Some(pos) => &jwks_key_pairs[pos],
+            None => {
+                separate_signer = TestKeyPair::new(num_jwks_keys + 1);
+                &separate_signer
+            }
+        };
```
DavSanchez: Binding to this expression seems to work:

```suggestion
        let kp_signer = match signer_position {
            Some(pos) => &jwks_key_pairs[pos],
            None => &TestKeyPair::new(num_jwks_keys + 1),
        };
```

sigilioso: Simplified here: [2d2db5e](https://github.com/newrelic/newrelic-agent-control/pull/2218/commits/2d2db5e52b9ee6b98e1f5595d84b31c4195dad0d). Thank you!

DavSanchez (approved):

PR #2405, opened 2026-04-10
Nr 548520 merge self replace unix windows

alvarocabanas:
Merge self replace unix/windows since logic is mostly shared.


self-replacer/src/replacer.rs
```diff
@@ -0,0 +1,551 @@
...
+
+        // Create backup manually (as replace_binary does)
+        #[cfg(unix)]
+        fs::copy(&test_file, &backup).unwrap();
+        #[cfg(windows)]
+        fs::rename(&test_file, &backup).unwrap();
+
+        assert!(backup.exists());
+        let backup_content = fs::read(&backup).unwrap();
+        assert_eq!(backup_content, b"test content");
```
sigilioso: Does it make sense to re-implement the function here? This test checks the `backup_path` function only, doesn't it?

self-replacer/src/replacer.rs
```diff
@@ -0,0 +1,551 @@
...
+        fs::write(&new_bin, b"new binary").unwrap();
+
+        let err = replace_binary(&current_exe, &new_bin).unwrap_err();
+
+        // Unix: fails at metadata read before backup
+        // Windows: fails at backup creation (rename)
+        #[cfg(unix)]
+        assert!(matches!(err, ReplaceError::Metadata(_)));
+        #[cfg(windows)]
+        assert!(matches!(err, ReplaceError::BackupCreationFailed(_)));
```
sigilioso: I'd say these specific errors are actually implementation details. We could simply check that it is returning an error:

```suggestion
        replace_binary(&current_exe, &new_bin).err().expect("Error expected when the current binary is missing");
```

If we consider this failure important (in order to detect potential failures), maybe we need to check for existence before trying to get metadata / create the backup and have an explicit name.

self-replacer/src/replacer.rs
```diff
@@ -0,0 +1,551 @@
...
+    #[test]
+    fn test_replace_fails_if_new_bin_missing() {
+        let temp_dir = TempDir::new().unwrap();
+        let new_bin = temp_dir.path().join("program-new");
+
+        // Don't create new_bin - it should not exist
+        let err = BinarySelfReplacer::self_replace(&new_bin).unwrap_err();
+
+        // Should fail with NewBinaryNotFound at the early validation check
+        assert!(matches!(err, ReplaceError::NewBinaryNotFound(_)));
```
sigilioso: Nit: `assert_matches!` is already a dev-dependency in other crates in the namespace we can leverage it if we add it here as well.

```suggestion
        let result = BinarySelfReplacer::self_replace(&new_bin);
        // Should fail with NewBinaryNotFound at the early validation check
        assert_matches!(result, Err(ReplaceError::NewBinaryNotFound(_))));
```

self-replacer/tests/integration_tests.rs
sigilioso: Aren't both implementations supposed to use `BinarySelfReplacer` now?

self-replacer/src/replacer.rs
```diff
@@ -0,0 +1,551 @@
...
+    #[error("failed to read file metadata")]
+    Metadata(#[source] io::Error),
+
+    #[cfg(unix)]
+    #[error("failed to copy new binary to temporary location")]
+    TempCopyFailed(#[source] io::Error),
+
+    #[cfg(unix)]
+    #[error("current executable has no parent directory")]
+    NoParentDirectory,
```
sigilioso, line 68: I wonder if it would make sense to group these three errors and `BackupCretionFailed` in something more generic such as `SetupFailed`. We could have the specific details anyway:

```rust
pub enum ReplaceError {
    // ...
    #error("failed to setup binary replacement {0}")
    SetupFailed(String)
     // ...
}


(...).map_err(ReplaceError::SetupFailed(format!("could not read file metadata for {file_path}: {err}"))?
```

sigilioso (commented):
Find some comments below, mostly _nits_

self-replacer/src/replacer.rs
alvarocabanas: Modified in last commit.

self-replacer/src/replacer.rs
alvarocabanas: Done in last commit.

self-replacer/src/replacer.rs
alvarocabanas: Done in last commit.

self-replacer/tests/integration_tests.rs
alvarocabanas: Done in last commit.

self-replacer/src/replacer.rs
alvarocabanas: I think it's better like this with a more granular handling.

sigilioso: The _group_ would make the error definition platform independent... In any case, we can evolve it as needed when the crate is actually used

alvarocabanas: perfect

self-replacer/src/replacer.rs
```diff
@@ -0,0 +1,545 @@
...
+            Err(if err.kind() == io::ErrorKind::NotFound {
+                ReplaceError::NewBinaryNotFound(source.display().to_string())
+            } else {
+                ReplaceError::ReplaceFailed(err)
+            })
+        }
+    }
+}
+
+/// Creates a backup path by appending .bak to the full filename.
```
sigilioso: really-nit:

```suggestion
/// Creates a backup path by appending [BACKUP_SUFFIX] to the full filename.
```

self-replacer/src/replacer.rs
```diff
@@ -0,0 +1,545 @@
...
+        // Should fail with NewBinaryNotFound at the early validation check
+        assert_matches!(result, Err(ReplaceError::NewBinaryNotFound(_)));
+    }
+
+    // ============================================================================
+    // Unix-specific tests
+    // ============================================================================
+
+    #[cfg(unix)]
+    mod unix_tests {
```
sigilioso: The comment is redundant here, isn't it?

self-replacer/src/replacer.rs
```diff
@@ -0,0 +1,545 @@
...
+            assert_eq!(fs::read(&temp_path).unwrap(), b"new binary");
+        }
+    }
+
+    // ============================================================================
+    // Windows-specific tests
+    // ============================================================================
+
+    #[cfg(windows)]
+    mod windows_tests {
```
sigilioso: Same here! However, I like the nested modules approach to separate platform-specific test 💯

sigilioso (commented):
I've left some _nits_, feel free to ignore them!

self-replacer/src/replacer.rs
alvarocabanas: Done!

self-replacer/src/replacer.rs
alvarocabanas: Done!

self-replacer/src/replacer.rs
alvarocabanas: Done!

sigilioso (approved):

sigilioso (approved):

PR #2406, opened 2026-04-10
feat(updater): on host self updater

paologallinaharbur:
This PR adds a new onHostACUpdater, but it does not change the AC behaviour.


agent-control/src/agent_control/version_updater/on_host.rs
```diff
@@ -1,16 +1,148 @@
...
-use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use tracing::{debug, error};
+use url::Url;
 use wrapper_with_default::WrapperWithDefault;
 
-use crate::agent_control::config::AgentControlDynamicConfig;
-use crate::agent_control::version_updater::updater::{UpdaterError, VersionUpdater};
-use crate::command::SubCommand;
+pub const AGENT_CONTROL_BIN: &str = "newrelic-agent-control";
```
gsanchezgavier, line 24: would be nice if we could extract this at compiled time from some cargo variable 🤔

agent-control/src/agent_control/version_updater/on_host.rs
```diff
@@ -1,16 +1,148 @@
...
+        if new_version == AGENT_CONTROL_VERSION {
+            debug!(
+                "Desired agent control version {new_version} is the same as the current version, skipping update process"
+            );
+            return Ok(());
+        }
+
+        debug!("Starting update process for agent control version {new_version}");
+
+        let Some(package) = &self.reference else {
```
gsanchezgavier: we might want to move this to a constructor

agent-control/src/agent_control/version_updater/on_host.rs
```diff
@@ -1,16 +1,148 @@
...
+        let string_reference = format!(
+            "{}/{}{}",
+            package.download.oci.registry, package.download.oci.repository, new_version
+        );
+
+        let reference = Reference::from(
+            ReferenceParser::from_str(string_reference.as_str()).map_err(|err| {
+                UpdaterError::UpdateFailed(format!("cannot parse reference: {err}"))
+            })?,
+        );
```
gsanchezgavier: i think we could also partially build this on creation to fail fast/hard...

agent-control/src/agent_control/version_updater/on_host.rs
```diff
@@ -197,34 +308,71 @@ mod tests {
...
+
+    #[rstest]
+    #[case("registry.io", "repo/binary:", "v1.0.0", None)]
+    #[case(
+        "ghcr.io",
+        "org/pkg:",
+        "v1.2.3",
+        Some("https://keys.example.com/jwks.json")
+    )]
+    fn test_get_package_data_parses_reference(
```
gsanchezgavier: this test/method seems a bit off here would make sense to move it to crate::agent_control::config::Download ?

agent-control/src/agent_control/config.rs
```diff
@@ -342,6 +350,45 @@ pub struct K8sConfig {
...
+/// OnHostConfig represents the AgentControl configuration for onHost environments
+#[derive(Debug, Deserialize, PartialEq, Clone, Default)]
+pub struct OnHostConfig {
+    /// ac_remote_update enables or disables remote update for agent-control binary
+    #[serde(default)]
+    pub ac_remote_update: bool,
+
+    // TODO provide a default that allows the user not to specify anything
+    #[serde(default)]
+    pub packages: HashMap<PackageID, Package>,
```
gsanchezgavier: I would proabaly not use this hashmap but try to make it open to extension by grouping like
```yaml
on_host:
  self_update:
     enabled: bool
     oci: {}
  {other compo}
```

agent-control/src/agent_control/config.rs
```diff
@@ -342,6 +350,45 @@ pub struct K8sConfig {
...
+
+#[derive(Debug, Deserialize, Default, Clone, PartialEq)]
+pub struct Oci {
+    /// OCI registry url.
+    pub registry: String,
+    /// Repository name.
+    pub repository: String,
+    /// Package version including tag, digest or tag + digest.
+    #[serde(default)]
+    pub version: String,
```
gsanchezgavier: this shouldn't be part of the config i think

agent-control/src/agent_control.rs
```diff
@@ -345,7 +345,11 @@ where
                                         update_opamp_attributes(c, attributes)
                                     .inspect_err(|e| error!(error = %e, select_arm = "agent_control_internal_consumer", "processing version message")));
                                 },
-                            }
+                                AgentControlInternalEvent::StopRequested() => {
```
gsanchezgavier, line 348: i would move this to the integration PR

agent-control/src/agent_control/version_updater/on_host.rs
paologallinaharbur: yes it was a "todo"

agent-control/src/agent_control/config.rs
paologallinaharbur: upsy, when splitting the PR I put it in the wrong place, it was here

```
    /// version represent the AC version that needs to be executed.
    #[serde(skip_serializing_if = "Option::is_none")]
    pub version: Option<String>,
    /// chart_version represent the AC chart version that needs to be executed.
    #[serde(skip_serializing_if = "Option::is_none")]
    pub chart_version: Option<String>,
    /// cd_chart_version represent the agent control cd chart version that needs to be executed.
    #[serde(skip_serializing_if = "Option::is_none")]
    pub cd_chart_version: Option<String>,
```

agent-control/src/agent_control.rs
paologallinaharbur: the problem is that the component send a message that need to exist, if it exists you need to handle it 😕

agent-control/src/agent_control/version_updater/on_host.rs
sigilioso: [CARGO_BIN_NAME](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates) maybe?

gsanchezgavier: just checked this but it only can be used on binary target files , we should define it at main_onhost.rs and passed down through wich is not pretty either. I will keep it as a const

sigilioso: We are already using some _cargo values_ for the [binary metadata](https://github.com/newrelic/newrelic-agent-control/blob/c91c84feba57a1aa76ac14db307143b4431ed7e9/agent-control/src/utils/binary_metadata.rs#L4), I guess we cannot guarantee the value if the library was used outside binary targets, but the agent-control lib is isn't supposed to be used, right?

agent-control/src/agent_control/version_updater/on_host.rs
```diff
@@ -1,16 +1,148 @@
...
+            "Attempting to self-replace with new binary {}",
+            new_binary_path.to_string_lossy()
+        );
+
+        //TODO we should consider managing the errors that can happen in the self-replace process
+        S::self_replace(&new_binary_path).map_err(|e| UpdaterError::UpdateFailed(e.to_string()))?;
+
+        debug!(
+            "Successfully updated agent control to version, stopping the agent control to allow the new version to start",
+        );
```
sigilioso:
```suggestion
        debug!(
            previous_version = AGENT_CONTROL_VERSION,
            %new_version,
            "Agent Control new version is ready, stopping the current binary to allow the new version to start",
        );
```

agent-control/src/agent_control/version_updater/on_host.rs
```diff
@@ -197,34 +308,71 @@ mod tests {
...
+        }
+    }
 
-        let updater = OnHostUpdater::new(executor);
-        assert!(
-            updater
-                .update(&AgentControlDynamicConfig::default())
-                .is_ok()
-        );
+    type TestUpdater = OnHostACUpdater<MockSelfReplacer, MockPackageManager, ProcessVerifyExecutor>;
```
sigilioso: We could also consider moving the function `get_package_data` outside the `OnHostACUpdater` struct and avoid the whole Mock boilerplate for now.

agent-control/src/agent_control/config.rs
sigilioso: If we did it that way we could even consider to omit `on_host`. If `self_update` had good defaults:

```yaml
self_update:
    enabled: bool
    oci: {}
    # ...
```

I know it a different approach from k8s, the on-host configuration is more "exposed" to users (there is no helm-chart in between).

agent-control/src/agent_control/version_updater/on_host.rs
gsanchezgavier: yes i thought about the same, also in the defaults , but as far as i could understand (not well documented) [CARGO_BIN_NAME](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates) is only available on the target file. If you try to use it outside even the compiler errors.

sigilioso: Oh, I see! I thought it behaved the same as `CARGO_PKG_RUST_VERSION`. Understood!

agent-control/src/agent_control/config.rs
gsanchezgavier: after my comment i checked the k8s configs and ended up with something more consistent with it, mainly to avoid a breaking change in k8s and keep consistency

agent-control/src/agent_control/version_updater/on_host.rs
gsanchezgavier: i ended up with different testing scenarios after the refactor

agent-control/src/agent_control/config.rs
gsanchezgavier: after discussing it we decide to make a flatten config in favor for a better UX

agent-control/src/agent_control/version_updater/on_host.rs
```diff
@@ -1,16 +1,160 @@
...
+pub const AGENT_CONTROL_BIN: &str = "newrelic-agent-control.exe";
+
+pub const AGENT_CONTROL_BIN_PACKAGE_ID: &str = "agent_control_bin";
+
+#[derive(Debug, Error)]
+pub enum BuildError {
+    #[error("invalid OCI reference in package config: {0}")]
+    InvalidReference(#[from] oci_client::ParseError),
+}
+pub struct OnHostACUpdater<S, P, V>
```
sigilioso: really-nit: missing newline

agent-control/src/agent_control/version_updater/on_host.rs
```diff
@@ -1,16 +1,160 @@
...
+
+        S::self_replace(&new_binary_path).map_err(|e| {
+            UpdaterError::UpdateFailed(format!("self replacing Agent Control binary: {e}"))
+        })?;
+
+        debug!(
+            previous_version = AGENT_CONTROL_VERSION,
+            %new_version,
+            "Agent Control current binary has replaced by new version, stopping to allow the new version to start",
+        );
```
sigilioso: nit: we might want to define `previous_version` and `new_version` fields consistently in all logs, we could even consider an `span` defined right after getting the new version.

agent-control/src/agent_control/version_updater/on_host.rs
sigilioso: Probably out of scope of this PR (and maybe makes sense for the next one): would a type alias for the on-host implementation used in "main" help?

agent-control/src/agent_control/version_updater/on_host.rs
```diff
@@ -1,16 +1,160 @@
...
+where
+    S: SelfReplacer,
+    P: PackageManager,
+    V: VerifyExecutor,
+{
+    fn update(&self, config: &AgentControlDynamicConfig) -> Result<(), UpdaterError> {
+        if !self.ac_remote_update_enabled {
+            debug!("Remote update is disabled, skipping update process");
+            return Ok(());
+        }
```
sigilioso, line 57: I think we already had this on K8s but we could also consider leaving this outside the struct and using the No-Op implementation when remote update is disabled.

agent-control/src/agent_control/version_updater/on_host.rs
```diff
@@ -1,16 +1,160 @@
...
+    pub fn try_new(
+        ac_remote_update_enabled: bool,
+        agent_control_internal_publisher: EventPublisher<AgentControlInternalEvent>,
+        package_manager: Arc<P>,
+        verify_executor: V,
+        package: AgentControlPackage,
+    ) -> Result<Self, BuildError> {
+        let base_reference = Reference::try_from(format!(
+            "{}/{}",
+            package.download.oci.registry, package.download.oci.repository
```
sigilioso: I'm not sure if it applies because there is no version in the base reference, but I'm pointing it out just in case. Should we use the reference parser introduced in #2301 instead of `Reference::try_from`?

agent-control/src/agent_control/version_updater/on_host.rs
sigilioso, line 1: We have more stuff in the module now, would it make sense to move the `VerifyExecutor` to its own module?  Something similar to `version_updater::on_host::verify`

agent-control/src/agent_control/version_updater/on_host.rs
```diff
@@ -197,34 +317,68 @@ mod tests {
...
+                true,
+                publisher,
+                Arc::new(MockPackageManager::new()),
+                MockVerifyExecutorMock::new(),
+                package,
+            )
+            .err()
+            .unwrap(),
+            BuildError::InvalidReference(_)
+        );
```
sigilioso, line 384: Nit: 

```suggestion
        assert_matches!(
            TestUpdater::try_new(
                true,
                publisher,
                Arc::new(MockPackageManager::new()),
                MockVerifyExecutorMock::new(),
                package,
            ),
            Err(BuildError::InvalidReference(_))
        );
```

agent-control/src/agent_control/config.rs
```diff
@@ -342,6 +363,40 @@ pub struct K8sConfig {
...
+impl Default for AgentControlPackage {
+    fn default() -> Self {
+        AgentControlPackage {
+            download: Download {
+                oci: Oci {
+                    registry: AC_OCI_PACKAGE_DEFAULT_REGISTRY.to_string(),
+                    repository: AC_OCI_PACKAGE_DEFAULT_REPOSITORY.to_string(),
+                    public_key_url: Url::parse(AC_OCI_PACKAGE_PUBLIC_KEY_URL)
+                        .expect("valid default url"),
+                },
```
sigilioso, line 379: Could we add a simple test that calls `AgentControlPackage::default()`? Just to assure that unit tests caught a potential typo on that constant.

agent-control/src/agent_control/config.rs
```diff
@@ -68,6 +72,20 @@ pub struct AgentControlConfig {
...
+    #[serde(default)]
+    pub self_update: SelfUpdateConfig,
+}
+
+#[derive(Debug, Default, Deserialize, PartialEq, Clone)]
+pub struct SelfUpdateConfig {
+    /// Indicates whether the self remote update mechanism is enabled or not
+    pub enabled: bool,
+    /// Indicates whether package signature verification is enabled or not
+    pub signature_verification_enabled: SignatureVerificationEnabled,
```
sigilioso, line 86: Would it be possible to enable/disable signature verification for sub-agents and Agent Control separately? I'm asking just in case we need to unify this field somehow.

sigilioso (commented):
It is looking good! Find some comments below

agent-control/src/agent_control/version_updater/on_host.rs
gsanchezgavier: good catch! i forgot about it ! I'll use it

agent-control/src/agent_control/version_updater/on_host.rs
gsanchezgavier: not sure if i follow , you mean for this? https://github.com/newrelic/newrelic-agent-control/pull/2407/changes#diff-790a5bd62678076c5484aa395ca3571f27c62c7227aec0dd521bec9d2364d0c7R234

agent-control/src/agent_control/version_updater/on_host.rs
gsanchezgavier: yes i didn't change it because of that, we could do a follow up PR after all is done with that change , i think it would be more easy to be done and review

agent-control/src/agent_control/version_updater/on_host.rs
gsanchezgavier: makes sense i'll create a refact pr and put this on top of that

agent-control/src/agent_control/config.rs
gsanchezgavier: as long as they support having different OCI registry definitions i think we should support a separate disable flag

agent-control/src/agent_control/version_updater/on_host.rs
gsanchezgavier: I'll do it after along with the other refactor actually , I think it will help to unblock the dependant tasks

agent-control/src/agent_control/version_updater/on_host.rs
gsanchezgavier: oh i tried but hit a wall in debug implementaions

agent-control/src/agent_control/version_updater/on_host.rs
gsanchezgavier: done i forgot we had it !

agent-control/src/agent_control/version_updater/on_host.rs
sigilioso: Not important in any case! 🙂

agent-control/src/agent_control/version_updater/on_host.rs
sigilioso: I meant something similar to:

```rust
type OnHostUpdater = OnHostACUpdater<RealImpl, RealImpl, RealImp>;
```

But after checking out the other PR, it is actually **not** needed. It is only useful if we need to implement additional (and specific) methods or use it as an argument, and it is not the case!.

agent-control/src/agent_control/config.rs
sigilioso: I was thinking of this https://github.com/newrelic/newrelic-agent-control/blob/4678614957fed5d0b4e418f5f6df97680cdbfcd7/agent-control/src/agent_control/config.rs#L73-L77

But we will likely need to consider both fields when doing the _wiring_, let's discuss that in the next PR

sigilioso (approved):
🚀

agent-control/src/agent_control/version_updater/on_host.rs
```diff
@@ -1,16 +1,157 @@
...
+        self.verify_executor
+            .execute(&new_binary_path, &["verify"])
+            .map_err(|e| {
+                UpdaterError::UpdateFailed(format!("verifying new Agent Control binary: {e}"))
+            })?;
+
+        debug!("Attempting to self-replace with new binary",);
+
+        BinarySelfReplacer::self_replace(&new_binary_path).map_err(|e| {
```
paologallinaharbur, line 101: how can you mock this if you are calling the "real" type?

paologallinaharbur (commented):
I cannot approve it, but apart from that doubt (that I guess will be addressed in the next PRs) it is good to go for me

agent-control/src/agent_control/version_updater/on_host.rs
gsanchezgavier: we are not mocking this in unit testing . As the previous approach currently there are no unit tests that exercise this ( we have integration tests in the other Pr ) so just removing the generic for simplicity.

PR #2407, opened 2026-04-10
feat(updater): change AC to leverage the updater

paologallinaharbur:
- Replace the NoOpUdater onhost updater to the implemented one
- Adds an integration test

### DIFF
--- a/agent-control/src/agent_control.rs
+++ b/agent-control/src/agent_control.rs
@@ -354,22 +354,12 @@ where
                     }
                 }
                 recv(self.application_event_consumer.as_ref()) -> agent_control_event => {
-                    match agent_control_event{
-                        Err(err) => {
-                            debug!("Error receiving Agent Control internal event {err}")
-                        },
-                        Ok(event) => {
-                            match event {
-                                ApplicationEvent::StopRequested => {
-                                    let span = info_span!("process_application_event", id=AGENT_CONTROL_ID);
-                                    let _span_guard = span.enter();
-                                    debug!("stopping Agent Control event processor");
-                                    self.agent_control_publisher.broadcast(AgentControlEvent::AgentControlStopped);
-                                    break sub_agents.stop();
-                                }
-                            }
-                        },
-                    }
+                    let span = info_span!("process_application_event", id=AGENT_CONTROL_ID);
+                    let _span_guard = span.enter();
+                    let _= agent_control_event.inspect_err(|err| error!(error = %err, select_arm = "application_event_consumer", "receiving application event"));
+                    debug!("stopping Agent Control event processor");
+                    self.agent_control_publisher.broadcast(AgentControlEvent::AgentControlStopped);
+                    break sub_agents.stop();
                 },
                 recv(uptime_reporter.receiver()) -> _tick => { let _ = uptime_reporter.report(); },
             }
--- a/agent-control/src/agent_control.rs
+++ b/agent-control/src/agent_control.rs
@@ -346,7 +346,7 @@ where
                                     .inspect_err(|e| error!(error = %e, select_arm = "agent_control_internal_consumer", "processing version message")));
                                 },
                                 AgentControlInternalEvent::StopRequested() => {
-                                    debug!("stopping Agent Control event processor after request");
+                                    debug!("Stopping Agent Control event processor after request");
                                     self.agent_control_publisher.broadcast(AgentControlEvent::AgentControlStopped);
                                     break sub_agents.stop();
                                 }}
@@ -356,8 +356,8 @@ where
                 recv(self.application_event_consumer.as_ref()) -> agent_control_event => {
                     let span = info_span!("process_application_event", id=AGENT_CONTROL_ID);
                     let _span_guard = span.enter();
-                    let _= agent_control_event.inspect_err(|err| error!(error = %err, select_arm = "application_event_consumer", "receiving application event"));
-                    debug!("stopping Agent Control event processor");
+                    let _= agent_control_event.inspect_err(|err| error!(error = %err, select_arm = "application_event_consumer", "Receiving application event"));
+                    debug!("Stopping Agent Control event processor");
                     self.agent_control_publisher.broadcast(AgentControlEvent::AgentControlStopped);
                     break sub_agents.stop();
                 },
--- a/agent-control/src/agent_control.rs
+++ b/agent-control/src/agent_control.rs
@@ -276,7 +276,7 @@ where
             <<S as SubAgentBuilder>::NotStartedSubAgent as NotStartedSubAgent>::StartedSubAgent,
         >,
     ) {
-        debug!("Listening for events from agents");
+        debug!("Listening for events");
         let never_receive = EventConsumer::from(never());
         let opamp_receiver = self
             .agent_control_opamp_consumer
@@ -353,12 +353,23 @@ where
                         },
                     }
                 }
-                recv(self.application_event_consumer.as_ref()) -> _agent_control_event => {
-                    let span = info_span!("process_application_event", id=AGENT_CONTROL_ID);
-                    let _span_guard = span.enter();
-                    debug!("stopping Agent Control event processor");
-                    self.agent_control_publisher.broadcast(AgentControlEvent::AgentControlStopped);
-                    break sub_agents.stop();
+                recv(self.application_event_consumer.as_ref()) -> agent_control_event => {
+                    match agent_control_event{
+                        Err(err) => {
+                            debug!("Error receiving Agent Control internal event {err}")
+                        },
+                        Ok(event) => {
+                            match event {
+                                ApplicationEvent::StopRequested => {
+                                    let span = info_span!("process_application_event", id=AGENT_CONTROL_ID);
+                                    let _span_guard = span.enter();
+                                    debug!("stopping Agent Control event processor");
+                                    self.agent_control_publisher.broadcast(AgentControlEvent::AgentControlStopped);
+                                    break sub_agents.stop();
+                                }
+                            }
+                        },
+                    }
                 },
                 recv(uptime_reporter.receiver()) -> _tick => { let _ = uptime_reporter.report(); },
             }
--- a/agent-control/src/agent_control/run/on_host.rs
+++ b/agent-control/src/agent_control/run/on_host.rs
@@ -1,5 +1,7 @@
 use crate::agent_control::AgentControl;
-use crate::agent_control::config::{AgentControlConfig, OpAMPClientConfig};
+use crate::agent_control::config::{
+    AGENT_CONTROL_BIN_PACKAGE_ID, AgentControlConfig, OpAMPClientConfig,
+};
 use crate::agent_control::config_repository::repository::AgentControlConfigLoader;
 use crate::agent_control::config_validator::RegistryDynamicConfigValidator;
 use crate::agent_control::defaults::{
@@ -12,7 +14,8 @@ use crate::agent_control::resource_cleaner::no_op::NoOpResourceCleaner;
 use crate::agent_control::run::{
     AgentControlRunner, Environment, RunError, RunningMode, setup_config_repository_and_store,
 };
-use crate::agent_control::version_updater::updater::NoOpUpdater;
+use crate::agent_control::version_updater::on_host::OnHostACUpdater;
+use crate::agent_control::version_updater::on_host::ProcessVerifyExecutor;
 use crate::agent_type::render::TemplateRenderer;
 use crate::agent_type::variable::Variable;
 use crate::checkers::health::noop::NoOpHealthChecker;
@@ -53,6 +56,10 @@ use oci_client::client::ClientProtocol;
 use opamp_client::http::StartedHttpClient;
 use opamp_client::http::client::OpAMPHttpClient;
 use opamp_client::operation::settings::{DescriptionValueType, StartSettings};
+#[cfg(target_family = "unix")]
+use self_replacer::UnixSelfReplacer;
+#[cfg(target_family = "windows")]
+use self_replacer::windows::WindowsSelfReplacer;
 use std::collections::HashMap;
 use std::path::PathBuf;
 use std::sync::Arc;
@@ -96,6 +103,8 @@ impl AgentControlRunner {
         let agent_control_config = config_storer
             .load()
             .map_err(|err| RunError(format!("failed to load Agent Control config: {err}")))?;
+
+        let on_host_config = agent_control_config.onhost.clone().unwrap_or_default();
 
         let identifiers = ac_identifiers(&agent_control_config)?;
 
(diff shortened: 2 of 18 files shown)


agent-control/tests/on_host/scenarios/ac_self_update.rs
```diff
@@ -0,0 +1,315 @@
...
+        .join("packages")
+        .join("binary")
+        .join("installation")
+        .join(AC_BINARY_NAME);
+
+    assert!(
+        new_binary_path.exists(),
+        "New binary should be downloaded and extracted"
+    );
+
```
paologallinaharbur: todo show logs

agent-control/tests/on_host/scenarios/ac_self_update.rs
```diff
@@ -0,0 +1,315 @@
+use crate::common::agent_control::start_agent_control_with_custom_config;
```
paologallinaharbur, line 1: disclaimer, this is just a draft, I've not seen it running

agent-control/src/agent_control/run/on_host.rs
```diff
@@ -52,6 +55,10 @@ use oci_client::client::ClientProtocol;
 use opamp_client::http::StartedHttpClient;
 use opamp_client::http::client::OpAMPHttpClient;
 use opamp_client::operation::settings::DescriptionValueType;
+#[cfg(target_family = "unix")]
+use self_replacer::UnixSelfReplacer;
+#[cfg(target_family = "windows")]
+use self_replacer::windows::WindowsSelfReplacer;
```
gsanchezgavier: uh this is ugly , but i think it would be unified by @alvarocabanas

paologallinaharbur: yes I believe we should hide this complexity inside the module, exspecially since the struct does not really need any attribute!

agent-control/src/agent_control.rs
```diff
@@ -353,12 +353,23 @@ where
...
-                    let span = info_span!("process_application_event", id=AGENT_CONTROL_ID);
-                    let _span_guard = span.enter();
-                    debug!("stopping Agent Control event processor");
-                    self.agent_control_publisher.broadcast(AgentControlEvent::AgentControlStopped);
-                    break sub_agents.stop();
+                recv(self.application_event_consumer.as_ref()) -> agent_control_event => {
+                    match agent_control_event{
+                        Err(err) => {
+                            debug!("Error receiving Agent Control internal event {err}")
+                        },
```
gsanchezgavier: is this ok? i think we were not checking the error because we want to gracefully shut down anyway . perhaps we could only print an error line

agent-control/tests/on_host/scenarios/ac_self_update.rs
gsanchezgavier: as main changes i had to change the script approach to a compiled binary to support also windows, i also changed the asserts to just validate the remote config status and restart of AC

agent-control/src/agent_control/run/on_host.rs
```diff
@@ -164,20 +165,23 @@ impl AgentControlRunner {
...
+            oci_client.clone(),
             agent_control_config
                 .agent_packages
                 .signature_verification_enabled
                 .into(),
         );
 
-        let package_manager =
-            OCIPackageManager::new(packages_downloader, DirectoryManagerFs, remote_dir.clone());
+        let agents_package_manager = OCIPackageManager::new(
```
vjripoll, line 176: nit: Perhaps we could extract the creation of the OCI package manager into a simple helper function to avoid code duplication:

Something like this?

```
fn create_oci_package_manager(
    oci_client: oci::Client,
    signature_verification_enabled: SignatureVerificationEnabled,
    remote_dir: PathBuf,
) -> Arc<OCIPackageManager<OCIArtifactDownloader, DirectoryManagerFs>> {
    let downloader = OCIArtifactDownloader::new(oci_client, signature_verification_enabled.into());
    Arc::new(OCIPackageManager::new(
        downloader,
        DirectoryManagerFs,
        remote_dir,
    ))
}
```

vjripoll (commented):
🚀

agent-control/src/agent_control/run/on_host.rs
```diff
@@ -212,6 +216,30 @@ impl AgentControlRunner {
             .map_err(|err| RunError(format!("failed to start HTTP server: {err}")))?;
 
         let (agent_control_internal_publisher, agent_control_internal_consumer) = pub_sub();
+
+        let packages_downloader = OCIArtifactDownloader::new(
```
paologallinaharbur: agent_control_packages_downloader?

agent-control/src/agent_control/run/on_host.rs
```diff
@@ -212,6 +216,30 @@ impl AgentControlRunner {
...
+            oci_client.clone(),
+            agent_control_config
+                .self_update
+                .signature_verification_enabled
+                .into(),
+        );
+
+        let package_manager = Arc::new(OCIPackageManager::new(
```
paologallinaharbur: I guess the arc is no longer needed if we are not sharing the packageManager with the subAgents

agent-control/src/agent_control.rs
```diff
@@ -346,17 +346,18 @@ where
...
                         },
                     }
                 }
-                recv(self.application_event_consumer.as_ref()) -> _agent_control_event => {
+                recv(self.application_event_consumer.as_ref()) -> agent_control_event => {
                     let span = info_span!("process_application_event", id=AGENT_CONTROL_ID);
                     let _span_guard = span.enter();
-                    debug!("stopping Agent Control event processor");
+                    let _= agent_control_event.inspect_err(|err| error!(error = %err, select_arm = "application_event_consumer", "Receiving application event"));
+                    debug!("Stopping Agent Control event processor");
```
paologallinaharbur, line 360: way better then my approach 😅

agent-control/tests/common/util.rs
```diff
@@ -0,0 +1,12 @@
+/// Returns the name of the current test by reading the test thread's name.
+///
+/// The Rust test harness names each test's thread after its full module path
+/// (e.g. `on_host::scenarios::ac_self_update::test_agent_control_self_update_with_oci_registry`).
+/// This gives a stable, unique-per-test identifier that is safe to use when
+/// concurrent tests need to distinguish their own artifacts from one another.
+pub(crate) fn current_test_id() -> String {
+    std::thread::current()
+        .name()
+        .unwrap_or("unknown")
```
paologallinaharbur: would make sense to just explode instead? I fear that we could hide issues if for any reason the resolution breaks

agent-control/src/agent_control.rs
```diff
@@ -346,17 +346,18 @@ where
...
-                                    debug!("stopping Agent Control event processor after request");
+                                    debug!("Stopping Agent Control event processor after request");
                                     self.agent_control_publisher.broadcast(AgentControlEvent::AgentControlStopped);
                                     break sub_agents.stop();
                                 }}
                         },
                     }
                 }
-                recv(self.application_event_consumer.as_ref()) -> _agent_control_event => {
+                recv(self.application_event_consumer.as_ref()) -> agent_control_event => {
```
paologallinaharbur, line 356: The only think I do not like about this approach is that we are not really checking that "StopRequested" is sent. This works only because there is only one variant, right?

paologallinaharbur, line 356: moreover I would change the description of the message:
```
/// Defines application events: these events are sent directly to the application. Eg: OS-signals.
#[derive(Clone, Debug, PartialEq)]
pub enum ApplicationEvent {
    StopRequested,
}
```

to explain that is used as well to self-replace.

A different possibility is changing ApplicationEvent to be just a StopRequestEvent and be a struct and not an enum

agent-control/tests/common/agent_control.rs
```diff
@@ -62,19 +63,37 @@ pub fn start_agent_control_with_custom_config(
 
 pub struct StartedAgentControl {
     application_event_publisher: EventPublisher<ApplicationEvent>,
-    handle: Option<std::thread::JoinHandle<()>>,
+
+    handle: Option<std::thread::JoinHandle<Result<(), RunError>>>,
+}
```
paologallinaharbur, line 68: why do we return the error as part of the handler? wouldn't be cleaner to have Result<StartedAgentControl, RunError> instead?

agent-control/tests/common/agent_control.rs
```diff
@@ -62,19 +63,37 @@ pub fn start_agent_control_with_custom_config(
...
+impl StartedAgentControl {
+    pub fn has_gracefully_stopped(&mut self) -> bool {
+        if self
+            .handle
+            .as_ref()
+            .is_some_and(|handle| handle.is_finished())
+        {
+            let result = self.handle.take().unwrap().join().unwrap();
+            return result.is_ok();
```
paologallinaharbur, line 77: would make sense to print the error if the result is not ok?

agent-control/tests/on_host/tools/oci_package_manager.rs
```diff
@@ -35,50 +33,63 @@ pub fn new_testing_oci_package_manager(
 pub struct TestDataHelper;
 
 impl TestDataHelper {
-    fn create_data_to_compress(tmp_dir_to_compress: &Path, file_dir: &str, file_content: &str) {
-        let file_path = tmp_dir_to_compress.join(file_dir);
-        File::create(file_path.clone()).unwrap();
-        std::fs::write(file_path.as_path(), file_content).unwrap();
-    }
```
paologallinaharbur, line 42: just out of curiosity, why are you removing this?

agent-control/tests/on_host/tools/oci_package_manager.rs
```diff
@@ -35,50 +33,63 @@ pub fn new_testing_oci_package_manager(
...
         let mut tar = tar::Builder::new(enc);
-        let file_path = source_path.join(filename);
         tar.append_path_with_name(&file_path, filename).unwrap();
         tar.finish().unwrap();
     }
 
+    /// Compresses a single file into a tar.gz archive, storing it with mode 0o755.
+    /// The entry name in the archive is taken from the file's own name.
+    #[cfg(target_family = "unix")]
+    pub fn compress_tar_gz_file(file_path: &Path, archive_path: &Path) {
```
paologallinaharbur: compress_tar_gz_executable?

agent-control/tests/on_host/tools/fake_binary.rs
```diff
@@ -0,0 +1,41 @@
...
+pub fn assert_is_fake_binary(path: &Path) {
+    let test_id = current_test_id();
+    let output = std::process::Command::new(path)
+        .arg("id")
+        .output()
+        .expect("failed to execute process");
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    assert!(
+        stdout.contains(&test_id),
+        "Expected current exe to be fake_ac built for test '{}', got id output: {}",
```
paologallinaharbur, line 47: I am a bit confused "got id output: {}" is wrong, I mean, if this fails only when test_id is not in the output, no?

agent-control/tests/on_host/scenarios/ac_self_update.rs
```diff
@@ -0,0 +1,324 @@
...
+use newrelic_agent_control::agent_control::run::on_host::OCI_TEST_REGISTRY_URL;
+use newrelic_agent_control::package::oci::artifact_definitions::PackageMediaType;
+use opamp_client::opamp::proto::RemoteConfigStatuses;
+use std::path::Path;
+use std::time::Duration;
+use tempfile::tempdir;
+
+#[test]
+#[ignore = "needs oci registry (use *with_oci_registry suffix)"]
+/// Be aware that this test executes a binary self-replace which in this case is the test binary itself.
```
paologallinaharbur: can you document a bit what the tests are doing and which is the approach?

agent-control/tests/on_host/scenarios/ac_self_update.rs
paologallinaharbur, line 1: does it make sense to test as well a failed validation? I believe it would be enough to push a empty binary or wrong package

paologallinaharbur, line 1: We can also check that it can "recover" from a failure. 

 - AC runs
 - it receives a wrong config/binary -> it fails
 - AC receives a new config
 - AC is capable of self-replace

agent-control/src/agent_control.rs
gsanchezgavier: yes i think so, anyway i think this is out of the scope of this PR , I just reverted the previous change as it was not compleately right .

gsanchezgavier: but the internal event is used for that AgentControlInternalEvent::StopRequested

agent-control/src/agent_control/run/on_host.rs
gsanchezgavier: i just remove the intermediate variable

agent-control/tests/common/util.rs
gsanchezgavier: yess

agent-control/tests/common/agent_control.rs
gsanchezgavier: no sure if get the question , but the idea is that whenever you join the thread that is running AC you can verify if it has been gracefully shutdown or not

agent-control/tests/common/agent_control.rs
gsanchezgavier: i wanted to favor the test readability . in case the test fails and someone is interested in the why , we have the full debug logs .

agent-control/tests/on_host/tools/oci_package_manager.rs
gsanchezgavier: yes makes sense!

agent-control/tests/on_host/tools/oci_package_manager.rs
gsanchezgavier: oh i think this is a leftover mod of another refactor that finally change, this helper still makes sense

agent-control/tests/on_host/tools/fake_binary.rs
gsanchezgavier: yes this is what is being checked . the binary prints the test id hardcoded at compiled time.

agent-control/tests/on_host/scenarios/ac_self_update.rs
gsanchezgavier: 👍  just added the test

gsanchezgavier: make sense. I will create a task for it, it will require to overcome the concurency issue that having two different test self-replacing the test binary in parallel have.

paologallinaharbur (commented):
I cannot approve it, but consider it approved!

gsanchezgavier (approved):

PR #2414, opened 2026-04-14
feat: support oci auth in agent types

danielorihuela:
Add OCI authentication support (anonymous, basic, bearer) to agent types.

### Context

Some clients can't use the docker public registry. This is the first step supporting authentication (including mirrors). Specifically, this PR modifies Agent Types and it's internal representation in Agent Control to support the different [authentication methods supported by the oci rust client](https://docs.rs/oci-client/latest/oci_client/secrets/enum.RegistryAuth.html).

### Technical decisions

**Implicit anonymous authentication**

The "anonymous" authentication method doesn't have any input parameters. There's nothing to configure on the `deployment` section of the agent type. We could have a boolean flag to enable or disable it, but I don't see what would be the advantage of doing that, other than making AC fail if no auth is provided and "anonymous" is disabled. However, I would expect the registry itself to return an error if we can't access it with an "anonymous" authentication account.

With that in mind, having it implicit seems the best approach. We don't add extra complexity to AC and let the registry handle authentication failures.

```yaml
auth:
  anonymous: {} --------> Doesn't make any sense
  basic:
    username: ${nr-var:oci.auth.basic.username}
    password: ${nr-var:oci.auth.basic.password}
  bearer: ${nr-var:oci.auth.bearer}
```

**Optional new auth field**

Old agent types don't have an `auth` field for the `oci` variable. Therefore, for backwards compatibility, we are forced to make this new field optional. This doesn't mean that the value is `Optional` in the rust codebase, but that we will use the default auth value if it's not there. The default value will end up transformed into `RegistryAuth::Anonymous`.

**Arbitrary authentication precedence**

The current implementation follows these rules:

1. If agent type doesn't have auth configured -> anonymous
2. If agent type configures auth and user configures basic credentials -> basic
3. If agent type configures auth and user configures bearer credentials -> bearer
4. If agent type configures auth and user configures basic + bearer credentials -> basic

We could easily throw an error if both basic + bearer credentials are configured. If we do that, then a working deployment might fail just because the user configured two methods of authentication instead of one. Not punishing it and forcing a new deployment seems a good idea. Specially when a user shouldn't configure on purpose more than one method.

That said, the precedence is arbitrary. We prefer basic credentials because they have lower risk of expiring.

Feedback on that front is welcome.

**AC understands empty username or password as wrong credentials**

We don't differentiate between empty basic credentials and none basic credentials. This makes the code representation and handling easier. No `Optional` and no changing behaviour.
This seems a sane behaviour given that an empty username and/or empty password decreases security. In that case, the user should be using anonymous, and that's what AC will chose if no bearer is configured.

I don't see a use case for empty username and/or password.

There's one disadvatange. If there's a use case for empty username and/or password, we don't support it. Again, I don't see why this would be the case so we discarded it.

**If auth is configured, all credentials are mandatory**

In other words, the `deployment` section is forced to have all the possible methods of authentication there.

GOOD
```yaml
auth:
  basic:
    username: ${nr-var:oci.auth.basic.username}
    password: ${nr-var:oci.auth.basic.password}
  bearer: ${nr-var:oci.auth.bearer}
```

Or what's the same. An agent type author cannot only configure one type of authentication.

BAD
```yaml
auth:
  bearer: ${nr-var:oci.auth.bearer}
```

Given that the burden of authentication is not on the agent type author, but on the registry configuration, we are just asking the author to copy paste a yaml template. Doesn't seem a big deal. On the bright side, this makes sure that clients with mirrors don't have issues because their preferred authentication method is unsupported by the agent (as long as it's supported on AC). By forcing agent type authors, we also "protect" them from having to publish a new version if an auth method is missing. A minor thing honestly. The big advantage here is on the user side.

From AC side, we avoid adding `Optional` for each method and then having to deal with that on the codebase and the parsing. Probably not a big change, but it's less code at the end of the day.

The disadvantage is that if we ever add any other authentication method, we will need to make it `Optional` and be inconsistent or make all of the methods `Optional` and change the behaviour of AC (that's not a breaking change, but could be confusing). I almost forgot, we could make it mandatory and add a breaking change to all agent types. Not good.

All in all, authentication methods shouldn't change. I believe the advatanges outweight the potential disadvantages.

### DIFF
--- a/agent-control/agent-type-registry/newrelic/com.newrelic.infrastructure-0.1.0.yaml
+++ b/agent-control/agent-type-registry/newrelic/com.newrelic.infrastructure-0.1.0.yaml
@@ -58,10 +58,11 @@ variables:
             required: false
             default: ""
         bearer:
-          description: "Bearer token for authentication"
-          type: string
-          required: false
-          default: ""
+          token:
+            description: "Bearer token for authentication"
+            type: string
+            required: false
+            default: ""
     version:
       description: "Agent version"
       type: string
@@ -127,10 +128,11 @@ variables:
             required: false
             default: ""
         bearer:
-          description: "Bearer token for authentication"
-          type: string
-          required: false
-          default: ""
+          token:
+            description: "Bearer token for authentication"
+            type: string
+            required: false
+            default: ""
     version:
       description: "Agent version"
       type: string
@@ -224,7 +226,8 @@ deployment:
               basic:
                 username: ${nr-var:oci.auth.basic.username}
                 password: ${nr-var:oci.auth.basic.password}
-              bearer: ${nr-var:oci.auth.bearer}
+              bearer:
+                token: ${nr-var:oci.auth.bearer.token}
     version:
       path: ${nr-sub:packages.infra-agent.dir}\\newrelic-infra.exe
       args:
@@ -299,7 +302,8 @@ deployment:
               basic:
                 username: ${nr-var:oci.auth.basic.username}
                 password: ${nr-var:oci.auth.basic.password}
-              bearer: ${nr-var:oci.auth.bearer}
+              bearer:
+                token: ${nr-var:oci.auth.bearer.token}
     version:
       path: ${nr-sub:packages.infra-agent.dir}/newrelic-infra
       args:
--- a/agent-control/agent-type-registry/newrelic/com.newrelic.infrastructure-0.1.0.yaml
+++ b/agent-control/agent-type-registry/newrelic/com.newrelic.infrastructure-0.1.0.yaml
@@ -45,6 +45,23 @@ variables:
         variants:
           ac_config_field: "oci_repository_urls"
           values: [ "newrelic/infrastructure-agent-artifacts" ]
+      auth:
+        basic:
+          username:
+            description: "Username for HTTP Basic authentication"
+            type: string
+            required: false
+            default: ""
+          password:
+            description: "Password for HTTP Basic authentication"
+            type: string
+            required: false
+            default: ""
+        bearer:
+          description: "Bearer token for authentication"
+          type: string
+          required: false
+          default: ""
     version:
       description: "Agent version"
       type: string
@@ -97,6 +114,23 @@ variables:
         variants:
           ac_config_field: "oci_repository_urls"
           values: [ "newrelic/infrastructure-agent-artifacts" ]
+      auth:
+        basic:
+          username:
+            description: "Username for HTTP Basic authentication"
+            type: string
+            required: false
+            default: ""
+          password:
+            description: "Password for HTTP Basic authentication"
+            type: string
+            required: false
+            default: ""
+        bearer:
+          description: "Bearer token for authentication"
+          type: string
+          required: false
+          default: ""
     version:
       description: "Agent version"
       type: string
@@ -186,6 +220,11 @@ deployment:
             repository: ${nr-var:oci.repository}
             version: ${nr-var:version}
             public_key_url: https://publickeys.newrelic.com/g/agent-control-oci/global/nrinfraagent/jwks.json
+            auth:
+              basic:
+                username: ${nr-var:oci.auth.basic.username}
+                password: ${nr-var:oci.auth.basic.password}
+              bearer: ${nr-var:oci.auth.bearer}
     version:
       path: ${nr-sub:packages.infra-agent.dir}\\newrelic-infra.exe
       args:
@@ -256,6 +295,11 @@ deployment:
             repository: ${nr-var:oci.repository}
             version: ${nr-var:version}
             public_key_url: https://publickeys.newrelic.com/g/agent-control-oci/global/nrinfraagent/jwks.json
+            auth:
+              basic:
+                username: ${nr-var:oci.auth.basic.username}
+                password: ${nr-var:oci.auth.basic.password}
+              bearer: ${nr-var:oci.auth.bearer}
     version:
       path: ${nr-sub:packages.infra-agent.dir}/newrelic-infra
       args:
--- a/agent-control/agent-type-registry/newrelic/com.newrelic.opentelemetry.collector-0.1.0.yaml
+++ b/agent-control/agent-type-registry/newrelic/com.newrelic.opentelemetry.collector-0.1.0.yaml
@@ -49,6 +49,23 @@ variables:
         variants:
           ac_config_field: "oci_nrdot_registry_repositories"
           values: [ "newrelic/nrdot-agent-artifacts" ]
+      auth:
+        basic:
+          username:
+            description: "Username for HTTP Basic authentication"
+            type: string
+            required: false
+            default: ""
+          password:
+            description: "Password for HTTP Basic authentication"
+            type: string
+            required: false
+            default: ""
+        bearer:
+          description: "Bearer token for authentication"
+          type: string
+          required: false
+          default: ""
     version:
       description: "Agent version"
       type: string
@@ -95,6 +112,23 @@ variables:
         variants:
           ac_config_field: "oci_nrdot_registry_repositories"
           values: [ "newrelic/nrdot-agent-artifacts" ]
+      auth:
+        basic:
+          username:
+            description: "Username for HTTP Basic authentication"
+            type: string
+            required: false
+            default: ""
+          password:
+            description: "Password for HTTP Basic authentication"
+            type: string
+            required: false
+            default: ""
+        bearer:
+          description: "Bearer token for authentication"
+          type: string
+          required: false
+          default: ""
     version:
       description: "Agent version"
       type: string
@@ -157,6 +191,11 @@ deployment:
             repository: ${nr-var:oci.repository}
             version: ${nr-var:version}
             public_key_url: https://publickeys.newrelic.com/g/agent-control-oci/global/nrdot/jwks.json
+            auth:
+              basic:
+                username: ${nr-var:oci.auth.basic.username}
+                password: ${nr-var:oci.auth.basic.password}
+              bearer: ${nr-var:oci.auth.bearer}
     version:
       path: ${nr-sub:packages.nrdot.dir}\\nrdot-collector.exe
       args:
@@ -190,6 +229,11 @@ deployment:
             repository: ${nr-var:oci.repository}
             version: ${nr-var:version}
             public_key_url: https://publickeys.newrelic.com/g/agent-control-oci/global/nrdot/jwks.json
+            auth:
+              basic:
+                username: ${nr-var:oci.auth.basic.username}
+                password: ${nr-var:oci.auth.basic.password}
+              bearer: ${nr-var:oci.auth.bearer}
     health:
       interval: 30s
       initial_delay: 90s
--- a/agent-control/agent-type-registry/newrelic/com.newrelic.opentelemetry.collector-0.1.0.yaml
+++ b/agent-control/agent-type-registry/newrelic/com.newrelic.opentelemetry.collector-0.1.0.yaml
@@ -62,10 +62,11 @@ variables:
             required: false
             default: ""
         bearer:
-          description: "Bearer token for authentication"
-          type: string
-          required: false
-          default: ""
+          token:
+            description: "Bearer token for authentication"
+            type: string
+            required: false
+            default: ""
     version:
       description: "Agent version"
       type: string
@@ -125,10 +126,11 @@ variables:
             required: false
             default: ""
         bearer:
-          description: "Bearer token for authentication"
-          type: string
-          required: false
-          default: ""
+          token:
+            description: "Bearer token for authentication"
+            type: string
+            required: false
+            default: ""
     version:
       description: "Agent version"
       type: string
@@ -195,7 +197,8 @@ deployment:
               basic:
                 username: ${nr-var:oci.auth.basic.username}
                 password: ${nr-var:oci.auth.basic.password}
-              bearer: ${nr-var:oci.auth.bearer}
+              bearer:
+                token: ${nr-var:oci.auth.bearer.token}
     version:
       path: ${nr-sub:packages.nrdot.dir}\\nrdot-collector.exe
       args:
@@ -233,7 +236,8 @@ deployment:
               basic:
                 username: ${nr-var:oci.auth.basic.username}
                 password: ${nr-var:oci.auth.basic.password}
-              bearer: ${nr-var:oci.auth.bearer}
+              bearer:
+                token: ${nr-var:oci.auth.bearer.token}
     health:
       interval: 30s
       initial_delay: 90s
--- a/agent-control/src/agent_type/runtime_config/on_host.rs
+++ b/agent-control/src/agent_type/runtime_config/on_host.rs
@@ -741,11 +741,6 @@ executables:
             password: "",
             bearer: "token",
         };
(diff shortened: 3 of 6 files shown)


alvarocabanas (dismissed):
LGTM, I like to use basic if both are there, but we may ask product about their thoughts

danielorihuela:
> LGTM, I like to use basic if both are there, but we may ask product about their thoughts

Last commit changes that behaviour. Now we throw an error if multiple auth methods are configured or if the basic auth method is configured but one of the values is empty.

Not sure if this is best. The change was motivated by @sigilioso. It's up for debate.

agent-control/agent-type-registry/newrelic/com.newrelic.opentelemetry.collector-0.1.0.yaml
```diff
@@ -190,6 +229,11 @@ deployment:
             repository: ${nr-var:oci.repository}
             version: ${nr-var:version}
             public_key_url: https://publickeys.newrelic.com/g/agent-control-oci/global/nrdot/jwks.json
+            auth:
+              basic:
+                username: ${nr-var:oci.auth.basic.username}
+                password: ${nr-var:oci.auth.basic.password}
+              bearer: ${nr-var:oci.auth.bearer}
```
alvarocabanas, line 236: There is a design decission that conditions the later design for the OCI Client.
We are setting one of this blocks per package, with it's own auth section. 
This makes it possible to have a different auth per package, but we are only defining one oci section above, but a user could define :
```
oci_1: 
   ....
oci_2:
   ....
```
And one package use oci_1 and the other oci_2, so this requires one OCI Client for each package of each agent.

Are we over complicating it? Are really users going to set more than one oci auth per agent for their different packages or is it something unlikely?

danielorihuela: I thought about other alternatives, but I didn't like them. This approach is more consistent I think. Let me elaborate!

First, let me paste here an example of a `packages` block.

```yaml
packages:
  infra-agent:
    download:
      oci:
        registry: ${nr-var:oci.registry}
        repository: ${nr-var:oci.repository}
        version: ${nr-var:version}
        public_key_url: https://publickeys.newrelic.com/g/agent-control-oci/global/nrinfraagent/jwks.json
        auth:
          basic:
            username: ${nr-var:oci.auth.basic.username}
            password: ${nr-var:oci.auth.basic.password}
          bearer: ${nr-var:oci.auth.bearer}
```

**auth block per `packages` section**

We could have a single auth block per `packages` section. However, there's nothing stopping the agent type creator of using different registries for the different packages. In that situation, we might need to use a different authentication mechanism. For example, a client might be okay using public newrelic agents (anonymous auth) and private agents on a private registry (with basic auth). I don't know if that's an actual use case that we want to support, but I had it in mind.

I discarded that option because it potentially made things harder in the future and I didn't think it was particularly consistent. Registry is under `oci` but `auth` would be above under `packages`. They are configuring "the same thing", but they are far apart and asymetric (one works per package, and the other is global).

```yaml
packages:
  auth:
    basic: ...
    bearer: ...
  infra-agent:
    download:
      oci:
        registry: ...
```

**auth block per `download` or `package` section**

No advantage. Is the same as having it under `oci`. The disadvantage is that we are configuring "the same thing", but they are far apart.

**breaking change - assume one registry per agent type is going to be used**

At some point I also thought... "pufff, we might be complicating things and preparing the code for use cases that will never come". So I thought! What if we moved the `oci` section just below `packages`. Like this.

```yaml
packages:
  oci:
    registry: ...
    public_key: ...
    auth:
      basic: ...
      bearer: ...
  infra-agent:
    download:
      package:
        repository: ...
        version: ...
```

In this new interface, we configure for all packages the registry info. Then, for each package we configure the "package" information in the new section called `package`.

I liked the shape of that solution (from an agent creator point of view), BUT it has several disadvatanges. 

1. Breaking change (we are public preview, we could break it, but we have some users playing with it so I guess it's better not to do it)
2. If we ever need to support multiple registries per agent type, we will need to add the older block back again under the new `package` section
3. AC code would get complex -> Check if the specific config is empty or not, if it's empty use the general, etc

I discarded it too.

---

These are the alternatives I came up with during the development. If you have any other idea of how we might simplify it @alvarocabanas, I'm all ears!!!

alvarocabanas: Yes, this tow last approaches seem good as long as we are sure we only use one oci for the whole packages (that is the actual behavior). We need to discuss it carefully.

sigilioso: On top of that, we may also consider a global registry definition (including authentication) that is used in every package unless it is overwritten. Something similar to what is done in [helm-charts' common library](https://github.com/newrelic/helm-charts/blob/master/library/common-library/templates/_images.tpl#L5) where `global.images.registry` is supported.

agent-control/agent-type-registry/newrelic/com.newrelic.opentelemetry.collector-0.1.0.yaml
paologallinaharbur, line 1: Nete that we have a `name: io.opentelemetry.collector` with this disclaimer `# It should not drift from the `com.newrelic.opentelemetry.collector` definition!`

I am not sure if we should remove it or just "avoid" the drift for now

agent-control/agent-type-registry/newrelic/com.newrelic.infrastructure-0.1.0.yaml
```diff
@@ -45,6 +45,23 @@ variables:
...
+            description: "Username for HTTP Basic authentication"
+            type: string
+            required: false
+            default: ""
+          password:
+            description: "Password for HTTP Basic authentication"
+            type: string
+            required: false
+            default: ""
+        bearer:
```
paologallinaharbur, line 60: should this be 
```yaml
        bearer:
          token:
            description: "Bearer token for authentication"
            type: string
            required: false
            default: ""
```
somehow it feels off that the value of "bearer" is the token itself

paologallinaharbur: This is NOT a strong opinion at all 😅

agent-control/agent-type-registry/newrelic/com.newrelic.infrastructure-0.1.0.yaml
```diff
@@ -186,6 +220,11 @@ deployment:
             repository: ${nr-var:oci.repository}
             version: ${nr-var:version}
             public_key_url: https://publickeys.newrelic.com/g/agent-control-oci/global/nrinfraagent/jwks.json
+            auth:
+              basic:
+                username: ${nr-var:oci.auth.basic.username}
```
paologallinaharbur, line 227: this supports double expansion, right?

I mean.. the variable value could be itself a reference to a secret or an environment variable and gets expanded

agent-control/src/agent_type/runtime_config/on_host/package.rs
```diff
@@ -87,13 +103,48 @@ impl Templateable for Oci {
...
         })
     }
 }
 
+impl Templateable for Auth {
+    type Output = RegistryAuth;
+    fn template_with(self, variables: &Variables) -> Result<Self::Output, AgentTypeError> {
+        let username = self.basic.username.template_with(variables)?;
+        let password = self.basic.password.template_with(variables)?;
+        let token = self.bearer.template_with(variables)?;
```
paologallinaharbur: There is the risk to expose the password/token with this error?
maybe just for this case it is safer to map the error to something fixed

paologallinaharbur: moreover forgetting about the security "issue" without context it is quite difficult to understand which of the 3 is failing since we are not wrapping the errors

agent-control/agent-type-registry/newrelic/com.newrelic.infrastructure-0.1.0.yaml
sigilioso: `bearer_token` could work as well, But I would opt for @paologallinaharbur 's proposal (`bearer.token`)

paologallinaharbur (dismissed):
mostly nits and questions about naming, btw awesome PR description, it really helped!

agent-control/agent-type-registry/newrelic/com.newrelic.infrastructure-0.1.0.yaml
sigilioso: It would be nice to have an explicit unit-test for that use case in the component rendering the variables (maybe there already one in place)

agent-control/src/agent_type/runtime_config/on_host/package.rs
```diff
@@ -87,13 +103,48 @@ impl Templateable for Oci {
...
+
+        match (
+            !&username.is_empty(),
+            !&password.is_empty(),
+            !&token.is_empty(),
```
sigilioso: Is there any particular reason to negate `!` the values? Doesn't it make reasoning a bit harder?

agent-control/agent-type-registry/newrelic/com.newrelic.opentelemetry.collector-0.1.0.yaml
danielorihuela: I think we should keep it. I recovered the following from another conversation:

> Thanks! we decided to keep it in AC for a while to keep supporting K8s old deployments, but we will remove it once we “extract” those agetTypes out of the AC binary

I guess we should support it while we don't get rid of them. Thanks!!!

agent-control/agent-type-registry/newrelic/com.newrelic.infrastructure-0.1.0.yaml
alvarocabanas: But are there going to be more elements than token? I don't think so, I would go for bearer_token, personally.

agent-control/src/agent_type/runtime_config/on_host/package.rs
```diff
@@ -168,5 +220,99 @@ mod tests {
         assert_eq!(rendered_oci.reference.tag(), expected_tag);
         assert_eq!(rendered_oci.reference.digest(), expected_digest);
         assert_eq!(rendered_oci.public_key_url, public_key_url);
+        assert_eq!(rendered_oci.auth, RegistryAuth::Anonymous);
+    }
+
+    #[test]
```
sigilioso, line 244: Could we also add a test-cases for the errors when the user or password are missing?

agent-control/agent-type-registry/newrelic/com.newrelic.opentelemetry.collector-0.1.0.yaml
danielorihuela: Wait, the introduced changes are for on-host. We are keeping "the copy" for k8s. I think we don't have to do anything.

agent-control/src/agent_type/runtime_config/on_host.rs
```diff
@@ -716,6 +719,75 @@ executables:
...
+        const BEARER: Self = Self {
+            username: "",
+            password: "",
+            bearer: "token",
+        };
+    }
+
+    #[rstest]
+    // This case checks backwards compatibility with old agent types not defining auth.
+    #[case::no_auth(PACKAGES_NO_AUTH, AuthCredentials::NONE, RegistryAuth::Anonymous)]
```
sigilioso, line 748: Nice! 💯

sigilioso (dismissed):
I've left some _nits_ feel free to ignore them!

alvarocabanas (dismissed):
LGTM just left a NIT

agent-control/agent-type-registry/newrelic/com.newrelic.infrastructure-0.1.0.yaml
danielorihuela: I'm not sure what's the advantage of having token inside bearer. https://swagger.io/docs/specification/v3_0/authentication/bearer-authentication/

Bearer is a type of token authentication. The only input is a token, isn't it?

paologallinaharbur: no real advanatage, I was just commenting the wording

IMO the value is of the token, not of the "bearer"

agent-control/agent-type-registry/newrelic/com.newrelic.infrastructure-0.1.0.yaml
danielorihuela: This is an "old" feature already present in the codebase. We have `test_render_double_expansion_with_env_variables`. Is that okay?

agent-control/src/agent_type/runtime_config/on_host/package.rs
danielorihuela: > There is the risk to expose the password/token with this error?

I think the text is not printed anywhere. It can fail with a 
* `ValueNotParseableFromString`, which prints the template.
* `MissingTemplateKey`, which prints the "captured var" (e.g. `nr-var:name`)
* `RenderingTemplate`, which prints an error

We should be safe. At least right now, but I'm going to apply your suggestion.

agent-control/agent-type-registry/newrelic/com.newrelic.opentelemetry.collector-0.1.0.yaml
danielorihuela: This was discussed offline a while ago.

We start simple with an oci client for package. Global defaults are discussed on a different PR.

agent-control/agent-type-registry/newrelic/com.newrelic.infrastructure-0.1.0.yaml
danielorihuela: Solved on https://github.com/newrelic/newrelic-agent-control/pull/2414/changes/fd19c5feb6068bae8d684106b6a21d0a8587996d. Used @paologallinaharbur approach just because two persons voted for it, versus one for `bearer_token`.

agent-control/src/agent_type/runtime_config/on_host/package.rs
danielorihuela: Done in https://github.com/newrelic/newrelic-agent-control/pull/2414/changes/e8a9ecb883d6556a0f59615272cdd8d4c2eed69a

agent-control/src/agent_type/runtime_config/on_host/package.rs
danielorihuela: The negation makes harder this first part but makes simpler understanding the combinations of `false` and `true` used inside. At least for me.

I changed it in https://github.com/newrelic/newrelic-agent-control/pull/2414/changes/e8a9ecb883d6556a0f59615272cdd8d4c2eed69a. Let me know what you think!

agent-control/src/agent_type/runtime_config/on_host/package.rs
danielorihuela: https://github.com/newrelic/newrelic-agent-control/pull/2414/changes/d1aa38312bda2dfc48a46d2303f14d3a01e6ca31

sigilioso (approved):
LGTM

PR #2417, opened 2026-04-15
docs: add code style guidelines

gsanchezgavier:
migrate our guidelines in confluence to the repo to facilitate AI coding agents skills contrains unification

### DIFF
--- a/docs/style/naming.md
+++ b/docs/style/naming.md
@@ -1,0 +1,108 @@
+# Naming
+
+## General
+
+- Names **must** be meaningful, well thought, and discussed/agreed when there are doubts.
+- Define a **ubiquitous language** and stick to it.
+- Follow the [Rust API naming guidelines](https://rust-lang.github.io/api-guidelines/naming.html).
+
+## Singular by default
+
+Use singular names for modules and structs:
+
+```rust
+// 👍 Good:
+mod agent_type;
+
+// 👎 Bad:
+mod agent_types;
+```
+
+## Struct and trait names
+
+Struct and trait names should contain the main object they represent. Names should be meaningful even when used outside their defining module:
+
+```rust
+// 👍 Good:
+struct SupervisorStopped;
+trait SupervisorID;
+
+// 👎 Bad:
+struct Stopped;   // unclear outside context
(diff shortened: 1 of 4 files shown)


docs/style/types.md
```diff
@@ -0,0 +1,80 @@
...
+struct Stopped;
+struct Running;
+
+// 👎 Bad: bool field to represent binary state
+struct Agent {
+    running: bool,
+}
+```
+
+Do not use a `bool` field inside a struct to represent binary state. If an enum is truly needed, it is almost always better than a `bool` because it can be extended and both the type and its variants can have explanatory names.
```
DavSanchez: I don't quite get this sentence. I agree that I tend to see the usage of `bool`s as a smell and an enum is almost always a better choice, but not sure if this sentence agrees with that or not.

```suggestion
Do not use a `bool` field inside a struct to represent binary state. An enum is almost always better than a `bool` because it can be extended and both the type and its variants can have explanatory names.
```

docs/style/naming.md
```diff
@@ -0,0 +1,108 @@
...
+for (variable_fqn, end_spec) in agent_type.variables.iter() { ... }
+
+// 👎 Bad:
+for (k, agent_cfg) in agents.iter() { ... }
+for (k, v) in agent_type.variables.iter() { ... }
+```
+
+## Generic type parameter names
+
+Generic parameters should be consistent. Prefer a word matching the trait or combination of traits over single letters:
```
DavSanchez: > `a word matching the trait`

I don't agree with this because can be really misleading.

sigilioso: In fact, we are following a single-letter approach, aren't we?

DavSanchez: Yes! I vaguely remember discussing this back in the day (and disagreeing the same 😅) perhaps it got lost in translation to the guidelines or we naturally gravitated to the single-letter way once we were more confortable.

gsanchezgavier: yesss looks like this one is completely outdated I'll remove it

DavSanchez (approved):

sigilioso (approved):

PR #2423, opened 2026-04-16
fix: add missing crates to AC log

gsanchezgavier:
Noticed that AC was not printed logs from self-replacer create because we forgot to add it to the list. Also auth crate was not added.
This PR adds them but also a mechanism to make this less likely to happen in the future whenever we add a new crate in the workspace.

### DIFF
--- a/agent-control/build.rs
+++ b/agent-control/build.rs
@@ -10,8 +10,25 @@ use std::path::Path;
 const REGISTRY_PATH: &str = "agent-type-registry/";
 const GENERATED_REGISTRY_FILE: &str = "generated_agent_type_registry.rs";
 
+// List of crates whose logs are enabled at the configured level.
+const LOGGING_ENABLED_CRATES: &[&str] = &[
+    "newrelic_agent_control",
+    "self_replacer",
+    "resource_detection",
+    "fs",
+    // External crates (not workspace members) can also be added here freely.
+    "opamp_client",
+    "nr_auth",
+];
+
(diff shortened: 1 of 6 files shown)


agent-control/build.rs
```diff
@@ -50,6 +67,58 @@ fn set_git_commit() {
...
+            .unwrap_or_else(|_| panic!("Could not read {member_path}/Cargo.toml"))
+            .parse()
+            .unwrap_or_else(|_| panic!("Could not parse {member_path}/Cargo.toml"));
+
+        let package_name = member_toml["package"]["name"]
+            .as_str()
+            .unwrap_or_else(|| panic!("package.name not found in {member_path}/Cargo.toml"));
+
+        // Cargo converts hyphens to underscores in crate names; tracing targets use the crate name form.
+        let crate_name = package_name.replace('-', "_");
```
sigilioso, line 104: We need to be careful and never import packages in the workspace with a name different from the crate's name (as we do with _external_ crates `nr-auth --> newrelic-auth-rs` and  `opamp-rs --> newrelic-opamp-rs`.

https://github.com/newrelic/newrelic-agent-control/blob/8eb8cbbb0eb3aca512e070ebb86f419b1feb08cc/agent-control/Cargo.toml#L34-L43

sigilioso (approved):
If we leave the little concern I mentioned above aside, LGTM

Thanks for taking the time to assure we don't forget to include/exclude the crate's logs by default!

PR #2424, opened 2026-04-16
refactor: extract OpAMP settings build into separed functions [NR-537428]

sigilioso:
The OpAMP StartSettings were previously built inline inside start_ac_opamp_client (on-host) and inline in the runner closure (k8s), tightly coupling settings assembly with client startup. This refactor separates the two concerns by introducing dedicated functions in both runners.

The motivation for this refactor was reusing the start-settings to be propagated to the status-http-server and enrich the information the status-http-server provides with these settings information.

### DIFF
--- a/agent-control/src/agent_control/run/k8s.rs
+++ b/agent-control/src/agent_control/run/k8s.rs
@@ -46,7 +46,7 @@ use crate::sub_agent::k8s::builder::SupervisorBuilderK8s;
 use crate::sub_agent::remote_config_parser::AgentRemoteConfigParser;
 use crate::utils::thread_context::StartedThreadContext;
 use crate::{k8s::configmap_store::ConfigMapStore, sub_agent::k8s::builder::K8sSubAgentBuilder};
-use opamp_client::operation::settings::DescriptionValueType;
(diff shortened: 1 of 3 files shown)


agent-control/src/agent_control/run/k8s.rs
```diff
@@ -298,6 +299,26 @@ fn start_cd_version_checker(
     )
 }
 
+/// Builds the OpAMP [StartSettings] for Agent Control on Kubernetes, including the agent
+/// identity and all identifying and non-identifying attributes.
+pub fn build_ac_opamp_start_settings(
```
DavSanchez, line 304: I'm not necessarily against free functions, but we usually use methods to group relevant functions into a particular type. Does `StartSettings::try_build` make sense?

DavSanchez (approved):
Thanks!

agent-control/src/agent_control/run/k8s.rs
sigilioso: The issue is that they are different for onhost and k8s.

I'm merging it as it is and re-work it a bit in the next PR. I've just realized that I cannot use the `agent_description` from OpAMP's start settings as it is.

PR #2427, opened 2026-04-16
feat(win): support self-replace from different drives

gsanchezgavier:
The win implementation of self-replacer did not support the case where the new binary was located in another Drive, as rename does not allow it. I faced this during the integrations test #2407 .This Pr unifies the behavior regarding using a temp intermediate file located next the the current exectuable for the replacement.
Also fixes some missing logs details


sigilioso (approved):
🚀

PR #2432, opened 2026-04-17
feat: add agent-description fields to http-status-server [NR-537428]

sigilioso:
The `/status` HTTP endpoint currently exposes lifecycle and health information for Agent Control and its sub-agents, but nothing about the agents themselves. Their version, instance UID, or other identifying attributes sourced from OpAMP's AgentDescription can be useful for throubleshooting. This PR lays the groundwork by adding a description field (a flexible String → JSON Value map) to both AgentControlStatus and SubAgentStatus. The field is omitted from the JSON response when empty, so there is no observable change for now. A follow-up PR will populate the field from the OpAMP agent description configured at startup.

### DIFF
--- a/agent-control/src/agent_control/http_server/status.rs
+++ b/agent-control/src/agent_control/http_server/status.rs
@@ -12,7 +12,7 @@ use std::time::SystemTime;
 use url::Url;
 
 /// Dynamic fields describing the agent; includes attributes like agent_version, instance_uid, ...
-pub(super) type AgentDescription = serde_json::Map<String, serde_json::Value>;
+pub(super) type AgentAttributes = HashMap<String, String>;
 
 /// Agent Control status and health information.
 /// This information will be shown when the status endpoint is called.
@@ -34,8 +34,8 @@ pub struct AgentControlStatus {
     last_error: Option<String>,
     #[serde(skip_serializing_if = "String::is_empty")]
     status: String,
-    #[serde(skip_serializing_if = "AgentDescription::is_empty")]
-    pub(super) description: AgentDescription,
+    #[serde(skip_serializing_if = "AgentAttributes::is_empty")]
+    pub(super) attributes: AgentAttributes,
 }
 
 impl AgentControlStatus {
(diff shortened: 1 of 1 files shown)


agent-control/src/agent_control/http_server/status.rs
```diff
@@ -6,10 +6,14 @@ use crate::checkers::health::with_start_time::HealthWithStartTime;
 use crate::opamp::{LastErrorCode, LastErrorMessage};
 use crate::sub_agent::identity::AgentIdentity;
 use serde::Serialize;
+
 use std::collections::HashMap;
 use std::time::SystemTime;
 use url::Url;
 
+/// Dynamic fields describing the agent; includes attributes like agent_version, instance_uid, ...
+pub(super) type AgentDescription = serde_json::Map<String, serde_json::Value>;
```
gsanchezgavier: would make sense to call it `attributes` and only support key: val ? or are we expecting to have more complex structures here ? I'm just thinking to have a simple API for any automation that could use this as interface

sigilioso: I started with `Map<String, String>` but then I realized that not everything are string, for instance we already have bools such as `"cd_remote_update: true"`.  Regarding the name, `attributes` is probably a better name!

gsanchezgavier: but for a "generic" field which doesn't have any schema , wouldn't be harder for the consumer this way?

gsanchezgavier: i might be bias by something like [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)

sigilioso: 🤔 it is harder to consume but also consistent with the OpAMP attributes...

sigilioso: I finally changed it: [b9294fc](https://github.com/newrelic/newrelic-agent-control/pull/2432/commits/b9294fcf883c36fe41093dc4463386097a2839bb)

gsanchezgavier (approved):
LGTM

agent-control/src/agent_control/http_server/status.rs
```diff
@@ -447,4 +466,33 @@ pub mod tests {
...
+        };
+        let value = serde_json::to_value(&status).unwrap();
+        assert_eq!(
+            value["attributes"],
+            json!({
+                "version": "1.2.3",
+                "instance_uid": "550e8400-e29b-41d4-a716-446655440000"
+            })
+        );
+    }
```
DavSanchez, line 497: Out of scope, but these two are good candidates for [property-based testing](https://github.com/BurntSushi/quickcheck).

DavSanchez (approved):

PR #2435, opened 2026-04-17
fix(renovate): regex is no longer needed for git dependencies

paologallinaharbur:
git dependencies are now supported by the cargo manager.
Having the regex creates two duplicates PRs
Screenshot 2026-04-17 at 16 22 47


sigilioso (approved):
Thanks

DavSanchez (approved):

PR #2438, opened 2026-04-20
WIP feat: add linux self update e2e

vjripoll:

### DIFF
--- a/test/e2e-runner/src/linux/fake_oci_registry.rs
+++ b/test/e2e-runner/src/linux/fake_oci_registry.rs
@@ -1,0 +1,209 @@
+use std::io::{Read, Write};
+use std::net::{TcpListener, TcpStream};
+use std::sync::Arc;
+use std::thread;
+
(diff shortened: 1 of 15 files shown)


test/e2e-runner/src/linux/fake_oci_registry.rs
```diff
@@ -0,0 +1,209 @@
...
+const ARTIFACT_TYPE: &str = "application/vnd.newrelic.agent.v1";
+const LAYER_MEDIA_TYPE: &str = "application/vnd.newrelic.agent.content.v1.tar+gzip";
+
+/// An in-process OCI registry that serves a single tar.gz package.
+/// Implements just enough of the OCI Distribution Spec for Agent Control to download a package.
+pub struct FakeOciRegistry {
+    pub addr: String,
+}
+
+impl FakeOciRegistry {
```
gsanchezgavier, line 19: would make sense to use a similar approach of the integration test, using zot? i think is already made to work cross platform

PR #2439, opened 2026-04-21
feat: publish agent and sub-agent attributes to HTTP status server [NR-537428]

sigilioso:
## Summary

The `/status` HTTP endpoint previously had no visibility into agent attributes (version, host name, cluster name, instance UID, etc.) Those were assembled only for the OpAMP client and never surfaced elsewhere. This PR wires the AgentDescription built during OpAMP setup into the status server so attributes are always visible, even when OpAMP is disabled.

## Notes for reviewers

* New integration tests included to check that the status-server works as expected (k8s and on-host).
* See the 'out of scope' section

## Out of scope

Attributes updates are not published to the status server for now. We can publish an additional event to the corresponding publisher for AC and sub-agents, but it involves an important refactor over information used to communicate the attributes change because we are currently using the OpAMP's proto types directly (See [attributes.rs](https://github.com/newrelic/newrelic-agent-control/blob/ab78a659473dfc09602fa1aed1abbcdd151b5b4d/agent-control/src/opamp/attributes.rs#L41-L45) for details). This will be addressed in a follow-up PR.

### DIFF
--- a/agent-control/src/agent_control/http_server/status.rs
+++ b/agent-control/src/agent_control/http_server/status.rs
@@ -12,22 +12,35 @@ use std::collections::HashMap;
 use std::time::SystemTime;
 use url::Url;
 
+const IDENTIFYING_ATTRIBUTES_PREFIX: &str = "identifying";
+const NON_IDENTIFYING_ATTRIBUTES_PREFIX: &str = "non-identifying";
+
 /// Dynamic fields describing the agent; includes attributes like agent_version, instance_uid, ...
 pub(super) type AgentAttributes = HashMap<String, String>;
 
 /// Encodes the provided [AgentDescription] as key-val. If there is an overlap in non-identifying
 /// and identifying attributes, the identifying attributes take precedence.
 pub(super) fn build_agent_attributes(agent_description: AgentDescription) -> AgentAttributes {
-    let mut attributes = encode_attributes(agent_description.non_identifying_attributes);
-    attributes.extend(encode_attributes(agent_description.identifying_attributes));
+    let mut attributes = encode_attributes(
+        agent_description.non_identifying_attributes,
+        NON_IDENTIFYING_ATTRIBUTES_PREFIX,
+    );
+    attributes.extend(encode_attributes(
+        agent_description.identifying_attributes,
+        IDENTIFYING_ATTRIBUTES_PREFIX,
+    ));
     attributes
 }
 
 /// Helper to encode attributes from `agent_description` as key-val.
-fn encode_attributes(attributes: HashMap<String, DescriptionValueType>) -> HashMap<String, String> {
+/// The keys are prefixed by `{prefix}/`. Example `"agent.id"` with
+fn encode_attributes(
+    attributes: HashMap<String, DescriptionValueType>,
+    prefix: &str,
+) -> HashMap<String, String> {
     attributes
         .into_iter()
-        .map(|(k, v)| (k.to_string(), encode_description_value(v)))
+        .map(|(k, v)| (format!("{prefix}/{k}"), encode_description_value(v)))
         .collect()
 }
 
@@ -535,22 +548,22 @@ pub mod tests {
     #[case::non_identifying_only(
         HashMap::new(),
         HashMap::from([("k".to_string(), DescriptionValueType::String("v".to_string()))]),
-        vec![("k", "v")]
+        vec![("non-identifying/k", "v")]
     )]
     #[case::identifying_only(
         HashMap::from([("k".to_string(), DescriptionValueType::String("v".to_string()))]),
         HashMap::new(),
-        vec![("k", "v")]
+        vec![("identifying/k", "v")]
     )]
     #[case::merged_no_overlap(
         HashMap::from([("id".to_string(), DescriptionValueType::String("id_val".to_string()))]),
         HashMap::from([("non_id".to_string(), DescriptionValueType::String("non_id_val".to_string()))]),
-        vec![("id", "id_val"), ("non_id", "non_id_val")]
+        vec![("identifying/id", "id_val"), ("non-identifying/non_id", "non_id_val")]
     )]
-    #[case::identifying_takes_precedence_on_overlap(
+    #[case::both_present_when_key_in_both(
         HashMap::from([("shared".to_string(), DescriptionValueType::String("id_value".to_string()))]),
         HashMap::from([("shared".to_string(), DescriptionValueType::String("non_id_value".to_string()))]),
-        vec![("shared", "id_value")]
+        vec![("identifying/shared", "id_value"), ("non-identifying/shared", "non_id_value")]
     )]
     fn test_agent_description(
         #[case] identifying_attributes: HashMap<String, DescriptionValueType>,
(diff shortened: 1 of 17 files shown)


agent-control/src/agent_control/http_server/status.rs
```diff
@@ -14,6 +15,35 @@ use url::Url;
 /// Dynamic fields describing the agent; includes attributes like agent_version, instance_uid, ...
 pub(super) type AgentAttributes = HashMap<String, String>;
 
+/// Encodes the provided [AgentDescription] as key-val. If there is an overlap in non-identifying
+/// and identifying attributes, the identifying attributes take precedence.
+pub(super) fn build_agent_attributes(agent_description: AgentDescription) -> AgentAttributes {
```
paologallinaharbur, line 23: why not dividing them? Returning (AgentAttributes AgentAttributes)

sigilioso: That's definitely another possible approach, I tried to follow an annotation-like approach as discussed here: https://github.com/newrelic/newrelic-agent-control/pull/2432#discussion_r3100180026

Do we prefer to report identifying and non-identifying attributes separately in the http-server response?

paologallinaharbur: if it is a small change I believe it is more correct, but would not lose too much time, right now it is an internal tool

agent-control/src/agent_control/run/k8s.rs
```diff
@@ -111,20 +111,26 @@ impl AgentControlRunner {
...
+        let agent_identity = AgentIdentity::new_agent_control_identity();
+        let opamp_start_settings = build_ac_opamp_start_settings(
+            &instance_id_getter,
+            &agent_identity,
+            &k8s_config,
+            &identifiers,
+        )?;
+
+        self.agent_control_publisher
+            .broadcast(AgentControlEvent::AgentDescriptionSet(
```
paologallinaharbur, line 122: it is a bit weird that we need the opamp_start_settings to initialise the status server that has nothing to do with opamp

agent-control/tests/on_host/scenarios/http_status_server.rs
```diff
@@ -0,0 +1,151 @@
...
+            .ok_or(format!("{OS_ATTRIBUTE_KEY} missing from sub-agent"))?;
+        if got_os != OS_ATTRIBUTE_VALUE {
+            return Err(format!(
+                "expected sub-agent {OS_ATTRIBUTE_KEY}={OS_ATTRIBUTE_VALUE}, got {got_os}"
+            )
+            .into());
+        }
+
+        Ok(())
+    });
```
paologallinaharbur, line 176: maybe I am missing something, but couldn't we try to parse it against an expected struct? it seems that we are checking field by field manually.

an other possibility is to check the text body directly getting rid of somehow of the fields we cannot control

agent-control/tests/k8s/scenarios/http_status_server.rs
```diff
@@ -0,0 +1,166 @@
...
+            .and_then(|v| v.as_str())
+            .ok_or(format!(
+                "{CLUSTER_NAME_ATTRIBUTE_KEY} missing from sub-agent"
+            ))?;
+        if got_sub_cluster != TEST_CLUSTER_NAME {
+            return Err(format!(
+                "expected sub-agent {CLUSTER_NAME_ATTRIBUTE_KEY}={TEST_CLUSTER_NAME}, got {got_sub_cluster}"
+            )
+            .into());
+        }
```
paologallinaharbur, line 181: same here, it seems a complex way to check an answer

agent-control/src/sub_agent/k8s/builder.rs
```diff
@@ -63,24 +63,33 @@ where
...
+                DescriptionValueType::String(self.k8s_config.cluster_name.to_string()),
+            )]),
+        )
+        .map_err(|e| SubAgentBuilderError::OpampClientBuilderError(e.to_string()))?;
+
+        self.sub_agent_publisher
+            .broadcast(SubAgentEvent::AgentDescriptionSet(
+                agent_identity.clone(),
+                opamp_start_settings.agent_description.clone(),
+            ));
```
paologallinaharbur, line 84: same here, it is quite ugly that we need opamp_start_settings without an opamp client, unless it is going to change in the follow up PRs IMO we need to change at least names

agent-control/tests/on_host/scenarios/http_status_server.rs
sigilioso: Yeah, I'll rework it and use the tex-body approach 👍

agent-control/tests/k8s/scenarios/http_status_server.rs
sigilioso: Resolving this, I'm addressing it in the previous comment.

agent-control/src/agent_control/run/k8s.rs
sigilioso: I actually thought of the same but didn't spend type on re-working functions to build the OpAMP's start settings. I'll take a look at it.

agent-control/src/sub_agent/k8s/builder.rs
sigilioso: Let's discuss here https://github.com/newrelic/newrelic-agent-control/pull/2439#discussion_r3118649164, I'm resolving this comment.

agent-control/src/agent_control/http_server/status.rs
sigilioso: What if we do something similar to:

```json
"attributes" {
  "identifying/my-attr": "my-other-att-value",
  "non-identifying/my-other-attr": "my-other-attr-value"
}
```

We would keep the key-val (annotation-like) approach and also keep the identifying/non-identifing info

paologallinaharbur: fine for me!

agent-control/tests/on_host/scenarios/http_status_server.rs
sigilioso: I tried the body approach, but we needed to exclude many fields (timestamps, the hostname, ...) and ended up being more complex. I tried to avoid some repetition by comparing specific parts directly and adding some loops. LMKWYT [f0c0a53](https://github.com/newrelic/newrelic-agent-control/pull/2439/commits/f0c0a535407719d83ecb0f007a7ab4a94399b815)

agent-control/src/agent_control/run/k8s.rs
sigilioso: Addressed here: [ccabbe0](https://github.com/newrelic/newrelic-agent-control/pull/2439/commits/ccabbe070292f923966c7811885e8f3de20e361a). Now when there is no OpAMP, it can't fail due to OpAMP stuff 😄 Thanks!

agent-control/src/agent_control/http_server/status.rs
sigilioso: I've just changed it here: [f28a6b7](https://github.com/newrelic/newrelic-agent-control/pull/2439/commits/f28a6b795a761a76ac39a34b2ea672f9dc561dc0) Let me know what you think!

paologallinaharbur (approved):

PR #2441, opened 2026-04-21
feat: reflect attributes update on http status server [NR-537428]

sigilioso:
PR on top of #2439 

## Summary

This PR wires up the status-server with attributes updates in order to reflect attributes changes such as `version` agents.

## Notes for reviewers

* An internal abstraction to represent attributes updates which was using OpAMP's proto types was replaced by OpAMP's `AgentDescription` directly, making it possible to publish the events that can be consumed by the status-server.
* The unit tests of the HTTP Server's status-updater module were refactored to use `rstest` instead of the previous table-like approach. Such unit-tests are also extended to cover the attributes update new use-case.

### DIFF
--- a/agent-control/src/agent_control/http_server/status.rs
+++ b/agent-control/src/agent_control/http_server/status.rs
@@ -329,6 +329,10 @@ pub mod tests {
                 attributes: Default::default(),
             }
         }
+        pub fn with_attributes(mut self, attrs: HashMap<String, String>) -> Self {
+            self.attributes = attrs;
+            self
+        }
     }
 
     impl SubAgentStatus {
@@ -347,8 +351,9 @@ pub mod tests {
             }
         }
 
-        pub fn agent_id(&self) -> AgentID {
-            self.agent_id.clone()
+        pub fn with_attributes(mut self, attrs: HashMap<String, String>) -> Self {
+            self.attributes = attrs;
+            self
         }
     }
 
@@ -371,22 +376,6 @@ pub mod tests {
     }
 
     impl OpAMPStatus {
-        pub fn new(
-            enabled: bool,
-            endpoint: Option<Url>,
-            reachable: bool,
-            error_code: Option<LastErrorCode>,
-            error_message: Option<LastErrorMessage>,
-        ) -> Self {
-            OpAMPStatus {
-                enabled,
-                endpoint,
-                reachable,
-                error_code,
-                error_message,
-            }
-        }
-
         pub fn enabled_and_reachable(endpoint: Option<Url>) -> Self {
             OpAMPStatus {
                 enabled: true,
--- a/agent-control/src/agent_control/http_server/status.rs
+++ b/agent-control/src/agent_control/http_server/status.rs
@@ -329,9 +329,8 @@ pub mod tests {
                 attributes: Default::default(),
             }
         }
-        pub fn with_attributes(mut self, attrs: HashMap<String, String>) -> Self {
-            self.attributes = attrs;
-            self
+        pub fn with_attributes(self, attributes: HashMap<String, String>) -> Self {
+            Self { attributes, ..self }
         }
     }
 
@@ -351,9 +350,8 @@ pub mod tests {
             }
         }
 
-        pub fn with_attributes(mut self, attrs: HashMap<String, String>) -> Self {
-            self.attributes = attrs;
-            self
+        pub fn with_attributes(self, attributes: HashMap<String, String>) -> Self {
+            Self { attributes, ..self }
         }
     }
 
--- a/agent-control/src/agent_control/http_server/status_updater.rs
+++ b/agent-control/src/agent_control/http_server/status_updater.rs
@@ -128,15 +128,12 @@ mod tests {
     use std::sync::Arc;
     use std::time::{Duration, SystemTime};
 
-    use fake::faker::boolean::en;
-    use fake::faker::filesystem::en::Semver;
-    use fake::faker::lorem::en::{Word, Words};
-    use fake::{Fake, Faker};
+    use opamp_client::operation::settings::{AgentDescription, DescriptionValueType};
+    use rstest::rstest;
     use tokio::runtime::Handle;
     use tokio::sync::RwLock;
     use tokio::sync::mpsc::unbounded_channel;
     use tokio::time::sleep;
-
     use url::Url;
 
     use crate::agent_control::agent_id::AgentID;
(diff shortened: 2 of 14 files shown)


alvarocabanas (commented):
LGTM

alvarocabanas (dismissed):

agent-control/src/agent_control/http_server/status.rs
```diff
@@ -329,6 +329,10 @@ pub mod tests {
                 attributes: Default::default(),
             }
         }
+        pub fn with_attributes(mut self, attrs: HashMap<String, String>) -> Self {
+            self.attributes = attrs;
+            self
+        }
```
DavSanchez: 😁

```suggestion
        pub fn with_attributes(self, attributes: HashMap<String, String>) -> Self {
            Self {
              attributes,
              ..self
            }
        }
```

agent-control/src/agent_control/http_server/status.rs
```diff
@@ -347,8 +351,9 @@ pub mod tests {
             }
         }
 
-        pub fn agent_id(&self) -> AgentID {
-            self.agent_id.clone()
+        pub fn with_attributes(mut self, attrs: HashMap<String, String>) -> Self {
+            self.attributes = attrs;
+            self
```
DavSanchez:
```suggestion
        pub fn with_attributes(self, attributes: HashMap<String, String>) -> Self {
            Self {
              attributes,
              ..self
            }
```

agent-control/src/opamp/attributes.rs
```diff
@@ -1,74 +1,21 @@
...
-            key_value: KeyValue {
-                key: String::from(key.as_ref()),
-                value: Some(AnyValue {
-                    value: Some(Value::StringValue(String::from(value.as_ref()))),
-                }),
-            },
-        }
-    }
-}
+pub type UpdateAttributesMessage = AgentDescription;
```
DavSanchez: Is the alias name missing a `d`? Seems a verb aliasing a noun.


```suggestion
pub type UpdatedAttributesMessage = AgentDescription;
```

sigilioso: The alias was actually already there as it is (I've just changed what it is aliasing). I think that the intention was using 'UpdateAttributes' as something to describe the message. In any case, 'UpdatedAttributesMessages' is clearer, I'm updating it!

sigilioso: Updated: [7244dc4](https://github.com/newrelic/newrelic-agent-control/pull/2441/commits/7244dc43d044fddc0329f610b839d26712ab6ca8)

agent-control/src/agent_control/http_server/status.rs
sigilioso: Updated: [7244dc4](https://github.com/newrelic/newrelic-agent-control/pull/2441/commits/7244dc43d044fddc0329f610b839d26712ab6ca8)

agent-control/src/agent_control/http_server/status.rs
sigilioso: Two `mut`s less! [7244dc4](https://github.com/newrelic/newrelic-agent-control/pull/2441/commits/7244dc43d044fddc0329f610b839d26712ab6ca8)

agent-control/src/opamp/attributes.rs
```diff
@@ -81,33 +28,25 @@ where
...
-    old_agent_description: AgentDescription,
-    new_attributes: Vec<Attribute>,
-) -> AgentDescription {
-    let (new_identifying_attributes, new_non_identifying_attributes): (Vec<_>, Vec<_>) =
-        new_attributes
-            .into_iter()
-            .partition(|attribute| attribute.attribute_type == AttributeType::Identifying);
-
+    old_agent_description: ProtoAgentDescription,
+    new_attributes: AgentDescription,
```
danielorihuela, line 34: Nit: It's a bit strange having a field named `new_attributes` which receives a whole `AgentDescription`. It's also strange having a function `update_agent_description_attributes` receive a whole `AgentDescription`.

danielorihuela: I saw the new struct.

agent-control/src/agent_control/http_server/status_updater.rs
```diff
@@ -75,9 +75,10 @@ async fn update_agent_control_status(
             );
             status.fleet.unreachable(error_code, error_message);
         }
-        AgentControlEvent::AgentDescriptionSet(agent_description) => {
+        AgentControlEvent::AgentDescriptionUpdated(agent_description) => {
             trace!("Setting Agent Control attributes for HTTP-Server");
-            status.agent_control.attributes = build_agent_attributes(agent_description)
+            let attributes_update = build_agent_attributes(agent_description);
+            status.agent_control.attributes.extend(attributes_update);
```
danielorihuela, line 81: The received `agent_description` is supposed to be the final state, right? Should we replace instead of `extend`?

Using `extend` is safer long term, but it feels unnecessary right now.

danielorihuela: I'm probably lacking context, because this is what was done before.

danielorihuela: I see. It's because of 

```rust
AgentControlInternalEvent::AgentControlAttributesUpdated(attributes) => {
    let _ = self.opamp_client.as_ref().map(|c|
        update_opamp_attributes(c, attributes.clone())
    .inspect_err(|e| error!(error = %e, select_arm = "agent_control_internal_consumer", "processing version message")));
    self.agent_control_publisher.broadcast(AgentControlEvent::AgentDescriptionUpdated(attributes));
},
```

Nit: Have you thought about modifying the signature of `update_opamp_attributes` to return the updated agent description?

agent-control/src/opamp/attributes.rs
```diff
@@ -1,74 +1,21 @@
...
+pub type UpdatedAttributesMessage = AgentDescription;
 
 /// Updates the attributes of the OpAMP agent
 ///
 /// If an attribute already exists, it will be updated. If it doesn't, it will be added.
-/// If the `new_attributes` contain duplicated keys, the last occurrence will be kept.
 pub fn update_opamp_attributes<C>(
     opamp_client: &C,
-    new_attributes: Vec<Attribute>,
+    new_attributes: AgentDescription,
```
danielorihuela: Nit: Should we change `new_attributes` to `agent_description`?

agent-control/src/opamp/attributes.rs
danielorihuela, line 34: Nit: Should we change `update_agent_description_attributes` to `merge_agent_descriptions`? And `new_attributes` to `new_agent_description`?

This should be in `opamp-rs` but I don't think it's worth the hassle.

agent-control/src/agent_control/http_server/status_updater.rs
```diff
@@ -155,396 +155,503 @@ mod tests {
...
+                .map(|(k, v)| (k.to_string(), DescriptionValueType::String(v.to_string())))
+                .collect(),
+            non_identifying_attributes: non_identifying
+                .iter()
+                .map(|(k, v)| (k.to_string(), DescriptionValueType::String(v.to_string())))
+                .collect(),
+        }
+    }
+
+    #[rstest]
```
danielorihuela, line 201: I feared all these test cases, but the test itself is quite simple!!! Kudos!!!

agent-control/src/agent_control/http_server/status_updater.rs
sigilioso: I actually considered it and liked the idea of "merging only once" but I finally didn't go for it because `update_opamp_attributes` is only executed when `self.opamp_client` is some. Besides, `c` (the OpAMP client) is what is holding the attributes state in 'proto format': `c.get_agent_description()` returns the `opamp::proto::AgentDescription` while attribute's type is the  'high-level format' (`opamp::operations::AgentDescription`), we would need to decode the proto description!

agent-control/src/opamp/attributes.rs
sigilioso: I've just updated it here: [8a9e3ee](https://github.com/newrelic/newrelic-agent-control/pull/2441/commits/8a9e3ee73699b497510a0661eb37a97e09d97694). I've also updated the method comment to clarify the expectations.

agent-control/src/opamp/attributes.rs
sigilioso: I agree, I've renamed it here. [8a9e3ee](https://github.com/newrelic/newrelic-agent-control/pull/2441/commits/8a9e3ee73699b497510a0661eb37a97e09d97694)

danielorihuela (dismissed):

danielorihuela (approved):

PR #2442, opened 2026-04-21
refactor: simplify render

danielorihuela:
Potential refactor for https://github.com/newrelic/newrelic-agent-control/pull/2436.
This is totally optional and should only be reviewed if the previous PR is already merged.

Now `templating` is "iterative". If a template string results into another template string, then we will keep templating until we reach the final value. This removes the need to template variables first, and then template the `deployment` section.
We still need to template the defaults in it's own step because of type safety.

### DIFF
--- a/agent-control/src/agent_type/render.rs
+++ b/agent-control/src/agent_type/render.rs
@@ -1,6 +1,6 @@
 use crate::agent_type::{
     agent_attributes::AgentAttributes,
-    definition::AgentType,
+    definition::{AgentType, Variables},
     error::AgentTypeError,
     runtime_config::rendered::Runtime,
     templates::Templateable,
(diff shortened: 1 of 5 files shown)


agent-control/src/agent_type/render.rs
alvarocabanas, line 87: I wonder if it makes sense to call it secrets now to be more consistent, still it's a bit weird we have mixed secrets and env_vars here.

agent-control/src/sub_agent/effective_agents_assembler.rs
```diff
@@ -183,21 +185,31 @@ where
...
 
-        let env_vars = load_env_vars();
-
         let mut secret_variables = SecretVariables::try_from(values.clone())?;
         let global_defaults_secrets =
             SecretVariables::try_from(YAMLConfig::from(self.global_defaults.clone()))?;
         secret_variables.merge(global_defaults_secrets);
-        let secrets = secret_variables.load_secrets(&self.secrets_providers)?;
+        let mut secrets = secret_variables.load_secrets(&self.secrets_providers)?;
+        secrets.extend(load_env_vars());
```
alvarocabanas, line 193: So now Secrets includes the env_vars as well, the reason for it is to reduce the number of parameters?

agent-control/src/agent_type/render.rs
danielorihuela: Yeah, I don't know how to feel. At the end, env vars are treated like secrets really. They even implement the same trait.

alvarocabanas: Ok, then I would rename this attribute to secrets

PR #2447, opened 2026-04-22
feat: exit(1467) after windows self-update

gsanchezgavier:
This PR allows windows successful self-update to exit with Win32(1467) so the SCM triggers a service restart, while keeping and exit(0) for linux. 

To do that introduces `GracefulShutdownReason`, an enum that encodes why Agent Control stopped cleanly. 

Currently, `AgentControl::run()`, `run_onhost()`, and `run_k8s()` all returned `Result<(), ...>`, making it impossible for callers to distinguish between a normal external stop request (SIGTERM, SCM stop) and other future graceful exit paths. The enum is threaded through the call stack up to `_main`, where it is discarded for now with.

### DIFF
--- a/agent-control/src/bin/main_onhost.rs
+++ b/agent-control/src/bin/main_onhost.rs
@@ -86,5 +86,7 @@ fn _main(context: Context) -> Result<(), Box<dyn Error>> {
         }
     }
 
-    run_result.map(|_| ())
+    // Shutdown reason is not used passed to this point.
+    run_result?;
+    Ok(())
 }
--- a/agent-control/src/bin/main_onhost.rs
+++ b/agent-control/src/bin/main_onhost.rs
@@ -74,8 +74,7 @@ fn _main(context: Context) -> Result<(), Box<dyn Error>> {
     // Create the actual agent control runner with the rest of required configs
     // and the application_event_consumer and capture the result to report the error in windows
     let run_result = AgentControlRunner::try_new(context.ac_runner_context)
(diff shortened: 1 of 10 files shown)


agent-control/src/bin/main_onhost.rs
```diff
@@ -86,5 +86,5 @@ fn _main(context: Context) -> Result<(), Box<dyn Error>> {
         }
     }
 
-    run_result
+    run_result.map(|_| ())
```
paologallinaharbur: I find this a bit obscure, if we are never using the content of OK() why are we getting it in the first place from run_onhost?

paologallinaharbur: However, I think that something like:
```
run_result?;
Ok(())
```
is cleaner, maybe with a comment

paologallinaharbur: ahh ok, it is used by teardown

paologallinaharbur (dismissed):
I was a bit confused by the notation, I am not sure if a comment or a sliglty different approach would simplify the reading

danielorihuela:
> This PR allows windows successful self-update to exit with Win32(1467) so the SCM triggers a service restart, while keeping and exit(0) for linux.

So windows automatically restarts the service when the exit is 1467?

PD: I just saw on `install.ps1` that restart is triggered for exit codes other than 1. This is just for semantics if I'm not wrong.

danielorihuela (approved):

PR #2448, opened 2026-04-22
feat: global registry support

danielorihuela:
Add support for a single configurable registry.

**Context**

We decided after some discussions to simplify the solution. Initially, we wanted to have a global value to configure the registry, that could then be override in an agent-per-agent basis. This was too complex. We decided to only give support for using one registry at a time, which can be configured.

This PR does that. It adds a global option in agent control to configure the registry, and uses that option to decide from where to download the packages. This is not a final solution, but a "shortcut" to unblock following tasks.

**Technical details**

**`registry` was removed from agent types**

We no longer give support to configure the registry on the agent type configuration. It must be configured from the agent control config. In other words, all packages must be downloaded from the same registry.

**shortcut solution**

Removing the `registry` impacts several parts of the codebase. It's not a simple removal of a field. To avoid blocking other tasks, we are moving ahead with this temporal solution. Basically, we "patch" the registry before doing any operation with the oci downloader.

* The parsed references when parsing the agent type aren't the ones that we will use
* I added a `base.io` registry when templating the references that's not used but complies with the format of a reference

In a different PR, we will modify the code to reflect the correct new structure.

### DIFF
--- a/agent-control/src/agent_control/config.rs
+++ b/agent-control/src/agent_control/config.rs
@@ -76,6 +76,10 @@ pub struct AgentControlConfig {
     /// Contains configuration for on-host self-update mechanism
     #[serde(default)]
     pub self_update: SelfUpdateConfig,
+
+    /// Oci configuration (used for AC and agents packages)
+    #[serde(default)]
+    pub oci: OciConfig,
 }
 
 #[derive(Debug, Default, Deserialize, PartialEq, Clone)]
@@ -92,6 +96,48 @@ pub struct SelfUpdateConfig {
 pub struct PackagesConfig {
     /// Indicates whether package signature verification is enabled or not
     pub signature_verification_enabled: SignatureVerificationEnabled,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
+pub struct OciConfig {
+    #[serde(default = "OciConfig::default_registry")]
+    pub registry: String,
+    #[serde(default)]
+    pub auth: OciAuth,
+}
+
+impl Default for OciConfig {
+    fn default() -> Self {
+        OciConfig {
+            registry: Self::default_registry(),
+            auth: OciAuth::default(),
+        }
+    }
+}
+
+impl OciConfig {
+    fn default_registry() -> String {
+        "docker.io".to_string()
+    }
+}
+
+#[derive(Debug, Deserialize, Serialize, Default, Clone, PartialEq)]
+pub struct OciAuth {
+    #[serde(default)]
+    pub basic: BasicAuth,
+    #[serde(default)]
+    pub bearer: BearerAuth,
+}
+
+#[derive(Debug, Deserialize, Serialize, Default, Clone, PartialEq)]
+pub struct BasicAuth {
+    pub username: String,
+    pub password: String,
+}
+
+#[derive(Debug, Deserialize, Serialize, Default, Clone, PartialEq)]
+pub struct BearerAuth {
+    pub token: String,
 }
 
 const DEFAULT_SIGNATURE_VERIFICATION_ENABLED: bool = true;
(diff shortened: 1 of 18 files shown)


agent-control/src/agent_control/config.rs
```diff
@@ -76,6 +76,10 @@ pub struct AgentControlConfig {
     /// Contains configuration for on-host self-update mechanism
     #[serde(default)]
     pub self_update: SelfUpdateConfig,
+
+    /// Oci configuration (used for AC and agents packages)
```
gsanchezgavier, line 80: is the self update config modification missing ?

gsanchezgavier: but perhaps a separate pr is better

agent-control/src/agent_control/config.rs
```diff
@@ -1041,6 +1087,31 @@ k8s:
...
+        let config_input = r#"agents: {}"#;
+        let config = serde_yaml::from_str::<AgentControlConfig>(config_input).unwrap();
+        assert_eq!("docker.io".to_string(), config.oci.registry);
+
+        let config_input = r#"
+agents: {}
+oci:
+  auth:
+    bearer:
+      token: "token"
```
alvarocabanas, line 1101: So we go for bearer: token: at the end.. ok.

agent-control/src/agent_control/config.rs
danielorihuela: Yes...
I managed to have it working, then I merged @paologallinaharbur changes.

I'll work on the internal struct on a different PR. But at least, with this one, I think @alvarocabanas could take his task, if he wants to.

danielorihuela: I mean. I'm aware we have two configs for the same thing now. I just didn't want to do more changes in this PR.

agent-control/src/package/oci/downloader.rs
gsanchezgavier, line 118: would make sense to change this fn to something that clearly states that the registry will be override?

gsanchezgavier, line 107: same this one

agent-control/src/package/oci/downloader.rs
```diff
@@ -136,6 +140,17 @@ impl OCIArtifactDownloader {
     }
 }
 
+fn reference_with_registry(reference: &Reference, registry: &str) -> Reference {
```
gsanchezgavier, line 143: nit: override_registry(

agent-control/src/agent_control/run/on_host.rs
```diff
@@ -225,6 +226,7 @@ impl AgentControlRunner {
         let agent_control_package_manager = OCIPackageManager::new(
             OCIArtifactDownloader::new(
                 oci_client.clone(),
+                self.bootstrap_config.oci.registry.clone(),
```
gsanchezgavier, line 229: o this pr is already taking car of the ac updater, you might miss to remove the config from it and perhaps something inside the updater itself

agent-control/src/agent_control/config.rs
```diff
@@ -94,6 +98,48 @@ pub struct PackagesConfig {
...
+
+#[derive(Debug, Deserialize, Serialize, Default, Clone, PartialEq)]
+pub struct BasicAuth {
+    pub username: String,
+    pub password: String,
+}
+
+#[derive(Debug, Deserialize, Serialize, Default, Clone, PartialEq)]
+pub struct BearerAuth {
+    pub token: String,
```
gsanchezgavier, line 140: at some point we are going to fetch this from other secured places but i think we can extend this without breaking

gsanchezgavier (commented):
thanks! i just lefts some minor comments

agent-control/src/agent_control/config.rs
paologallinaharbur: > I managed to have it working, then I merged @paologallinaharbur changes.

🤣

alvarocabanas (approved):
LGTM

agent-control/src/agent_control/config.rs
danielorihuela: I'm not sure about that. From where are we going to take that in the future?
We don't support secrets in the agent control config. At least for now.

PR #2449, opened 2026-04-22
fix: garbage collection of configmaps

DavSanchez:
On Agent Control restart, the Kubernetes garbage collector (`retain`) listed all dynamic objects
including fleet-data ConfigMaps. These ConfigMaps share the same agent-ID label as
supervisor-created resources, so they were included in the GC pass — but they lack the
`newrelic.io/agent-type-id` annotation that GC uses to decide whether to retain or delete.

This caused two problems:
1. **Missing annotation → wrong deletion decision**: In `RetainConfig` mode, if an active
   agent's fleet-data ConfigMap had no annotation, the GC would error out.
2. **No annotation was ever written**: `store_remote` / `set_remote_data` never wrote the
   `agent-type-id` annotation onto fleet-data ConfigMaps in the first place.

### Solution

**Write the annotation at creation time**: `ConfigMapStore::set_remote_data` now accepts an
optional `AgentTypeID`. When a sub-agent stores its remote config, the annotation
`newrelic.io/agent-type-id` is written onto the fleet-data ConfigMap. The K8s client's
`set_configmap_key` is extended to accept and merge an annotations map.

**Make GC conservative on missing annotations**: In both `RetainConfig` and `Collect` modes,
if `retrieve_annotated_agent_type_id` returns `MissingAnnotations`, GC now logs a warning and
returns `Ok(false)` (skip) instead of propagating an error. Objects without the annotation are
never accidentally deleted.

**Simplify annotation comparison**: `retrieve_annotated_agent_type_id` returns a raw `String`
instead of a parsed `AgentTypeID`, and comparisons use `.to_string()`. This removes
`AgentTypeIDError` and `AgentIDError` as error variants from `K8sGarbageCollectorError` —
invalid agent IDs in labels are now also logged and skipped rather than surfaced as errors.

### Changes

- `k8s_garbage_collector.rs` — conservative GC logic for missing annotations and invalid agent
  IDs; string-based type comparison; 6 new unit tests covering all retain/collect ×
  annotation-present/absent/mismatch combinations
- `configmap_store.rs` — `set_remote_data` accepts `Option<AgentTypeID>` and forwards the
  annotation to the K8s client
- `client.rs` — `set_configmap_key` (sync + async) accepts an annotations map and merges it
  into the ConfigMap metadata
- `sub_agent.rs` — passes `agent_type_id` from identity when calling `store_remote`
- `config_repository.rs` / `store.rs` — `store_remote` trait and implementations updated to
  thread through `Option<AgentTypeID>`
- `without_flux.rs` — new integration test `k8s_config_map_type_gc_does_not_fail_on_restart`
  that validates the full restart scenario; existing test extended to assert the annotation
  is written

### DIFF
--- a/agent-control/src/agent_control/config.rs
+++ b/agent-control/src/agent_control/config.rs
@@ -609,6 +609,13 @@ pub fn secret_type_meta() -> TypeMeta {
     }
 }
 
+pub fn configmap_type_meta() -> TypeMeta {
+    TypeMeta {
+        api_version: "v1".to_string(),
+        kind: "ConfigMap".to_string(),
+    }
+}
+
 pub fn default_group_version_kinds() -> Vec<TypeMeta> {
     // In flux health check we are currently supporting just a single helm_release_type_meta
     // Each time we support a new version we should decide if and how to support retrieving its health
@@ -617,6 +624,9 @@ pub fn default_group_version_kinds() -> Vec<TypeMeta> {
         // Agent Operator CRD
         instrumentation_v1beta3_type_meta(),
         secret_type_meta(),
+        // This allows ConfigMaps created by sub-agents to be cleaned up by the GC.
+        // AC internal ConfigMaps (e.g. fleet-data) are handled by garbage_collect_agent_control_resources.
+        configmap_type_meta(),
         helmrepository_type_meta(),
         helmrelease_v2_type_meta(),
     ]
--- a/agent-control/src/agent_control/config_repository/store.rs
+++ b/agent-control/src/agent_control/config_repository/store.rs
@@ -7,6 +7,7 @@ use crate::agent_control::config_repository::repository::{
 };
 use crate::agent_control::defaults::{AGENT_CONTROL_CONFIG_ENV_VAR_PREFIX, default_capabilities};
 use crate::opamp::remote_config::hash::ConfigState;
+use crate::resource_ownership::ResourceOwnership;
 use crate::values::config::RemoteConfig;
 use crate::values::config_repository::ConfigRepository;
 use crate::values::yaml_config::YAMLConfigError;
@@ -45,13 +46,21 @@ where
 
     fn store(&self, config: &RemoteConfig) -> Result<(), AgentControlConfigError> {
         self.values_repository
-            .store_remote(&self.agent_control_id, None, config)
+            .store_remote(
+                &self.agent_control_id,
+                ResourceOwnership::AgentControl,
+                config,
+            )
             .map_err(|e| AgentControlConfigError(format!("storing Agent Control config: {e}")))
     }
 
     fn update_state(&self, state: ConfigState) -> Result<(), AgentControlConfigError> {
         self.values_repository
-            .update_state(&self.agent_control_id, state)
+            .update_state(
+                &self.agent_control_id,
+                ResourceOwnership::AgentControl,
+                state,
+            )
             .map_err(|e| AgentControlConfigError(format!("updating Agent Control config: {e}")))
     }
 
@@ -166,6 +175,7 @@ pub(crate) mod tests {
     use crate::agent_control::config_repository::repository::AgentControlConfigLoader;
     use crate::agent_control::config_repository::store::AgentControlConfigStore;
     use crate::agent_type::agent_type_id::AgentTypeID;
+    use crate::resource_ownership::ResourceOwnership;
     use crate::values::config_repository::ConfigRepository;
     use crate::values::config_repository::tests::InMemoryConfigRepository;
     use crate::values::yaml_config::YAMLConfig;
@@ -225,7 +235,11 @@ pub(crate) mod tests {
         .unwrap();
 
         config_repository
-            .store_remote(&AgentID::AgentControl, None, &remote_config.clone().into())
+            .store_remote(
+                &AgentID::AgentControl,
+                ResourceOwnership::AgentControl,
+                &remote_config.clone().into(),
+            )
             .unwrap();
 
         let store = AgentControlConfigStore::new(Arc::new(config_repository));
--- a/agent-control/src/agent_control/config_repository/store.rs
+++ b/agent-control/src/agent_control/config_repository/store.rs
@@ -45,7 +45,7 @@ where
 
     fn store(&self, config: &RemoteConfig) -> Result<(), AgentControlConfigError> {
         self.values_repository
-            .store_remote(&self.agent_control_id, config)
+            .store_remote(&self.agent_control_id, None, config)
             .map_err(|e| AgentControlConfigError(format!("storing Agent Control config: {e}")))
     }
 
@@ -225,7 +225,7 @@ pub(crate) mod tests {
         .unwrap();
 
         config_repository
-            .store_remote(&AgentID::AgentControl, &remote_config.clone().into())
+            .store_remote(&AgentID::AgentControl, None, &remote_config.clone().into())
             .unwrap();
 
         let store = AgentControlConfigStore::new(Arc::new(config_repository));
--- a/agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
+++ b/agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
@@ -208,7 +208,9 @@ impl K8sGarbageCollectorMode<'_> {
                     // Objects without the annotation (e.g. fleet-data ConfigMaps) are not
                     // supervisor-created resources and should not be deleted here.
                     match Self::retrieve_annotated_agent_type_id(obj_meta) {
-                        Ok(annotated_agent_type_id) => Ok(&annotated_agent_type_id != agent_type_id),
+                        Ok(annotated_agent_type_id) => {
+                            Ok(&annotated_agent_type_id != agent_type_id)
+                        }
                         Err(K8sGarbageCollectorError::MissingAnnotations) => Ok(false),
                         Err(e) => Err(e),
                     }
@@ -223,7 +225,9 @@ impl K8sGarbageCollectorMode<'_> {
                     // Objects without the annotation (e.g. fleet-data ConfigMaps) are not
                     // supervisor-created resources and should not be deleted here.
                     match Self::retrieve_annotated_agent_type_id(obj_meta) {
-                        Ok(annotated_agent_type_id) => Ok(annotated_agent_type_id == **agent_type_id),
+                        Ok(annotated_agent_type_id) => {
+                            Ok(annotated_agent_type_id == **agent_type_id)
+                        }
                         Err(K8sGarbageCollectorError::MissingAnnotations) => Ok(false),
                         Err(e) => Err(e),
                     }
@@ -375,7 +379,10 @@ mod tests {
         })
     }
 
-    fn dynamic_object_without_annotation(agent_id: &AgentID, namespace: &str) -> Arc<DynamicObject> {
+    fn dynamic_object_without_annotation(
+        agent_id: &AgentID,
+        namespace: &str,
+    ) -> Arc<DynamicObject> {
         Arc::new(DynamicObject {
             types: None,
             metadata: ObjectMeta {
--- a/agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
+++ b/agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
@@ -202,11 +202,19 @@ impl K8sGarbageCollectorMode<'_> {
         match self {
             K8sGarbageCollectorMode::RetainConfig(agent_identities) => {
                 if let Some(agent_type_id) = agent_identities.get(agent_id) {
-                    let annotated_agent_type_id = Self::retrieve_annotated_agent_type_id(obj_meta)?;
-                    // Check if the agent type is different from the one in the config.
-                    // This is to support the case where the agent id exists in the config,
-                    // but it's a different agent type. See PR#655 for some details.
-                    Ok(&annotated_agent_type_id != agent_type_id)
+                    // Objects without the annotation (e.g. fleet-data ConfigMaps from
+                    // ConfigMapStore) are not supervisor-created and should not be deleted here;
+                    // they are handled by garbage_collection_config_maps via label selector.
+                    match Self::retrieve_annotated_agent_type_id(obj_meta) {
+                        Ok(annotated_agent_type_id) => {
+                            // Check if the agent type is different from the one in the config.
+                            // This is to support the case where the agent id exists in the config,
+                            // but it's a different agent type. See PR#655 for some details.
+                            Ok(&annotated_agent_type_id != agent_type_id)
+                        }
+                        Err(K8sGarbageCollectorError::MissingAnnotations) => Ok(false),
+                        Err(e) => Err(e),
+                    }
                 } else {
                     // Delete if the agent id does not exist in the passed config
                     Ok(true)
@@ -215,8 +223,14 @@ impl K8sGarbageCollectorMode<'_> {
 
             K8sGarbageCollectorMode::Collect(id, agent_type_id) => {
                 if agent_id == *id {
-                    let annotated_agent_type_id = Self::retrieve_annotated_agent_type_id(obj_meta)?;
-                    Ok(annotated_agent_type_id == **agent_type_id)
+                    // Same as above: objects without the annotation are not supervisor-created.
+                    match Self::retrieve_annotated_agent_type_id(obj_meta) {
+                        Ok(annotated_agent_type_id) => {
+                            Ok(annotated_agent_type_id == **agent_type_id)
+                        }
+                        Err(K8sGarbageCollectorError::MissingAnnotations) => Ok(false),
+                        Err(e) => Err(e),
+                    }
                 } else {
                     Ok(false)
                 }
@@ -272,6 +286,7 @@ impl From<K8sGarbageCollectorError> for ResourceCleanerError {
 mod tests {
     use super::*;
     use crate::k8s::client::tests::MockK8sClient;
+    use kube::api::{DynamicObject, ObjectMeta};
     use mockall::predicate;
 
     const TEST_NAMESPACE: &str = "test-namespace";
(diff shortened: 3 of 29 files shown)


agent-control/tests/k8s/tools/k8s_api.rs
```diff
@@ -38,6 +38,29 @@ pub async fn check_config_map_exist(
...
+    k8s_client: Client,
+    name: &str,
+    namespace: &str,
+    annotation_key: &str,
+) -> Result<(), Box<dyn Error>> {
+    let api: Api<ConfigMap> = Api::namespaced(k8s_client.clone(), namespace);
+    let cm = api
+        .get(name)
+        .await
+        .map_err(|err| format!("ConfigMap {name} not found: {err}"))?;
```
danielorihuela: Style: Error message must start with lower case

I don't know if we should write it lowercase like in other places (e.g. `Deleting configmaps using label selector`). Or just use upper case because it's the name of a concept. Up to you.

agent-control/src/agent_control/config_repository/store.rs
```diff
@@ -45,7 +45,7 @@ where
 
     fn store(&self, config: &RemoteConfig) -> Result<(), AgentControlConfigError> {
         self.values_repository
-            .store_remote(&self.agent_control_id, config)
+            .store_remote(&self.agent_control_id, None, config)
```
danielorihuela: I wonder if having a `store_remote_with_agent_type_id` would be cleaner.
Or maybe we should have a `SubAgentControlRepository` similar the the `AgentControlRepository`.
That would make it clear that we use `None` for AC and the agent type for subagents. Or maybe a `K8sConfigRepository`.

The current shape is great as it is. Maybe it's worth working on a refactor after or in the future.

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
```diff
@@ -146,14 +143,16 @@ impl K8sGarbageCollector {
...
             // We must not delete anything with reserved AgentIDs (currently only Agent Control)
-            Err(AgentIDError::Reserved(_)) => return Ok(false),
-            Err(e) => return Err(e.into()),
-        };
-
-        mode.should_delete_agent_id(&agent_id_from_labels, obj_meta)
+            Err(AgentIDError::Reserved(_)) => Ok(false),
+            // We should also be conservative, so we do not delete an object if we cannot retrieve a valid AgentID from it
+            Err(e) => {
+                warn!("invalid agent id: {e}");
```
danielorihuela: Why do we need to be conservative in this case? AC it's installed on it's own namespace. Everything there should have an agent id, shouldn't it?

alvarocabanas (dismissed):
LGTM

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
DavSanchez: The resources might be manipulated from outside (though honestly I'm not sure what for xD). I assume that if some user does this AC should opt for not cleaning the resource.

agent-control/tests/k8s/tools/k8s_api.rs
DavSanchez: Hmm don't remember but maybe I wrote it like this due to being the name of a K8s resource. I'll add something before so I start it lowercase xD

alvarocabanas (dismissed):

agent-control/tests/k8s/tools/k8s_api.rs
danielorihuela: No need. I mean, I don't want that our logs are uglier because we are super strict about the rules. I just mentioned it just in case.

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
danielorihuela: Should we add more information about the object with the invalid agent id? Namespace and name, maybe?

danielorihuela: Maybe we can add that info the the `MissingAnnotations` error

danielorihuela: I see. They shouldn't do it, but if they do, what's the best behaviour? Log warnings and potential AC miss behaviour or throw an error and avoid miss behaviour.

I think I personally prefer throwing an error, but I don't have an strong opinion.

DavSanchez: https://github.com/newrelic/newrelic-agent-control/pull/2449/commits/51b7f49e583b6526d4845bcfd89221655af21cb7

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
gsanchezgavier: have you consider to decouple the internal resources from the agent's resources . I think it will be much easier to understand what's happening under the hood and less error prone. We could add a label to all resources that AC uses to persist data and a different one to the resources created by the supervisor , these functions will be like 
```
self.collect_agent_control_agent_resources
self.collect_sub_agent_resources
```

gsanchezgavier: in the current approach, if I understood correctly , `garbage_collection_config_maps` cleans both the agent and ac resources, and `garbage_collection_dynamic_object` tries to clean them but they were already cleaned , and it is quite hard to understand it

paologallinaharbur: I agree, that was my initiatl though as well

DavSanchez: Indeed I had initially considered introducing another annotation/tag but eventually discarded it. Let me introduce it in a new commit as well as method separation to see how it looks like! I'll share the change here once done so you can take a look.

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
```diff
@@ -52,9 +40,9 @@ impl<C: K8sClient> K8sGarbageCollector<C> {
         active_agents: HashMap<AgentID, AgentTypeID>,
     ) -> Result<(), K8sGarbageCollectorError> {
         let mode = K8sGarbageCollectorMode::RetainConfig(&active_agents);
-        self.garbage_collection_config_maps(&mode)?;
-        self.garbage_collection_dynamic_object(&mode, &self.namespace_agents)?;
-        self.garbage_collection_dynamic_object(&mode, &self.namespace)
+        self.garbage_collect_agent_control_resources(&mode)?;
```
danielorihuela, line 43: Seems like we lost a bit of generalization, but I like this approach!!! 🚀

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
```diff
@@ -71,25 +59,42 @@ impl<C: K8sClient> K8sGarbageCollector<C> {
...
         self.k8s_client
-            .delete_configmap_collection(&self.namespace, &label_selector_query)?;
-
-        Ok(())
+            .list_configmaps(&self.namespace, &label_selector_query)?
+            .into_iter()
+            .filter(|cm| {
+                let empty_map = BTreeMap::new();
+                let annotations = cm.metadata.annotations.as_ref().unwrap_or(&empty_map);
+                annotations::is_owned_by_agent_control(annotations) // should we log the skips?
```
danielorihuela: I don't have a strong opinion. We could add it to `trace` level.

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
```diff
@@ -71,25 +59,42 @@ impl<C: K8sClient> K8sGarbageCollector<C> {
...
+            })
+            .try_for_each(|cm| {
+                let name = cm.metadata.name.as_deref().unwrap_or("unknown");
+                debug!("deleting agent-control ConfigMap: `{name}`");
+                self.k8s_client.delete_configmap(&self.namespace, name)?;
```
danielorihuela: Will this fail when using `unknown`? If yes, should we use a match or any other flow control instead of `unwrap_or`?
Why do we remove by name, if we use the annotations to know what to remove? I'm a bit lost there.

agent-control/src/agent_control/run/k8s.rs
```diff
@@ -209,7 +209,7 @@ impl AgentControlRunner {
         // Cleanup of the existing resources managed by Agent Control but not existing in the
         // config loaded from the first time, for example from previous executions.
         garbage_collector
-            .retain(K8sGarbageCollector::active_config_ids(
+            .retain(K8sGarbageCollector::<SyncK8sClient>::active_config_ids(
```
danielorihuela, line 214: Is this branch up to date? Isn't this related to a PR you made a couple days ago?

agent-control/src/agent_control/config.rs
```diff
@@ -566,6 +573,9 @@ pub fn default_group_version_kinds() -> Vec<TypeMeta> {
         // This allows Secrets created as dynamic objects to be cleaned up by the GC
         // This should not be needed anymore whenever the GC detection logic doesn't rely on this list.
         secret_type_meta(),
+        // This allows ConfigMaps created by sub-agents to be cleaned up by the GC.
+        // AC internal ConfigMaps (e.g. fleet-data) are handled by garbage_collect_agent_control_resources.
+        configmap_type_meta(),
```
danielorihuela, line 629: > // In flux health check we are currently supporting just a single helm_release_type_meta
   // Each time we support a new version we should decide if and how to support retrieving its health

Does this have any implication?

agent-control/src/cli/k8s/install.rs
```diff
@@ -436,10 +436,16 @@ mod tests {
...
-                    "newrelic.io/agent-type-id".to_string(),
-                    agent_identity.agent_type_id.to_string(),
-                )])),
+                annotations: Some(BTreeMap::from_iter(vec![
+                    (
+                        "newrelic.io/agent-type-id".to_string(),
+                        agent_identity.agent_type_id.to_string(),
+                    ),
+                    (
+                        "newrelic.io/owned-by".to_string(),
```
danielorihuela: Is this necessary? It was working without that before. I wonder if we could have just added the types in our code `ResourceOwnership::AgentControl` but avoid all the changes related to the annotations. I'm assuming that agent control always runs on two namespaces.

That said. I think I like it!!!

agent-control/src/resource_ownership.rs
```diff
@@ -0,0 +1,41 @@
...
+/// applied, which would cause resources to be incorrectly handled by the garbage collector.
+#[derive(Debug, Clone, PartialEq)]
+pub enum ResourceOwnership {
+    /// Resource is owned by Agent Control itself (e.g., AC's own fleet-data ConfigMap, instance IDs)
+    AgentControl,
+    /// Resource is owned by a sub-agent (e.g., sub-agent's fleet-data ConfigMap, supervisor-created objects)
+    SubAgent(AgentTypeID),
+}
+
+impl ResourceOwnership {
```
danielorihuela: Are these functions used?

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
```diff
@@ -340,4 +375,346 @@ mod tests {
 
         assert!(garbage_collector.collect(ac_id, agent_type_id).is_ok());
     }
+
+    fn new_dynamic_object(
+        agent_id: &AgentID,
+        agent_type_id: Option<&AgentTypeID>,
+        namespace: &str,
+    ) -> Arc<DynamicObject> {
+        let annotations = agent_type_id.map(|id| Annotations::new_sub_agent_owned(id).get());
```
danielorihuela: Are we missing the `new_agent_type_id_annotation` here?

danielorihuela:
In general, I like the code. I'm not approving it in case anyone wants to take a look!!!

agent-control/src/cli/k8s/install.rs
```diff
@@ -704,10 +716,16 @@ mod tests {
...
-            )]
+            [
+                (
+                    "newrelic.io/agent-type-id".to_string(),
+                    agent_identity.agent_type_id.to_string(),
+                ),
+                (
+                    "newrelic.io/owned-by".to_string(),
+                    "agent-control".to_string(),
+                ),
```
sigilioso: Would it make sense the constants from `k8s::annotations` public in order to use them here?

agent-control/src/on_host/file_store.rs
```diff
@@ -158,7 +159,13 @@ where
         self.get(self.local_dir.get_file_path(agent_id, key))
     }
 
-    fn set_remote_data<T>(&self, agent_id: &AgentID, key: &str, data: &T) -> Result<(), Self::Error>
+    fn set_remote_data<T>(
+        &self,
+        agent_id: &AgentID,
+        _ownership: ResourceOwnership,
```
sigilioso, line 165: I wonder if the k8s storer implementation could directly infer the proper ownership from the provided `agent_id`. We would avoid updating the corresponding trait and the on-host implementation.

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
sigilioso:
```suggestion
                if let Some(name) = cm.metadata.name.as_deref() {
                    debug!("deleting agent-control ConfigMap: `{name}`");
                    self.k8s_client.delete_configmap(&self.namespace, name)?;
                }
                // We could also consider adding a warning if there were resources with no name
                
```

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
```diff
@@ -340,4 +375,346 @@ mod tests {
...
+            .once()
+            .with(
+                predicate::eq(type_meta.clone()),
+                predicate::eq(TEST_NAMESPACE),
+            )
+            .return_once(move |_, _| Ok(vec![cm]));
+        k8s_client
+            .expect_delete_dynamic_object()
+            .once()
+            .returning(|_, _| Ok(either::Either::Right(kube::core::Status::default())));
```
sigilioso: Nit: would it make sense to add a helper to setup the mocks, there are differences between the different tests but they are minimal. We could have something similar to:

```rust
let mut k8s_client = k8s_client_listing_objects(vec![], vec![], vec![cm]);
k8s_client.expec_delete_dynamic_object()
    .once() // ...
```

sigilioso (commented):
I really like the approach! I found a few _nits_ only.

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
DavSanchez: Removed comment.

agent-control/src/agent_control/run/k8s.rs
DavSanchez: Yes it's related, I now need to pass the actual type parameter because the method became generic.

agent-control/src/resource_ownership.rs
DavSanchez: They were on a previous iteration. Removing!

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
sigilioso: I thought this comment https://github.com/newrelic/newrelic-agent-control/pull/2449#discussion_r3186579996 would show up as part of the same thread 😅

agent-control/src/on_host/file_store.rs
DavSanchez: But we cannot get an `AgentTypeID` from an `AgentID` no? It might be possible because we are enforcing that each sub-agent uses a fixed id on the `agents` field of the AC config, but there's no actual guarantee that a particular id maps to a known agent type.

DavSanchez: We can do that with the agent type id

DavSanchez: Well I guess we can make do just knowing that the ownership is _not of agent control_. Checking.

sigilioso: Right! I confused the types (agent id vs. agent type id), we need to add something anyway, so I said nothing! 🤐

sigilioso: > Well I guess we can make do just knowing that the ownership is _not of agent control_. Checking.

If that was enough it would keep the change k8s-scoped

agent-control/src/cli/k8s/install.rs
DavSanchez: yeah they were already public, thanks!

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
DavSanchez: Done, thanks!

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
DavSanchez: I guess we could pass a `ResourceOwnership` value to these functions for their behavior so they keep being generic over ownership, but well, maybe a follow-up PR can propose it!

agent-control/src/cli/k8s/install.rs
DavSanchez: I don't think we can avoid writing an additional annotation without the GC code being ad-hoc for the Agent Control case, I had initially gone through that route but changed after a review with @gsanchezgavier

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
danielorihuela: Nah. I like the current shape. I just wanted to write it somewhere haha.

agent-control/src/agent_control/config.rs
DavSanchez: I don't think so but it might be escaping me. WDYT @sigilioso @paologallinaharbur @gsanchezgavier ?

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
DavSanchez: It was added by the `new_sub_agent_owned` function. Took the time to change these a little bit on https://github.com/newrelic/newrelic-agent-control/pull/2449/commits/c05d8ddabc1f4d87dea47cbfded6733988ba9895.

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
DavSanchez: Applied @sigilioso's suggestion. Thanks!

agent-control/src/agent_control/config.rs
sigilioso: I don't think it has any implication for ConfigMaps. The comment comes to say that we support only one version of the Flux's CRs.

gsanchezgavier: it might create another reflector for config maps , i'm don't remember if that is properly deduplicated by the reflector manager. Apart from that the comment seems correct to me, now that both garbage collect AC and sub-agents are decoupled this make sense to be there.  

IIRC we discussed many times some approaches to remove the list and make it dynamically based on resources that AC have created but we never prioritized to work on it

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
```diff
@@ -52,9 +40,9 @@ impl<C: K8sClient> K8sGarbageCollector<C> {
         active_agents: HashMap<AgentID, AgentTypeID>,
     ) -> Result<(), K8sGarbageCollectorError> {
         let mode = K8sGarbageCollectorMode::RetainConfig(&active_agents);
-        self.garbage_collection_config_maps(&mode)?;
-        self.garbage_collection_dynamic_object(&mode, &self.namespace_agents)?;
-        self.garbage_collection_dynamic_object(&mode, &self.namespace)
+        self.garbage_collect_agent_control_resources(&mode)?;
+        self.garbage_collect_sub_agent_resources(&mode, &self.namespace_agents)?;
+        self.garbage_collect_sub_agent_resources(&mode, &self.namespace)
```
gsanchezgavier, line 45: nice much clear, i think we were exposing impl details at this layer, also logs looks more clear now

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
```diff
@@ -71,25 +59,45 @@ impl<C: K8sClient> K8sGarbageCollector<C> {
...
         self.k8s_client
-            .delete_configmap_collection(&self.namespace, &label_selector_query)?;
-
-        Ok(())
+            .list_configmaps(&self.namespace, &label_selector_query)?
+            .into_iter()
+            .filter(|cm| {
+                let empty_map = BTreeMap::new();
+                let annotations = cm.metadata.annotations.as_ref().unwrap_or(&empty_map);
+                annotations::is_owned_by_agent_control(annotations)
```
gsanchezgavier, line 87: why adding it like an annotation and not a label?

agent-control/src/on_host/file_store.rs
DavSanchez: I think it cannot be done without significantly changing how the agent id and agent type id information is propagated around the code. I think modifying the trait is less invasive than that.

agent-control/src/agent_control/resource_cleaner/k8s_garbage_collector.rs
DavSanchez: The dynamic objects wouldn't be able to leverage label selectors since its current API does not support it right? I would need to add label selector params to `list_dynamic_objects` and so on, so I'd still need to filter on AC side instead of the server.

Maybe I'm missing something and there's a better way to do it?

gsanchezgavier: you're right , didn't check the dynamic

gsanchezgavier (approved):

PR #2450, opened 2026-04-23
Revert "feat: support oci auth in agent types (#2414)"

danielorihuela:
Removes support of oci auth for agent types.

We radically changed how we are going to support mirrors for oci registries. This PR removes the introduced code.


paologallinaharbur (approved):
rip

alvarocabanas (approved):
LGTM

PR #2451, opened 2026-04-23
feat(k8s): introduce explicit health-check definition [NR-552587]

sigilioso:
## Summary

This PR lays the foundation for explicit K8s health-check configuration. Allowing agent type definitions to declare which K8s resources to health-check directly, rather than deriving them implicitly from the objects section.

## Background

Today, health checkers are derived implicitly from the deployment `objects`:

- When there are `HelmRelease` objects it creates HelmReleases + StatefulSet + Deployment + Daemonset health-checkers
- When there are `Instrumentation` objects it creates a single Instrumentation checker.

This coupling means the health check behavior can't be tuned per agent type, and the connection between what is deployed and what is health-checked is implicit.

More importantly, **it prevents Agent Type authors from defining health-check pointing to objects that are not directly defined by the Agent Type**. This is needed for supporting configuration-only Agent Types.

## The approach

The new approach makes health checking explicit in the agent type YAML

**HelmRelease** based agent-types:

```yaml
health:
  interval: 30s
  initial_delay: 30s
  checks:
    - namespace: ${nr-ac:namespace}
      name: ${nr-sub:agent_id}
      kind: HelmReleaseWorkload # Will keep current behavior for helm-releases (HelmRelease + Deployment + Statefulset + Daemonset)
      target_namespace: ${nr-ac:namespace_agents}  # optional; falls back to namespace
```

**Instrumentation** based agent-types:

```yaml
health:
  interval: 30s
  initial_delay: 30s
  checks:
    - namespace: ${nr-ac:namespace_agents}
      name: ${nr-sub:agent_id}
      kind: Instrumentation
```

**Config Only** based agent-types:

```yaml
health:
  interval: 30s
  initial_delay: 30s
  checks:
    - namespace: ${nr-ac:namespace_agents}
      name: ${nr-var:deployment_name}
      kind: Deployment
    - namespace: ${nr-ac:namespace_agents}
      name: ${nr-var:statefulset_name}
      kind: Statefulset
    - namespace: ${nr-ac:namespace_agents}
      name: ${nr-var:daemonset_name}
      kind: Daemonset
```

> [!NOTE]
> It will require an update of existing Agent Types but it doesn't involve a breaking change because Agent Types are still embeded.

Label selectors were also considered but won't be introduced unless we really need them (they would make the Agent Type more flexible but also the implementation slightly more complex).

## The changes

* Add support for new types in Agent Type's health-check section for K8s.
* Introduce a `ResourceFilter` with `name` support for all workload-based health-checkers: `K8sHealthDeamonSet`, `K8sHealthDeployment` and `K8sHealthStatefulSet`.

> [!NOTE]
> There are no behavioral changes, those will come in a **follow-up PR**

## 🚧 Next steps

* Wire the new definition with the health-checkers (update the supervisor and health-checker constructor to use what's defined in the `health` section of the agent type instead of inferring the health-checks from objects.
* Update the existing Agent Types and integration tests
* Update docummentation.

### DIFF
--- a/agent-control/src/agent_type/runtime_config/k8s.rs
+++ b/agent-control/src/agent_type/runtime_config/k8s.rs
@@ -21,7 +21,6 @@ pub struct K8s {
 }
 
 /// A K8s object, usually a CR, to be managed by the agent-control.
-// TODO: at lest, the spec should be templatable.
 #[derive(Debug, Deserialize, Default, Clone, PartialEq)]
 pub struct K8sObject {
     #[serde(rename = "apiVersion")]
@@ -50,7 +49,10 @@ impl Templateable for K8s {
                 .into_iter()
                 .map(|(k, v)| Ok((k, v.template_with(variables)?)))
                 .collect::<Result<HashMap<String, K8sObject>, AgentTypeError>>()?,
-            health: self.health,
+            health: self
+                .health
+                .map(|h| h.template_with(variables))
+                .transpose()?,
             version: self.version,
             guid_checker: self.guid_checker,
         })
@@ -85,6 +87,43 @@ impl Templateable for K8sObjectMeta {
     }
 }
 
+/// The kind of Kubernetes resource to health-check.
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+#[serde(rename_all = "PascalCase")]
+pub enum K8sHealthResourceKind {
+    Deployment,
+    DaemonSet,
+    StatefulSet,
+    HelmRelease,
+    Instrumentation,
+}
+
+/// A single Kubernetes resource to include in health checking.
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+pub struct K8sHealthResource {
+    pub(crate) name: String,
+    pub(crate) namespace: String,
+    pub(crate) kind: K8sHealthResourceKind,
+    /// This field allows referencing related resources in a different namespace. Eg:
+    /// `HelmRelease -> Deployment`, defaults to `namespace`.
+    pub(crate) target_namespace: Option<String>,
+}
+
+impl Templateable for K8sHealthResource {
+    type Output = Self;
+    fn template_with(self, variables: &Variables) -> Result<Self, AgentTypeError> {
+        Ok(Self {
+            name: self.name.template_with(variables)?,
+            namespace: self.namespace.template_with(variables)?,
+            kind: self.kind,
+            target_namespace: self
+                .target_namespace
+                .map(|ns| ns.template_with(variables))
+                .transpose()?,
+        })
+    }
+}
+
 #[derive(Debug, Deserialize, Default, Clone, PartialEq)]
 pub struct K8sHealthConfig {
     /// The duration to wait between health checks.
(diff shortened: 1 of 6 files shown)


agent-control/src/agent_type/runtime_config/k8s.rs
```diff
@@ -85,6 +87,43 @@ impl Templateable for K8sObjectMeta {
...
 }
 
+/// The kind of Kubernetes resource to health-check.
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+#[serde(rename_all = "PascalCase")]
+pub enum K8sHealthResourceKind {
+    Deployment,
+    DaemonSet,
+    StatefulSet,
+    HelmRelease,
```
gsanchezgavier: would make sense to slightly change this name to reflect that is a composed one ? so is a bit more transparent what is happening under the hood when this is chosen

gsanchezgavier (dismissed):
cool i like the idea! just a left a nit regarding improving the visibility of what HelmRelease health check means

agent-control/src/agent_type/runtime_config/k8s.rs
```diff
@@ -85,6 +87,43 @@ impl Templateable for K8sObjectMeta {
...
+
+/// A single Kubernetes resource to include in health checking.
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+pub struct K8sHealthResource {
+    pub(crate) name: String,
+    pub(crate) namespace: String,
+    pub(crate) kind: K8sHealthResourceKind,
+    /// This field allows referencing related resources in a different namespace. Eg:
+    /// `HelmRelease -> Deployment`, defaults to `namespace`.
+    pub(crate) target_namespace: Option<String>,
```
paologallinaharbur, line 109: This is the only think that I do not 100% like.
The issue is that this field makes sense just for a specifig resource 😕

agent-control/src/agent_type/runtime_config/k8s.rs
paologallinaharbur: I agree, the UI looks like those are k8s resources, but under the hood is more complex.
I believe that that could help to clarify the usage and behaviour of target_namespace

agent-control/src/agent_type/runtime_config/k8s.rs
paologallinaharbur: maybe if "resources" is called "checks" or something similar it will be clearer that those are not specific k8s resources but an AC concept

sigilioso: 🤔 I guess we could try extend the types in order to allow it only when the type is HelmRelease. I'll take a look at it!

agent-control/src/agent_type/runtime_config/k8s.rs
sigilioso: `HelmReleaseWorkload` maybe?

agent-control/src/agent_type/runtime_config/k8s.rs
sigilioso: I've just updated the name to `checks` (and also updated the PR description): [ce768cc](https://github.com/newrelic/newrelic-agent-control/pull/2451/commits/ce768cc796538b3eb909cdc57ba3ae1729c5253f)

agent-control/src/agent_type/runtime_config/k8s.rs
sigilioso: I've just changed it to `HelmReleaseWorkload` and used `checks` instead of `resources` (the PR description is also updated). I also think it improves the yaml! [ce768cc](https://github.com/newrelic/newrelic-agent-control/pull/2451/commits/ce768cc796538b3eb909cdc57ba3ae1729c5253f)

agent-control/src/agent_type/runtime_config/k8s.rs
sigilioso: I finally didn't make the `target_namesapace` and the helm-release thing depending on each other. It made things more complex.

gsanchezgavier (approved):
thanks

PR #2453, opened 2026-04-24
feat: single registry configuration

danielorihuela:
Use single registry for all oci configs

### DIFF
--- a/agent-control/src/agent_control/version_updater/on_host.rs
+++ b/agent-control/src/agent_control/version_updater/on_host.rs
@@ -4,6 +4,7 @@ use crate::agent_control::agent_id::AgentID;
 use crate::agent_control::config::{AgentControlDynamicConfig, AgentControlPackage};
 use crate::agent_control::defaults::AGENT_CONTROL_VERSION;
 use crate::agent_control::version_updater::updater::{UpdaterError, VersionUpdater};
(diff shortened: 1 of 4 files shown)


agent-control/src/agent_control/version_updater/on_host.rs
```diff
@@ -121,11 +121,7 @@ where
         package: AgentControlPackage,
     ) -> Result<Self, BuildError> {
         let base_reference = Reference::from(ReferenceParser::from_str(
-            format!(
-                "{}/{}",
-                package.download.oci.registry, package.download.oci.repository
-            )
-            .as_str(),
+            format!("base.io/{}", package.download.oci.repository).as_str(),
```
gsanchezgavier: could you move the placeholder url to a const with some comments , and also use something under newrelic.com ?

sigilioso (dismissed):
I'd change the placeholder as @gsanchezgavier suggested, that said, LGTM

agent-control/src/agent_control/version_updater/on_host.rs
danielorihuela: If we need to merge that before the release, I'll do it. Otherwise, I'll work on the following up PR, which will remove the "base.io"!!!!

gsanchezgavier (approved):
thanks

PR #2454, opened 2026-04-24
fix: use custom tracing for verify command

gsanchezgavier:
The `verify` subcommand was sharing `build_context` with the run path, which unconditionally initializes file-based and OTel tracing, causing verify's logs to land in the running AC's log files.

The fix extracts a `BootstrapContext` (base paths + bootstrap config, no tracing/pub-sub/signal handlers) so verify can initialize only a stderr layer via the new try_init_stderr_tracing.

### DIFF
--- a/agent-control/src/command.rs
+++ b/agent-control/src/command.rs
@@ -15,7 +15,10 @@ use crate::command::on_host_checks::config::check_config;
 use crate::command::on_host_checks::opamp::check_connectivity;
 use crate::event::ApplicationEvent;
 use crate::event::channel::{EventConsumer, EventPublisher, pub_sub};
-use crate::instrumentation::tracing::{TracingConfig, TracingGuardBox, try_init_tracing};
+use crate::instrumentation::config::logs::config::LoggingConfig;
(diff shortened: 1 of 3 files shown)


sigilioso (dismissed):
Thanks!

agent-control/src/command.rs
```diff
@@ -207,39 +218,51 @@ impl Command {
         }
     }
 
+    fn build_bootstrap_context(args: &Args) -> Result<BootstrapContext, Box<dyn Error>> {
+        let base_paths = BasePaths::default();
+
+        #[cfg(debug_assertions)]
+        let base_paths = set_debug_dirs(base_paths, args);
```
DavSanchez, line 225: Unrelated to this PR but just saw it and wondered why `set_debug_dirs` is not a method of `BasePaths`...

DavSanchez (approved):

agent-control/src/command.rs
gsanchezgavier: yeah it might fit, perhaps it pollutes a bit more the code with more conditional compilation though.

PR #2456, opened 2026-04-24
feat: get oci auth from the config

alvarocabanas:
Use OCI registry auth from config instead of anonymous access.

- The auth is added to OCIArtifactDownloader rather than the client so the client remains reusable across different credential contexts, allowing to use a single client if we support multiple registries with different auth in the  future.                                           
- OciAuth now uses Option fields for basic and bearer, and deserialization actively rejects configurations that set both — ensuring users must choose one or the other rather than relying on silent precedence rules.

### DIFF
--- a/agent-control/src/agent_control/config.rs
+++ b/agent-control/src/agent_control/config.rs
@@ -103,15 +103,14 @@ pub struct PackagesConfig {
 pub struct OciConfig {
     #[serde(default = "OciConfig::default_registry")]
     pub registry: String,
-    #[serde(default)]
-    pub auth: OciAuth,
+    pub auth: Option<OciAuth>,
 }
 
 impl Default for OciConfig {
     fn default() -> Self {
         OciConfig {
             registry: Self::default_registry(),
-            auth: OciAuth::default(),
+            auth: None,
         }
     }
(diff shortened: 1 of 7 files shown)


agent-control/src/agent_control/config.rs
```diff
@@ -117,16 +118,34 @@ impl Default for OciConfig {
...
 pub struct OciAuth {
-    #[serde(default)]
-    pub basic: BasicAuth,
-    #[serde(default)]
-    pub bearer: BearerAuth,
+    pub basic: Option<BasicAuth>,
+    pub bearer: Option<BearerAuth>,
+}
+
+impl<'de> Deserialize<'de> for OciAuth {
```
danielorihuela, line 130: Did you thought of transforming `OciAuth` to an enum and making it optional in `OciConfig`?

alvarocabanas: Done in last commit.

danielorihuela: I thought we could get rid of the custom deserializer. Is this a leftover?

danielorihuela (approved):

PR #2457, opened 2026-04-24
chore(releaseName): threat empty string as None

paologallinaharbur:
This PRs make explicit the fact that the cd and ac release names could be not set.
The previous approach (managing empty strings) was not consistent around the code, possibly hiding subtle bugs

For example I think we had a bug or a confusing logic with the flux health. We were doing the following:
```
        let cd_remote_updates_enabled =
            k8s_config.cd_remote_update && !k8s_config.cd_release_name.is_empty();
let health_checker_builder = agent_control_health_checker_builder([...] 
            cd_remote_updates_enabled.then(|| k8s_config.cd_release_name.clone()),
        );
```
IMO the fact that the cd_remote_update is disabled should not disable as well the health monitoring

### DIFF
--- a/agent-control/src/agent_control/config.rs
+++ b/agent-control/src/agent_control/config.rs
@@ -326,6 +326,15 @@ pub struct OpAMPClientConfig {
     pub signature_validation: SignatureValidatorConfig,
 }
 
+/// Deserializes an empty string as None, otherwise as Some(String)
+fn deserialize_empty_string_as_none<'de, D>(deserializer: D) -> Result<Option<String>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let maybe_string = Option::<String>::deserialize(deserializer)?;
+    Ok(maybe_string.and_then(|s| if s.is_empty() { None } else { Some(s) }))
+}
+
 impl<'de> Deserialize<'de> for OpAMPClientConfig {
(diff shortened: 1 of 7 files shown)


agent-control/src/agent_control/config.rs
```diff
@@ -326,6 +326,15 @@ pub struct OpAMPClientConfig {
     pub signature_validation: SignatureValidatorConfig,
 }
 
+/// Deserializes an empty string as None, otherwise as Some(String)
+fn deserialize_empty_string_as_none<'de, D>(deserializer: D) -> Result<Option<String>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let maybe_string = Option::<String>::deserialize(deserializer)?;
+    Ok(maybe_string.and_then(|s| if s.is_empty() { None } else { Some(s) }))
```
DavSanchez, line 335:
```suggestion
    Ok(maybe_string.filter(|s| !s.is_empty()))
```

DavSanchez (approved):
🚀

agent-control/src/agent_control/config.rs
paologallinaharbur: I am going to change it in the next PR to avoid restarting ci/cd!

DavSanchez: Yes yes! It was just letting know that this was possible, but no strong opinion on changing it 👌

PR #2458, opened 2026-04-24
refactor: rendered oci

danielorihuela:
Rework the rendered oci structure.

**Context**

In https://github.com/newrelic/newrelic-agent-control/pull/2448, we added support for a global registry. However, we didn't rework internal structures. This PRs reworks them so that the rendered oci structure matches with the information passed in the agent type definition.

We also changed intermediate structures and functions. Specifically, some functions now take as input a `PackageData` instead of a `Reference`. The `Reference` is still used in some internal functions.


agent-control/src/oci/reference_parser.rs
```diff
@@ -93,6 +93,19 @@ impl From<ReferenceParser> for Reference {
     }
 }
 
+impl TryFrom<(&str, &str, &str)> for ReferenceParser {
```
alvarocabanas: Does it make sense to make reference, repository and version custom types with validation and here this TryFrom to receive those custom types instead?

alvarocabanas: I see the thing is we delegate the validation to the reference Parser, I don't like it so much, I think would have been better to Validate these fields on the first parse, but maybe now it's a big refactor that doesn't make much sense.

danielorihuela: > Does it make sense to make reference, repository and version custom types with validation

Which validations? We don't validate anything on them when separated.

> I think would have been better to Validate these fields on the first parse

What's the first parse? You mean you preferred it like it was before? That the `Reference` was created when we parsed the agent type definition? But we can't do that anymore because we removed the `registry` field.

alvarocabanas: I see we are not pre-validating this fields because we are doing all in the ReferenceParser that is optimized to not have the Memory leak found by David, so we may continue like this.

danielorihuela: Did a PoC on https://github.com/newrelic/newrelic-agent-control/pull/2467. Not sure it's worth it, or if it would be easier to maintain in the future if we ever move away from OCI or need to support something else.

danielorihuela: It's getting in better shape the code on this other PR. I'm resolving this comment since the debate moved over there.

agent-control/tests/on_host/tools/oci_artifact.rs
```diff
@@ -61,7 +62,13 @@ pub fn push_agent_package(
         )
         .await;
 
-        (blob_digest, index_reference)
+        let oci = Oci {
```
gsanchezgavier: would make sense to build this out of this heleper?

danielorihuela: I thought on that, but then we need to call `testing_unique_tag` in each test. We also need to pass in the repository in each test.

I did that, which is ugly and I couldn't come up with something better and cleaner that's also simple enough.

danielorihuela: If you think it's better, we can go with the explicit path

danielorihuela: Or maybe we can use the structures created in https://github.com/newrelic/newrelic-agent-control/pull/2467 to return a `PackageData` from this function and get what we need from it. Either the `Oci` or re-build the reference.

If you like it. We can postpone this refactor to the other PR.

gsanchezgavier: > but then we need to call testing_unique_tag in each test.

but the tag is inside of the returned reference isn't ?

gsanchezgavier: also the repository, both are inside the returned reference if i'm not missing something.

The reason why i'm asking it, is because i'm planning to move this helper to an external crate and i would like to keep it as decouple of AC as possible

danielorihuela: done in https://github.com/newrelic/newrelic-agent-control/pull/2458/changes/b45ae0913803390d00b774954cff78db77d89cae

danielorihuela: You were right. At some point I forgot I could get the data from the `Reference`.

agent-control/src/package/oci/package_manager.rs
```diff
@@ -276,18 +267,18 @@ fn get_generic_package_location_path(
...
 ///
 /// The sanitization process:
 /// 1. Prepends "oci_" to the filename to avoid reserved filenames (e.g. "CON" on Windows) and
 ///    to prevent the filename from being exactly "." or "..".
 /// 2. Replaces directory separators (`/`, `\`) with `__`.
 /// 3. Replaces any other character that is not alphanumeric, `.`, `-`, `_`, `@`, etc with `_`.
-fn compute_path_suffix(package: &Reference) -> Result<PathBuf, OCIPackageManagerError> {
-    let package_full_reference = package.whole();
+fn compute_path_suffix(package_data: &PackageData) -> Result<PathBuf, OCIPackageManagerError> {
+    let package_full_reference = format!("{}:{}", package_data.repository, package_data.version);
```
gsanchezgavier: this is changing the fs layout , would that affect something? could break existing installations?

agent-control/src/package/oci/package_manager.rs
```diff
@@ -276,18 +267,18 @@ fn get_generic_package_location_path(
...
+/// is a valid filename on both Windows and Unix, and to prevent path traversal or injection.
 ///
 /// The sanitization process:
 /// 1. Prepends "oci_" to the filename to avoid reserved filenames (e.g. "CON" on Windows) and
 ///    to prevent the filename from being exactly "." or "..".
 /// 2. Replaces directory separators (`/`, `\`) with `__`.
 /// 3. Replaces any other character that is not alphanumeric, `.`, `-`, `_`, `@`, etc with `_`.
-fn compute_path_suffix(package: &Reference) -> Result<PathBuf, OCIPackageManagerError> {
-    let package_full_reference = package.whole();
+fn compute_path_suffix(package_data: &PackageData) -> Result<PathBuf, OCIPackageManagerError> {
```
gsanchezgavier, line 280: why this change? we still build a reference to fetch the package (now using the internal registry), couldn't we keep passing it ?

agent-control/src/package/oci/downloader.rs
```diff
@@ -68,6 +78,26 @@ impl OCIAgentDownloader for OCIArtifactDownloader {
...
+    repository: &str,
+    version: &str,
+) -> Result<Reference, OCIDownloaderError> {
+    let mut version = version.to_string();
+    if !version.is_empty() && !version.starts_with('@') {
+        version = format!(":{}", version);
+    }
+    let string_reference = format!("{}/{}{}", registry, repository, version);
+
+    ReferenceParser::from_str(string_reference.as_str())
```
gsanchezgavier: this is the good one 👍

gsanchezgavier (dismissed):
Thanks as we discussed this will remove the registry from the folder name of the package in the fs, but i don't think it will be a big issue as the packages are already separated by agent-id and package-id

agent-control/src/package/oci/package_manager.rs
danielorihuela: Discussed offline.

This would require updating a bunch of `template_with` functions. We decided to let it be.

alvarocabanas (approved):
LGTM

PR #2464, opened 2026-04-27
feat: modify ac service on linux to restart always

vjripoll:
Add RestartForceExitStatus=75 to the systemd unit and make AC exit with code 75 (POSIX TEMPFAIL) when the shutdown reason is SelfUpdate systemd treats exit codes listed in RestartForceExitStatus as "force restart regardless of the Restart= policy", so the updated binary is launched automatically.

The exit-code mapping is handled through a new IntoExitCode trait in Command, which lets each binary's _main return a typed result (Result<GracefulShutdownReason, _> for on-host, Result<(), _> for k8s) and derive the appropriate exit code without duplicating logic or using process::exit. The Windows path is unaffected — it already communicates ERROR_RESTART_APPLICATION (1467) to the SCM via teardown

### DIFF
--- a/agent-control/src/bin/main_onhost.rs
+++ b/agent-control/src/bin/main_onhost.rs
@@ -4,6 +4,7 @@
 //! performing one-shot actions or starting the main agent control process.
 #![warn(missing_docs)]
 use newrelic_agent_control::agent_control::run::AgentControlRunner;
+use newrelic_agent_control::agent_control::run::GracefulShutdownReason;
 use newrelic_agent_control::agent_control::run::on_host::AGENT_CONTROL_MODE_ON_HOST;
 use newrelic_agent_control::command::{Command, Context};
 use newrelic_agent_control::utils::is_elevated::is_elevated;
@@ -56,7 +57,7 @@ fn main() -> ExitCode {
 /// could not read Agent Control config from /invalid/path: error loading the agent control config: \`error retrieving config: \`missing field \`agents\`\`\`
 /// Error: ConfigRead(LoadConfigError(ConfigError(missing field \`agents\`)))
 /// ```
-fn _main(context: Context) -> Result<(), Box<dyn Error>> {
+fn _main(context: Context) -> Result<GracefulShutdownReason, Box<dyn Error>> {
     #[cfg(not(feature = "disable-asroot"))]
     if !is_elevated()? {
         return Err("Program must run with elevated permissions".into());
@@ -86,7 +87,5 @@ fn _main(context: Context) -> Result<(), Box<dyn Error>> {
         }
     }
 
-    // Shutdown reason is not used passed to this point.
-    run_result?;
-    Ok(())
+    run_result
 }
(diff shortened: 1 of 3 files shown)


DavSanchez (dismissed):
I am not necessarily against `Restart=always` (I think it was me who proposed it xD), but I'll just suggest that if we want finer control we can exit the self-replace with a specific exit code and list that exit code on the `RestartForceExitStatus` setting to trigger a restart without having to restart on absolutely all cases.

We might need to revisit this decision if the UX of always restarting needs improvement.

Ref: https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#RestartForceExitStatus=

agent-control/src/bin/main_onhost.rs
```diff
@@ -86,6 +88,15 @@ fn _main(context: Context) -> Result<(), Box<dyn Error>> {
...
+    // without requiring Restart=always (which would also restart on clean shutdowns).
+    #[cfg(target_family = "unix")]
+    if matches!(run_result, Ok(GracefulShutdownReason::SelfUpdate)) {
+        drop(context.tracer);
+        std::process::exit(75);
+    }
+
     // Shutdown reason is not used passed to this point.
     run_result?;
     Ok(())
```
DavSanchez: We can start with this approach but I'd strongly consider either changing the `_main` signature to return an [`ExitCode`](https://doc.rust-lang.org/std/process/struct.ExitCode.html) struct (on both host and K8s) and leverage it with `match` on the different code paths to decide the exit code or `match` directly on `main` like the Windows version is doing and have `_main(Context) -> Result<GracefulShutdownReason, Box<dyn Error>>`.

```rust
fn main() -> ExitCode {
    // all of our logic...
    //
    match run_result {
      Ok(GracefulShutdownReason::SelfUpdate) => ExitCode::from(75),
      Ok(_) => ExitCode::SUCCESS,
      Err(_) => ExitCode::FAILURE, // we can match each error type and return different exit codes
    }
}
```

agent-control/src/command.rs
```diff
@@ -30,6 +28,36 @@ use std::process::ExitCode;
...
+            // On Linux/Unix, exit with TEMPFAIL (75) for self-update so systemd's
+            // RestartForceExitStatus=75 triggers a restart of the newly installed binary.
+            #[cfg(target_family = "unix")]
+            Ok(GracefulShutdownReason::SelfUpdate) => Ok(ExitCode::from(75)),
+            Ok(_) => Ok(ExitCode::SUCCESS),
+            Err(e) => Err(e.to_string()),
+        }
+    }
+}
+
```
DavSanchez, line 60: This has a resemblance to the [`Termination` trait](https://doc.rust-lang.org/std/process/trait.Termination.html), which enables an arbitrary type to be returned from `main` and is already [implemented for `Result<T, E>` types](https://doc.rust-lang.org/std/process/trait.Termination.html#impl-Termination-for-Result%3CT,+E%3E) logging the error. Perhaps we can fine-tune this on a follow-up PR and delegate `match`ing on the value to generate the exit code at the very end of the execution. I'd involve the rest to see if it's useful.

DavSanchez (approved):
Left a comment about fine-tuning the exit code generation, but for now it's good!

PR #2466, opened 2026-04-27
feat(k8s): define health checks explicitly in agent type definitions [NR-552587]

sigilioso:
### Summary

Previously, the K8s health checker was built by introspecting the list of deployed `DynamicObject` resources at runtime: it had to infer which checks to run from the objects' TypeMeta. This coupling between the deployment machinery and health-check logic made it impossible to health-check resources not directly managed by AC (Eg: PCG deployments on config-only scenarios),.

This PR moves to an explicit model: agent type definitions declare a `checks` list under `health` that was already introduced in #2451.

### Changes

- Simplified health-checker construction: the list of resources is no longer needed and it doesn't return an error anymore.
- Updated all 11 built-in agent type YAML definitions to declare their health checks explicitly
- Added `k8s_direct_workload_health_checks` integration test covering named-Deployment health checks (healthy and unhealthy paths)
- Updated `INTEGRATING_AGENTS.md` with new `checks` schema docs and examples for all resource kinds.

### Notes for reviewers

- `HelmReleaseWorkload` preserves the existing 4-checker expansion: the HelmRelease CR itself plus StatefulSet, DaemonSet, and Deployment workloads discovered via the Flux label `helm.toolkit.fluxcd.io/name`. No behavior change for existing Helm-based agents.
- If `checks` is absent, health checking is disabled for that sub-agent. This is not a breaking change because Agent Types are still embedded in the Agent Control source code.


agent-control/src/cli/k8s/install.rs
```diff
@@ -343,14 +344,29 @@ fn check_installation(
...
+        })
+        .filter_map(|obj| {
+            let target_namespace = get_target_namespace(&obj);
+            Some(K8sHealthCheckDefinition {
+                name: obj.metadata.name?,
+                namespace: obj.metadata.namespace?,
+                kind: K8sHealthResourceKind::HelmReleaseWorkload,
+                target_namespace,
+            })
+        })
```
sigilioso, line 361: I also considered a single `filter_map` here, but I found more readable this approach:
- The first keeps helm-releases only
- The second discards the helm-release if it wasn't complete (missing name or namespace)

agent-control/src/sub_agent/k8s/supervisor.rs
```diff
@@ -190,32 +188,31 @@ impl NotStartedSupervisorK8s {
...
-        resources: Arc<Vec<DynamicObject>>,
-    ) -> Result<Option<StartedThreadContext>, SupervisorError> {
+    ) -> Option<StartedThreadContext> {
         let start_time = StartTime::now();
 
         let Some(health_config) = &self.k8s_config.health else {
             debug!("Health checks are disabled for this agent");
-            return Ok(None);
+            return None;
         };
```
DavSanchez, line 192: Also

```rust
if &self.k8s_config.health.is_none() {
    debug!("Health checks are disabled for this agent");
};
let health_config = &self.k8s_config.health?;
```

agent-control/tests/k8s/scenarios/health_checks.rs
```diff
@@ -0,0 +1,166 @@
...
+    let server = FakeServer::start_new();
+
+    let mut k8s = block_on(K8sEnv::new());
+    let ac_ns = block_on(k8s.test_namespace());
+    let agents_ns = block_on(k8s.test_namespace());
+    let tmp_dir = tempdir().expect("failed to create local temp dir");
+
+    // Create the deployment that the health check will monitor by name.
+    // Using 0 replicas: no pods are scheduled, so the deployment is immediately healthy
+    // (0 available == 0 desired, no unavailable), without requiring an image pull.
```
DavSanchez, line 42: 👌 💯

DavSanchez (dismissed):
Thanks!

agent-control/tests/k8s/scenarios/health_checks.rs
```diff
@@ -0,0 +1,166 @@
...
+    block_on(create_deployment(
+        k8s.client.clone(),
+        &agents_ns,
+        "hello-world",
+        0,
+    ));
+
+    let _ac = start_agent_control_with_testdata_config(
+        test_name,
+        CUSTOM_AGENT_TYPE_DIRECT_CHECKS_PATH,
```
paologallinaharbur, line 52: IMO the behaviour of the health check is not correct

We have:
```
        - namespace: ${nr-ac:namespace_agents}
          name: ${nr-sub:agent_id}
          kind: Deployment
        - namespace: ${nr-ac:namespace_agents}
          name: ${nr-sub:agent_id}
          kind: StatefulSet
        - namespace: ${nr-ac:namespace_agents}
          name: ${nr-sub:agent_id}
          kind: DaemonSet
```

how could we say that is healthy if only the deployment id matching?

am I missing something? A not found can be easily mistaken for a healthy

agent-control/tests/k8s/data/custom_agent_type_split_ns.yml
```diff
@@ -13,6 +13,11 @@ deployment:
     health:
       interval: 5s
       initial_delay: 2s
+      checks:
```
paologallinaharbur, line 16: this is identical to custom_agent_type.yml  is it a mistake?

paologallinaharbur (commented):
I like the PR, but I do not like an assumption from a prevoius one.

IMO 
```
            ResourceFilter::ByName(name) => check_health_for_items(
                deployments.into_iter().filter(name_filter(name.clone())),
                Self::check_deployment_health,
            )?,
```
Should fail if none is found

agent-control/tests/k8s/scenarios/health_checks.rs
paologallinaharbur: for context in the health release we are not erroring out if an object is not found since there is the check on the helm-release assuring that! (and that one should fail as well if not found)

sigilioso: I agree, a non-existing resource could be wrongly reported as healthy. However, how could we support use cases such as the PCG one where the agent can be deployed with different workload definitions?

We could allow templating the `kind` field.

```
        - namespace: ${nr-ac:namespace_agents}
          name: ${nr-sub:agent_id}
          kind: ${nr-var:workload_kind}
```

But it would be more complex and harder to use.

agent-control/tests/k8s/data/custom_agent_type_split_ns.yml
sigilioso: Both agent-types are almost identical, I think the checks need to be the same. Maybe we can simplify and keep one of the custom agent types only, we can tackle that in a different PR.

agent-control/tests/k8s/scenarios/health_checks.rs
paologallinaharbur: currently the only one supported is deployment, but it is true that it might change in the future.
I can agree with your reasoning, but we need to make sure to clarify it in the docs, I believe that it could be counterintuitive now

sigilioso: I've just clarified the behavior in the docs: [2cf6b7b](https://github.com/newrelic/newrelic-agent-control/pull/2466/commits/2cf6b7b2c3af47d74a711d3c2dd65595c5affe0a) LMKWYT

sigilioso:
> I like the PR, but I do not like an assumption from a prevoius one.
> 
> IMO
> 
> ```
>             ResourceFilter::ByName(name) => check_health_for_items(
>                 deployments.into_iter().filter(name_filter(name.clone())),
>                 Self::check_deployment_health,
>             )?,
> ```
> 
> Should fail if none is found

Discussed here https://github.com/newrelic/newrelic-agent-control/pull/2466#discussion_r3152672285 (and also _offline_) 🙂

DavSanchez (approved):

PR #2467, opened 2026-04-27
refactor: validate repository version sooner

danielorihuela:
Introduces oci types for `Registry`, `Repository` and `Version`. On top of https://github.com/newrelic/newrelic-agent-control/pull/2458.

Idea from @alvarocabanas.

**Context**

https://github.com/newrelic/newrelic-agent-control/pull/2458 changes several structures to avoid the workaround introduced in the past for allowing a global registry default. In that PR we removed the `Reference` as expected and replaced it with `String`s. The problem with that is that we delay the validation of the data. @alvarocabanas then suggested introducing some "Value Objects".

We created three value objects: `Registry`, `Repository` and `Version`. They validate the data as soon as we obtain it from the agent type definition and from the agent control config.

The only problem is that we are taking the parsing into our own hands now. We can't rely in the oci client crate, because it only has public the `Reference` object.


alvarocabanas (commented):
I like the modifications, but shouldn't also be removed the redundant checks in reference_parser?
We shouldn't need to validate again the repository, registry and version.
Also did you ensure the regex memory leak is not reproduced?

danielorihuela:
> I like the modifications, but shouldn't also be removed the redundant checks in reference_parser?
> We shouldn't need to validate again the repository, registry and version.
> Also did you ensure the regex memory leak is not reproduced?

No, I did nothing because I wanted first to get some feedback whether it was worth it. I'll work on it.

agent-control/src/agent_control/config.rs
```diff
@@ -98,26 +101,43 @@ pub struct PackagesConfig {
...
+
+impl FromStr for Registry {
+    type Err = InvalidRegistry;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s.is_empty() {
+            return Err(InvalidRegistry("registry cannot be empty".to_string()));
+        }
+
+        // Note: we are not validating the registry string since it can be a url or just a hostname, and the rules for both are quite complex.
```
danielorihuela, line 139: I'm not sure which validations should we add for the registry. I couldn't find the specifications of it. I fear it might be something quite complex.

With the current validation, we don't need a custom deserializer. However, if we want something fancier, we will have to add one.

danielorihuela:
@alvarocabanas Now it's done.
Regex memory leak cannot be reproduced cause we are not using the `from_str` or `try_from` of the `Reference` in the oci client crate.

agent-control/src/oci/reference_parser.rs
```diff
@@ -1,89 +1,49 @@
...
+        };
 
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        Self::try_from(s)
+        Self(reference)
     }
 }
 
-impl TryFrom<&str> for ReferenceParser {
+impl TryFrom<(&str, &str, &str)> for ReferenceParser {
```
alvarocabanas: Why not using TryFrom<(Registry, Repository, Version)?

Then you would have it statically typed already knowing those are the right objects. You then leave this let this:
 
```
        let registry =
            Registry::from_str(registry).map_err(|_| ParseError::ReferenceInvalidFormat)?;
        let repository =
            Repository::from_str(repository).map_err(|_| ParseError::ReferenceInvalidFormat)?;
        let version = Version::from_str(version).map_err(|_| ParseError::ReferenceInvalidFormat)?;

        
```

To the caller.

Even if it seems verbose, I think it's the right approach.

danielorihuela: It's only used in tests. We can remove it if needed.

alvarocabanas: Can you add a comment to specify it's only to be used in tests?

sigilioso: Can we move it under `#[cfg(test)]` maybe?

danielorihuela: we need it for integration tests

agent-control/src/oci/reference_parser.rs
```diff
@@ -1,89 +1,49 @@
...
+};
+
 const DEFAULT_TAG: &str = "latest";
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct ReferenceParser(Reference);
 
-impl FromStr for ReferenceParser {
-    type Err = ParseError;
+impl From<(&Registry, &Repository, &Version)> for ReferenceParser {
```
sigilioso: Considering that validation is already in the `package::rendered` module, I wonder if we should move this there as well.

We could have something similar to:

```rust

impl Oci {

    pub fn reference(&self, registry: &Registry) -> Reference {
        // ...
    }

}

```

Do we still need the `ReferenceParser`?

agent-control/src/oci/reference_parser.rs
```diff
@@ -1,89 +1,49 @@
...
-//!
-//! The implementation has been proposed upstream, but this way, we can keep using the library
-//! without resorting to a fork while the proposed changes are discussed and land.
-//!
-//! Some existing private constants and functions have been copied from upstream.
-//!
-//! # ⚠️ DELETE ME ‼️
-//!
-//! This module must be entirely removed from the codebase when
-//! [oci-spec-rs#322](https://github.com/youki-dev/oci-spec-rs/pull/322) is addressed!
```
sigilioso, line 12: The comment doesn't apply anymore, but I'd keep something to remind us why we should not use `TryFrom<&str> for Reference` and the link to the issue.

danielorihuela: Where? Now that we have our own parsing, we should always use that.
It's strange having a comment talking about an issue we no longer have from a function we are no longer using.

Maybe this is one of this things we could put to a `CLAUDE.md` for agent control. I don't know.

sigilioso: 🤔 I was thinking of the function where the `Reference` is built, but you are right, maybe it doesn't make sense anymore considering that we are not parsing from strings.

agent-control/src/oci/reference_parser.rs
sigilioso: Updating the integration tests to use the `Oci` structure would be too much?

danielorihuela: https://github.com/newrelic/newrelic-agent-control/pull/2467/changes/c7b4c5457017a7f971f2b2f73b18559603016e6e

agent-control/src/oci/reference_parser.rs
danielorihuela: https://github.com/newrelic/newrelic-agent-control/pull/2467/changes/c7b4c5457017a7f971f2b2f73b18559603016e6e

alvarocabanas (approved):
LGTM

sigilioso (approved):
🚀

PR #2468, opened 2026-04-28
test: Add oci auth tests scenarios

gsanchezgavier:
Adds two tests scenarios to cover the self-update and remote install with a registry with basic auth


.github/workflows/push_pr_checks_tests.yml
danielorihuela: Nit: This name is wrong I think, can we rename it?

tools/oci-registry.sh
```diff
@@ -45,22 +52,35 @@ else
...
+
+# ----- ARGUMENTS -----
+OPTION="${1:-empty}"
+VERSION="${VERSION:-v2.1.11}"
+PORT="${PORT:-5001}"
+BASIC_AUTH="${BASIC_AUTH:-false}"
+
+# Parse flags from remaining arguments
+shift || true
+for arg in "$@"; do
```
danielorihuela, line 79: Nit: `"${@:2}"` instead of shift

Or maybe even remove that part and just use the env var from the Makefile

tools/oci-registry.sh
```diff
@@ -33,6 +33,13 @@ readonly ARCH
 readonly CONFIG_FOLDER="$HOME/.zot"
 readonly CONFIG_FILE="$CONFIG_FOLDER/config.json"
 readonly BINARY="$CONFIG_FOLDER/zot"
+readonly HTPASSWD_FILE="$CONFIG_FOLDER/htpasswd"
```
danielorihuela: I think the test is failing cause we are not using `cygpath -m` for `HTPASSWD_FILE` like we do for `STORAGE_PATH`. So the path that ends on the config is wrong.

gsanchezgavier: indeed thanks!

danielorihuela:
I was wondering if we should test the `bearer` path too. I was taking a look into `zot` and I'm not sure it's easy.
It would be great though to have the test if it's easy.

gsanchezgavier:
> I was wondering if we should test the bearer path too. I was taking a look into zot and I'm not sure it's easy.
It would be great though to have the test if it's easy.

@danielorihuela I think that we should rely on the tests of the crate itself, these tests covers that we don't messed up in the wirings , specially the ones the cascade the single config to the different components

danielorihuela (approved):

PR #2470, opened 2026-04-28
feat: adding new configuration field cd_enabled

paologallinaharbur:
Suggestions for a better name are welcomed.

The main idea is to provide a unique toggle in order to simplify the life of the user and have a clear way to detect the use-case

Notice that:
```
    /// cd_enabled indicates whether the CD utility is enabled or not.
    /// Agent control disables health checks and automatic upgrade for the CD chart when this field is set to false.
    /// However, it will still try to install any Agent configured without ensuring compatibility.
    /// This responsibility is left to the user and fleet control.
```

### DIFF
--- a/agent-control/src/agent_control/config.rs
+++ b/agent-control/src/agent_control/config.rs
@@ -23,6 +23,7 @@ use kube::api::TypeMeta;
 use serde::{Deserialize, Deserializer, Serialize};
 use std::collections::HashMap;
 use thiserror::Error;
+use tracing::info;
 use url::Url;
 use wrapper_with_default::WrapperWithDefault;
 
@@ -398,6 +399,9 @@ pub struct K8sConfig {
     /// agent_control_cd release name. If not set and cd_enabled is true, AC expect an externally managed CD
     pub cd_release_name: Option<String>,
     /// cd_enabled indicates whether the CD utility is enabled or not.
+    /// Agent control disables health checks and automatic upgrade for the CD chart when this field is set to false.
+    /// However, it will still try to install any Agent configured without ensuring compatibility.
+    /// This responsibility is left to the user and fleet control.
     pub cd_enabled: bool,
     /// Specifies the key name within the Kubernetes Secret
     /// used to retrieve the required secret for credentials.
@@ -435,6 +439,9 @@ impl<'de> Deserialize<'de> for K8sConfig {
         if let Some(cd_enabled) = intermediate.cd_enabled
             && !cd_enabled
         {
+            info!(
+                "CD is disabled, setting cd_release_name to None, cd_remote_update to false and removing flux and instrumentation types from cr_type_meta"
+            );
             intermediate.cd_remote_update = Some(false);
             intermediate.cd_release_name = None;
             intermediate.cr_type_meta = Some(
--- a/agent-control/src/agent_control/config.rs
+++ b/agent-control/src/agent_control/config.rs
@@ -326,15 +326,6 @@ pub struct OpAMPClientConfig {
     pub signature_validation: SignatureValidatorConfig,
 }
 
-/// Deserializes an empty string as None, otherwise as Some(String)
-fn deserialize_empty_string_as_none<'de, D>(deserializer: D) -> Result<Option<String>, D::Error>
-where
-    D: Deserializer<'de>,
-{
-    let maybe_string = Option::<String>::deserialize(deserializer)?;
-    Ok(maybe_string.and_then(|s| if s.is_empty() { None } else { Some(s) }))
-}
-
 impl<'de> Deserialize<'de> for OpAMPClientConfig {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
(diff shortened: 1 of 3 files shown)


agent-control/src/agent_control/config.rs
```diff
@@ -395,29 +387,90 @@ pub struct K8sConfig {
...
+        }
+
+        let mut intermediate = IntermediateK8sConfig::deserialize(deserializer)?;
+
+        // If the release names are set to empty strings we leverage None instead
+        intermediate.cd_release_name = intermediate.cd_release_name.filter(|s| !s.is_empty());
+        intermediate.ac_release_name = intermediate.ac_release_name.filter(|s| !s.is_empty());
+
+        // if cd_enabled is false, we make sure that cd_release_name is None, cd_remote_update is false
+        // and we set the cr_type_meta to a default value without flux and instrumentation types (since flux won't be installed)
```
sigilioso:
```suggestion
        // if cd_enabled is false, we make sure that cd_release_name is None, cd_remote_update is false
        // and cr_type_meta is set to a default value without flux and instrumentation types (since flux won't be installed) unless it its set, in that case it is honored.
```

agent-control/src/agent_control/config.rs
```diff
@@ -395,29 +387,90 @@ pub struct K8sConfig {
...
+            intermediate.cd_remote_update = Some(false);
+            intermediate.cd_release_name = None;
+            intermediate.cr_type_meta = Some(
+                intermediate
+                    .cr_type_meta
+                    .unwrap_or_else(default_group_version_kinds_no_flux),
+            );
+        } else {
+            intermediate.cd_enabled = Some(true);
+        }
```
sigilioso, line 456: Should we also move here the `cr_type_meta` defaulting to the values with flux and instrumentation? It would be nice to have all that logic together. Maybe we could remove the `Default` implementation for `K8sConfig`.

agent-control/src/agent_control/config.rs
```diff
@@ -395,29 +387,90 @@ pub struct K8sConfig {
...
+        // if cd_enabled is false, we make sure that cd_release_name is None, cd_remote_update is false
+        // and we set the cr_type_meta to a default value without flux and instrumentation types (since flux won't be installed)
+        if let Some(cd_enabled) = intermediate.cd_enabled
+            && !cd_enabled
+        {
+            info!(
+                "CD is disabled, setting cd_release_name to None, cd_remote_update to false and removing flux and instrumentation types from cr_type_meta"
+            );
+            intermediate.cd_remote_update = Some(false);
+            intermediate.cd_release_name = None;
```
sigilioso, line 447: Considering that these two values are ignored, I wonder if a warning in case they are set would worth it:

```rust
if let Some(cd_remote_update) = intermediate.cd_remote_update && cd_remote_update {
    warn!("'cd_remote_update' value is ignored because cd is not enabled");
}
```

agent-control/src/agent_control/config.rs
```diff
@@ -920,6 +976,34 @@ k8s:
...
+        assert!(!k8s.cd_enabled);
+        assert_eq!(k8s.cd_release_name, None);
+        assert!(!k8s.cd_remote_update);
+        assert!(
+            k8s.cr_type_meta
+                .iter()
+                .filter(|o| o.kind == "HelmRelease" || o.kind == "Instrumentation")
+                .collect::<Vec<_>>()
+                .is_empty()
+        );
```
sigilioso, line 1002: Nit:

```suggestion
        assert!(
            !k8s.cr_type_meta
                .iter()
                .any(|o| o.kind == "HelmRelease" || o.kind == "Instrumentation")
        );
```

agent-control/src/cli/k8s/uninstall/agent_control.rs
```diff
@@ -81,6 +81,8 @@ fn delete_owned_objects(
 /// objects_to_delete retrieves the static list of object known by AC, ignoring any dynamic list.
 /// Moreover, it adds ConfigMap to the list since it is not part of the default_group_version_kinds().
 /// it also filters away object that are not available in the cluster.
+/// On top of it, in the fluxless scenarios it loads the HelmRelease and HelmRepository CRs to be deleted,
+/// but it it a noop since they are not actually present in the cluster.
```
sigilioso, line 85: 🤔 Does it work well even when the CRDs don't exist?

sigilioso (commented):
Looks good! I've left some nits and questions

agent-control/src/cli/k8s/uninstall/agent_control.rs
paologallinaharbur: yes we are removing those according to     `let kinds_available = retrieve_api_resources(&k8s_client)?;`

It is the same it happens with the NR instrumentation objects

agent-control/src/agent_control/config.rs
paologallinaharbur: it is heavily used in tests I think, I can move it under tests, but I wound not remove it completely

agent-control/src/agent_control/config.rs
paologallinaharbur: there is already the info! saying that, I believe it would be "too much"

sigilioso (approved):
LGTM

PR #2472, opened 2026-04-29
feat: health fluxless

paologallinaharbur:
This PR add a NOOP for the agent control health check whenever flux is not enabled. Due to rust I needed to include a new enum to return either one or the other implementation to avoid a box of a box


sigilioso (approved):
LGTM

PR #2473, opened 2026-04-29
fix: rollback restart linux changes and back to restart always

vjripoll:
Rollback restart changes for linux self upgrade


sigilioso (approved):
LGTM

PR #2474, opened 2026-04-29
refactor: extract fake opamp server

vjripoll:
- Moves the fake OpAMP server implementation out of agent-control's integration test helpers and into a dedicated crate at test/fake-opamp-server/, following the same convention that test/e2e-runner/ uses. 

- test/fake-opamp-server does not depend on newrelic_agent_control — that would create a circular dependency, since newrelic_agent_control dev-depends on fake-opamp-server. Instead, fake-opamp-server exposes InstanceUid directly from opamp-client

- agent-control/tests/common/opamp.rs is kept as a thin wrapper — it bridges the InstanceID type used across AC integration tests with the InstanceUid key the new crate uses. Removing it would require changing every integration test that calls FakeServer.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ resolver = "3"
 
 members = [
   "agent-control",
+  "test/crates/fake-opamp-server",
   "resource-detection",
   "fs",
   "wrapper_with_default",
@@ -44,6 +45,10 @@ reqwest = { version = "0.13.3", features = ["blocking", "socks"] }
 rstest = "0.26.1"
 windows = "0.62.2"
 windows-sys = "0.61.2"
+aws-lc-rs = "1.16.3"
+base64 = "0.22.1"
+opamp-client = { git = "https://github.com/newrelic/newrelic-opamp-rs.git", tag = "0.0.37" }
+tokio = { version = "1.52.1", features = ["rt-multi-thread", "macros"] }
 
 [profile.release-debug]
 inherits = "release"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1561,6 +1561,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "fake-opamp-server"
+version = "0.1.0"
+dependencies = [
+ "actix-web",
+ "aws-lc-rs",
+ "base64 0.22.1",
+ "opamp-client",
+ "prost",
+ "serde",
+ "serde_json",
+ "tokio",
+]
+
+[[package]]
 name = "fastrand"
 version = "2.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2933,6 +2947,7 @@ dependencies = [
  "duration-str",
  "either",
  "fake",
+ "fake-opamp-server",
  "flate2",
  "fs",
  "futures",
@@ -2956,7 +2971,6 @@ dependencies = [
  "opentelemetry-semantic-conventions",
  "opentelemetry_sdk",
  "predicates",
- "prost",
  "rcgen",
  "regex",
  "reqwest 0.13.3",
--- a/THIRD_PARTY_NOTICES.md
+++ b/THIRD_PARTY_NOTICES.md
@@ -864,6 +864,11 @@ Distributed under the following license(s):
 * Apache-2.0
 * MIT
 
+## fake-opamp-server <file:///home/runner/work/newrelic-agent-control/newrelic-agent-control/test/crates/fake-opamp-server>
+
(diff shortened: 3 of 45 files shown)


Cargo.toml
```diff
@@ -3,6 +3,7 @@ resolver = "3"
 
 members = [
   "agent-control",
+  "test/fake-opamp-server",
```
sigilioso: We need to align the crate's location @gsanchezgavier introduced `test/crates` in #2477. I'm fine with any option!

agent-control/tests/common/opamp.rs
```diff
@@ -1,364 +1,70 @@
...
-        Self(CustomMessage {
-            capability: SIGNATURE_CUSTOM_CAPABILITY.to_string(),
-            r#type: SIGNATURE_CUSTOM_MESSAGE_TYPE.to_string(),
-            data: serde_json::to_vec(&custom_message_data).unwrap(),
-        })
+impl FakeServer {
+    pub fn start_new() -> Self {
+        Self(fake_opamp_server::FakeServer::start(
+            tokio_runtime().handle(),
+        ))
```
sigilioso: Don't we need to expose the inner Fake server? Maybe a static function could do the trick.

```rust
pub fn start_fake_server() -> fake_opamp_server::FakeServer {
    fake_opamp_server::FakeServer::start(tokio_runtime().handle())
}
```

test/fake-opamp-server/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+
+[dependencies]
+actix-web = "4.13.0"
+aws-lc-rs = "1.16.3"
+base64 = "0.22.1"
+opamp-client = { git = "https://github.com/newrelic/newrelic-opamp-rs.git", tag = "0.0.37" }
+prost = "0.14.3"
+serde = { workspace = true, features = ["derive"] }
+serde_json = { workspace = true }
+tokio = { version = "1.52.1", features = ["rt-multi-thread", "macros"] }
```
sigilioso: Would it make sense to move the shared dependencies to the workspace level in order to assure that agent-control and the fake server use the same?  I think all of them are also used in the agent-control crate.

vjripoll: i moved some of them to workspace execpt actix-web and prost, because are exclusive for fake opamp :)

agent-control/tests/common/opamp.rs
vjripoll: The wrapper is needed because tests use InstanceID while fake_opamp_server::FakeServer uses InstanceUid internally. 
Without the wrapper, that conversion would need to happen at every call site across ~30 test files.
```
  let agent_control_instance_id: InstanceID = get_instance_id(...);

  opamp_server.set_config_response(agent_control_instance_id.clone(), agents.as_str());

  Without the wrapper, every call site would need an explicit conversion:

  opamp_server.set_config_response(InstanceUid::from(agent_control_instance_id.clone()), agents.as_str());
```

Cargo.toml
vjripoll: align in test/crates :) 👍

agent-control/tests/common/opamp.rs
sigilioso: What if the opamp server received a type that knows how to be converted to `InstanceUid`?

```rust
pub fn set_config_response(instance_id: impl Into<InstanceUid>, ...)
```

gsanchezgavier (dismissed):
LGTM

agent-control/tests/common/opamp.rs
```diff
@@ -1,364 +1,2 @@
...
-    }
-    .encode_to_vec()
-}
-
-fn generate_key_pair() -> Ed25519KeyPair {
-    let pkcs8 = Ed25519KeyPair::generate_pkcs8(&SystemRandom::new()).unwrap();
-    Ed25519KeyPair::from_pkcs8(pkcs8.as_ref()).unwrap()
-}
+// Re-export FakeServer directly - no wrapper needed
+pub use fake_opamp_server::FakeServer;
```
sigilioso: Nit: do we need this module just to re-export the `FakeServer`?

sigilioso (dismissed):
We could remove the `tests/common/opamp.rs` module completely, but it is just _nit_, feel free to ignore it.

agent-control/tests/common/opamp.rs
vjripoll: yep is deleted and reexported now :)

sigilioso (approved):

PR #2477, opened 2026-04-29
refactor: extracts oci artifact test helpers 

gsanchezgavier:
The plan is to push locally compiled binary to a local OCI registry during e2e execution for self-update tests. This PR moves the helper to a common place to be imported by the e2e runner without the need to import the AC code.


test/crates/oci-package-publisher/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+    format!("{}", hasher.finish())
+}
+
+fn hex_bytes(bytes: &[u8]) -> String {
+    bytes.iter().map(|b| format!("{b:02x}")).collect()
+}
+
+fn blob_digest(data: &[u8]) -> String {
+    format!("sha256:{}", hex_bytes(digest(&SHA256, data).as_ref()))
+}
```
sigilioso: Should we make this public and avoid repeating it in `agent-control/tests/common/oci.rs`?

agent-control/tests/common/oci.rs
```diff
@@ -2,9 +2,7 @@ use aws_lc_rs::digest::SHA256;
 use aws_lc_rs::digest::digest;
 use oci_client::Client;
 use oci_client::Reference;
-use oci_client::config::Architecture;
-use oci_client::config::Os;
-use oci_client::manifest::{IMAGE_CONFIG_MEDIA_TYPE, OciDescriptor};
+use oci_client::manifest::OciDescriptor;
 
 pub async fn push_empty_config_descriptor(
```
sigilioso: If used, should we also move this helper to the new crate?

test/crates/oci-package-publisher/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    /// Pushes `file` as an OCI agent package artifact and returns the index manifest reference.
+    /// The artifact is structured as a manifest index (multiarch) with a
+    /// single entry for the current platform.
+    pub fn push(&self, file: &Path, media_type: PackageMediaType) -> Reference {
+        tokio::runtime::Builder::new_current_thread()
+            .enable_all()
+            .build()
+            .unwrap()
+            .block_on(async {
```
sigilioso: We were sharing [the same runtime](https://github.com/newrelic/newrelic-agent-control/blob/2cf6b7b2c3af47d74a711d3c2dd65595c5affe0a/agent-control/tests/common/runtime.rs#L7) across all tests before. We should't expect a different behavior right?

Tests executed in parallel and the `new_current_thread()` worry me a bit.

test/crates/oci-test-utils/src/publisher.rs
```diff
@@ -0,0 +1,218 @@
...
+                    push_manifest(&client, &index_reference, blob_descriptor, file_name).await;
+
+                push_index(&client, &index_reference, manifest_digest, manifest_size).await;
+
+                index_reference
+            })
+    }
+}
+
+fn unique_tag() -> String {
```
sigilioso: Nit: I'd keep the clarifying comment regarding the way the unique hash is calculated. Now that the `registry_url` is a parameter instead of a constant might also need some clarifications regarding why this is needed in the first place.

sigilioso (commented):
It is nice to see that the code wasn't very coupled even if it was merely a set of test helpers! 🙂 

I've left some questions

test/crates/oci-package-publisher/src/lib.rs
gsanchezgavier: yes i didn't pay much attention to this but it also will make hard to execute this from async, so i will change it to async fn and forget about the runtime here

agent-control/tests/common/oci.rs
gsanchezgavier: it was used by the signer so I sent everything

test/crates/oci-package-publisher/src/lib.rs
gsanchezgavier: as we discussed offline i'm passing the runtime handle , and making al this sync

sigilioso (dismissed):
LGTM

vjripoll (dismissed):

vjripoll (approved):
🚀

PR #2478, opened 2026-04-29
refactor: remove mockall_double in favor of K8sClient trait

DavSanchez:
The `mockall_double` proc-macro on `use` statements (e.g. `#[cfg_attr(test, mockall_double::double)]`) confuses `rust-analyzer`, which frequently reports spurious type errors such as:

```sh
    expected Arc<MockSyncK8sClient, Global>, found Arc<SyncK8sClient, Global>
```

This is distracting in the IDE and can mask real errors.

Instead of swapping the type at the import site via a proc-macro, we now generalize every consumer of `SyncK8sClient` over a new `K8sClient` trait and generate mocks explicitly with `mock!` as we do elsewhere.

Key changes:
- Introduce `pub trait K8sClient: Debug + Send + Sync + 'static` with all 18 instance methods previously on `SyncK8sClient`.
- `impl K8sClient for SyncK8sClient` forwards each call.
- Replace `#[cfg_attr(test, mockall::automock)]` on `impl SyncK8sClient` with a `mock!` block that generates `MockK8sClient` inside `#[cfg(test)]`.
- Generalize structs/functions that store or accept `Arc<SyncK8sClient>` with a generic `C: K8sClient = SyncK8sClient` parameter. This keeps static dispatch and requires zero changes to production call-sites thanks to the default type parameter.
- Remove all 27 `#[cfg_attr(test, mockall_double::double)]` attributes.
- Update every test that relied on `MockSyncK8sClient` to use the new `MockK8sClient`.
- Remove the `mockall_double` dependency from workspace and package Cargo.toml.
- Update `agent-control/tests/k8s/README.md` to document the new pattern.

All 859 unit tests pass. K8s integration tests that exercise real cluster behavior are unaffected because they continue to use `SyncK8sClient` directly.


agent-control/Cargo.toml
```diff
@@ -108,7 +108,6 @@ mockall = { workspace = true }
 schemars = { workspace = true }
 serde_json = { workspace = true }
 assert_matches = { workspace = true }
-mockall_double = { workspace = true }
```
paologallinaharbur, line 111: Rest In Hell

sigilioso (approved):
Thanks for taking the time to get rid of this!

I definitely think this approach is less confusing: with `mockall_double` we were **implicitly** doing something similar that messed up with the developer tools and caused some confusion.

I cannot say that I like the huge trait, but that's actually a different problem we already had in the k8s client design.

DavSanchez:
> I cannot say that I like the huge trait, but that's actually a different problem we already had in the k8s client design.

A follow-up for tackling this might be identifying what operations each consumer of the trait uses, split the trait by the different kinds of operations that can be done (like `ListK8sResources`, `DeleteK8sResources`) and have each consumer parameterized only by the operations it uses, but needs more careful review.

PR #2481, opened 2026-04-30
feat: allow configmap only agents to change with remote configs

alvarocabanas:
allow configmap only agents to change with remote configs and add an integration test

### DIFF
--- a/agent-control/src/k8s/dynamic_object.rs
+++ b/agent-control/src/k8s/dynamic_object.rs
@@ -258,22 +258,25 @@ impl DynamicObjectManagers {
 /// object `from_manifest` can be different from the object `from_cluster` and still have equivalent
 /// content. Metadata is not considered.
 ///
+/// Supported object types:
+/// - `Secret`: compares `data` fields with secret-aware equality.
+/// - `ConfigMap`: compares `data` fields.
+/// - Any other object: compares `spec` fields.
+///
 /// # Arguments
 ///
 /// * `from_manifest` - DynamicObject representing manifest that can be applied to the cluster.
 /// * `from_cluster` - DynamicObject representing an object read from the cluster.
 fn object_contents_are_equal(from_manifest: &DynamicObject, from_cluster: &DynamicObject) -> bool {
-    // Compare Secrets data
-    if from_manifest
+    match from_manifest
         .types
         .as_ref()
-        .is_some_and(|t| t.api_version == "v1" && t.kind == "Secret")
+        .map(|t| (t.api_version.as_str(), t.kind.as_str()))
     {
-        return secrets_data_are_equal(from_manifest, from_cluster);
+        Some(("v1", "Secret")) => secrets_data_are_equal(from_manifest, from_cluster),
+        Some(("v1", "ConfigMap")) => configmaps_data_are_equal(from_manifest, from_cluster),
+        _ => from_manifest.data["spec"] == from_cluster.data["spec"],
     }
-    // Compare any other object
-    // TODO: consider support for particular objects only and validate object type meta when agent-types are defined.
-    from_manifest.data["spec"] == from_cluster.data["spec"]
 }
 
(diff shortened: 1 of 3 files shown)


agent-control/src/k8s/dynamic_object.rs
```diff
@@ -263,17 +263,16 @@ impl DynamicObjectManagers {
...
+    match from_manifest
         .types
         .as_ref()
-        .is_some_and(|t| t.api_version == "v1" && t.kind == "Secret")
+        .map(|t| (t.api_version.as_str(), t.kind.as_str()))
     {
-        return secrets_data_are_equal(from_manifest, from_cluster);
+        Some(("v1", "Secret")) => secrets_data_are_equal(from_manifest, from_cluster),
+        Some(("v1", "ConfigMap")) => configmap_data_are_equal(from_manifest, from_cluster),
+        // TODO: consider support for particular objects only and validate object type meta when agent-types are defined.
```
sigilioso: The comen twas already in place, but I don't think it is the proper place to consider it. We currently support:

- Secrets
- ConfigMaps
- Any other object whose content is defined as `spec`

I think we should remove this TODO and update the comment above to reflect the behavior.

agent-control/src/k8s/dynamic_object.rs
```diff
@@ -431,4 +438,37 @@ mod tests {
...
     }
+
+    #[rstest]
+    #[case::both_empty(json!({}), json!({}), json!({}), json!({}), true)]
+    #[case::matching_data(json!({"key": "value"}), json!({}), json!({"key": "value"}), json!({}), true)]
+    #[case::matching_binary_data(json!({}), json!({"key": "dmFsdWU="}), json!({}), json!({"key": "dmFsdWU="}), true)]
+    #[case::data_value_differs(json!({"key": "value"}), json!({}), json!({"key": "other"}), json!({}), false)]
+    #[case::data_key_missing_in_cluster(json!({"key": "value"}), json!({}), json!({}), json!({}), false)]
+    #[case::binary_data_differs(json!({}), json!({"key": "dmFsdWU="}), json!({}), json!({"key": "b3RoZXI="}), false)]
+    fn test_configmap_data_are_equal(
```
sigilioso, line 373: Nice! We could also update the secret's test to make them uniform, but it can also be out of scope for this PR.

sigilioso (approved):
🚀

agent-control/src/k8s/dynamic_object.rs
```diff
@@ -258,22 +258,25 @@ impl DynamicObjectManagers {
...
-    if from_manifest
+    match from_manifest
         .types
         .as_ref()
-        .is_some_and(|t| t.api_version == "v1" && t.kind == "Secret")
+        .map(|t| (t.api_version.as_str(), t.kind.as_str()))
     {
-        return secrets_data_are_equal(from_manifest, from_cluster);
+        Some(("v1", "Secret")) => secrets_data_are_equal(from_manifest, from_cluster),
+        Some(("v1", "ConfigMap")) => configmap_data_are_equal(from_manifest, from_cluster),
```
danielorihuela: Nit: Rename `configmap_data_are_equal` to `configmapS_data_are_equal`

danielorihuela (dismissed):

sigilioso (dismissed):
🚀

agent-control/src/k8s/dynamic_object.rs
alvarocabanas: Done in last commit

agent-control/src/k8s/dynamic_object.rs
alvarocabanas: Done in last commit

agent-control/src/k8s/dynamic_object.rs
alvarocabanas: Done in last commit

danielorihuela (approved):

sigilioso (approved):

PR #2486, opened 2026-05-05
feat(doc): oci auth parameters

sigilioso:
This PR includes docs for the `oci` configuration in Agent Control.


danielorihuela (approved):
Thanks!!!!

I noticed that `self_update` and `uptime_report` are not documented. Should they?

sigilioso:
🤔 `self_update` development is still on-going, I'm not sure about `uptime_report`. Let's open a different PR if makes sense.

PR #2487, opened 2026-05-05
chore(public_key): removing null references

paologallinaharbur:
We were told that the IAM team is about to remove null values from from payload returned.

Since we were ignoring already such values nothing should change, but we are updating mocks and internal tests to reflect the change


sigilioso (approved):
Thanks!

PR #2488, opened 2026-05-05
feat(docs): local registry mirrors and key rotation

sigilioso:
This PR extends the package-manager docs to explain the support of local registry mirrors and their requirements for signature validation (including the key-rotation scenario).

### DIFF
--- a/docs/package_manager.md
+++ b/docs/package_manager.md
@@ -242,7 +242,7 @@ Considering the [key rotation](#key-rotation) timeline:
 
 4. Old Key is Retired: After a safe transition period, the old Key A is removed from the public JWKS file.
 
-In order to ensure that Agent Control can always use packages from your local mirror, even after a public key rotation, your mirroring infrastructure must satisfy two fundamental requirements at the moment of verification.
+In order to ensure that Agent Control can always use packages from your mirror, even after a key rotation, your mirroring infrastructure must satisfy two fundamental requirements at the moment of verification.
 
 1. Availability of the public keys: the public key urls need to be reachable from the host where Agent Control is installed.
 
--- a/docs/package_manager.md
+++ b/docs/package_manager.md
@@ -220,7 +220,7 @@ What happens during a key rotation? It depends on the specific use case. AC trie
 
 ## Registry mirror support
 
-A  OCI registry mirror is a copy of the New Relic registry where agent packages are stored. Its support is essential for:
+An OCI registry mirror is a copy of the New Relic registry where agent packages are stored. Its support is essential for:
 
 * Air-gapped environments: where production servers have no access to the public internet (the default registry is not reachable).
 * Local development: to speed up testing without needing to fetch packages remotely.
--- a/docs/package_manager.md
+++ b/docs/package_manager.md
@@ -218,33 +218,33 @@ sequenceDiagram
 
 What happens during a key rotation? It depends on the specific use case. AC tries to verify every signature in the signature-manifest layers with every public key published for that package, this approach avoids downtimes on key rotation. Note that a non-revoked key must exist on the list, otherwise signature verification will fail.
 
-## Local registry mirrors
+## Registry mirror support
 
-A local OCI registry mirror is a local copy of the New Relic registry where agent packages are stored. Its support is essential for:
+A  OCI registry mirror is a copy of the New Relic registry where agent packages are stored. Its support is essential for:
 
-* Air-gapped environments: where production servers have no access to the public internet.
+* Air-gapped environments: where production servers have no access to the public internet (the default registry is not reachable).
 * Local development: to speed up testing without needing to fetch packages remotely.
 * Network efficiency: to reduce bandwidth usage in large-scale deployments.
 
-Agent Control supports local registry mirrors by setting them up in the corresponding [configuration section](./CONFIG.md#oci).
+Agent Control supports registry mirrors by setting them up in the corresponding [configuration section](./CONFIG.md#oci).
 
-### Local registry mirrors and key rotation
+### Registry mirrors and key rotation
 
 Considering the [key rotation](#key-rotation) timeline:
 
 1. New Key is Added: When New Relic introduces a new signing key (Key B), the public part of it is added to the JWKS file. The old key (Key A) is not immediately removed. The JWKS now contains both Key A and Key B.
 
-2. New Packages are Signed: All new agent packages are now signed with Key B.
+2. New signatures are created: All agent packages are signed with Key B resulting in new artifacts.
 
 3. Seamless Verification:
-   - When verifying an old package (signed with Key A), Agent Control tries Key B (fails) and then tries Key A (succeeds). Verification passes.
-   - When verifying a new package (signed with Key B), Agent Control tries Key B and it succeeds immediately.
+   - When verifying a package (signed with Key A only), Agent Control tries Key B (fails) and then tries Key A (succeeds). Verification passes.
+   - When verifying a new package (signed with Key A and Key B), Agent Control tries Key B and it succeeds immediately.
 
 4. Old Key is Retired: After a safe transition period, the old Key A is removed from the public JWKS file.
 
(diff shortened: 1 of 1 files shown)


docs/package_manager.md
```diff
@@ -218,6 +218,39 @@ sequenceDiagram
 
 What happens during a key rotation? It depends on the specific use case. AC tries to verify every signature in the signature-manifest layers with every public key published for that package, this approach avoids downtimes on key rotation. Note that a non-revoked key must exist on the list, otherwise signature verification will fail.
 
+## Local registry mirrors
+
+A local OCI registry mirror is a local copy of the New Relic registry where agent packages are stored. Its support is essential for:
+
+* Air-gapped environments: where production servers have no access to the public internet.
```
danielorihuela: I was thinking about that. Is AC prepared for that scenario? If it's air-gapped, it won't communicate with fleet, will it? In that case, it won't receive the deployments.

docs/package_manager.md
```diff
@@ -218,6 +218,39 @@ sequenceDiagram
 
 What happens during a key rotation? It depends on the specific use case. AC tries to verify every signature in the signature-manifest layers with every public key published for that package, this approach avoids downtimes on key rotation. Note that a non-revoked key must exist on the list, otherwise signature verification will fail.
 
+## Local registry mirrors
```
danielorihuela: Why do we talk about "local"? We also support "remote" mirrors. We don't care where the mirror is located as long as we have access to it.

docs/package_manager.md
```diff
@@ -218,6 +218,39 @@ sequenceDiagram
...
+
+Agent Control supports local registry mirrors by setting them up in the corresponding [configuration section](./CONFIG.md#oci).
+
+### Local registry mirrors and key rotation
+
+Considering the [key rotation](#key-rotation) timeline:
+
+1. New Key is Added: When New Relic introduces a new signing key (Key B), the public part of it is added to the JWKS file. The old key (Key A) is not immediately removed. The JWKS now contains both Key A and Key B.
+
+2. New Packages are Signed: All new agent packages are now signed with Key B.
```
danielorihuela:
```suggestion
2. New Signatures are created: All agent packages are signed with Key B, resulting in new oci artifacts.
```

docs/package_manager.md
```diff
@@ -218,6 +218,39 @@ sequenceDiagram
...
+
+3. Seamless Verification:
+   - When verifying an old package (signed with Key A), Agent Control tries Key B (fails) and then tries Key A (succeeds). Verification passes.
+   - When verifying a new package (signed with Key B), Agent Control tries Key B and it succeeds immediately.
```
danielorihuela:
```suggestion
   - When verifying a package (signed with Key A), Agent Control tries Key B (fails) and then tries Key A (succeeds). Verification passes.
   - When verifying a new package (signed with Key A and Key B), Agent Control tries Key B and it succeeds immediately.
```

docs/package_manager.md
```diff
@@ -218,6 +218,39 @@ sequenceDiagram
...
+
+4. Old Key is Retired: After a safe transition period, the old Key A is removed from the public JWKS file.
+
+In order to ensure that Agent Control can always use packages from your local mirror, even after a public key rotation, your mirroring infrastructure must satisfy two fundamental requirements at the moment of verification.
+
+1. Availability of the public keys: The set of public keys available to the Agent Control must be a superset of all keys required to verify every package you intend to serve from your mirror (the public key urls are reachable by Agent Control).
```
danielorihuela:
```suggestion
1. Availability of the public keys: The set of public keys available to Agent Control must be a superset of all keys required to verify every package you intend to serve from your mirror (the public key urls are reachable by Agent Control).
```

I don't understand this paragraph. Keys are only served by NewRelic. The path is hardcoded in the agent type. Either all keys are available or none is. How can the user make sure that AC has access to a superset of the keys?

docs/package_manager.md
danielorihuela: I think talking about "new packages" and "old packages" is confusing. They don't change. Only signature artifacts. We even mention them later

> 2. Your mirror must store and serve the original, unaltered agent package and its corresponding signature artifact as they were published by New Relic.

danielorihuela (commented):
Thanks for the documentation!!!

I just left a few nits and a couple questions.

docs/package_manager.md
sigilioso: It could still work with local config, but probably it is not the most common scenario. I can update it and mention restricted environment where New Relic's repository is not reachable. WDYT?

docs/package_manager.md
sigilioso: Lets ~Local~

docs/package_manager.md
danielorihuela: Nah, it's okay.

docs/package_manager.md
sigilioso: You are right, considering that the key url cannot be configured, we just need to state that it needs to be reachable.

docs/package_manager.md
sigilioso: [4ebf2f2](https://github.com/newrelic/newrelic-agent-control/pull/2488/commits/4ebf2f2096c55b6e8d16cf8db3ace82717063dd8)

docs/package_manager.md
sigilioso: You were totally right, changed: [4ebf2f2](https://github.com/newrelic/newrelic-agent-control/pull/2488/commits/4ebf2f2096c55b6e8d16cf8db3ace82717063dd8)

docs/package_manager.md
sigilioso: [4ebf2f2](https://github.com/newrelic/newrelic-agent-control/pull/2488/commits/4ebf2f2096c55b6e8d16cf8db3ace82717063dd8)

docs/package_manager.md
sigilioso: [4ebf2f2](https://github.com/newrelic/newrelic-agent-control/pull/2488/commits/4ebf2f2096c55b6e8d16cf8db3ace82717063dd8)

docs/package_manager.md
```diff
@@ -218,6 +218,39 @@ sequenceDiagram
 
 What happens during a key rotation? It depends on the specific use case. AC tries to verify every signature in the signature-manifest layers with every public key published for that package, this approach avoids downtimes on key rotation. Note that a non-revoked key must exist on the list, otherwise signature verification will fail.
 
+## Registry mirror support
+
+A  OCI registry mirror is a copy of the New Relic registry where agent packages are stored. Its support is essential for:
```
danielorihuela:
```suggestion
An OCI registry mirror is a copy of the New Relic registry where agent packages are stored. Its support is essential for:
```

sorry for the pickiness

sigilioso: I appreciate it!

docs/package_manager.md
```diff
@@ -218,6 +218,39 @@ sequenceDiagram
...
+
+2. New signatures are created: All agent packages are signed with Key B resulting in new artifacts.
+
+3. Seamless Verification:
+   - When verifying a package (signed with Key A only), Agent Control tries Key B (fails) and then tries Key A (succeeds). Verification passes.
+   - When verifying a new package (signed with Key A and Key B), Agent Control tries Key B and it succeeds immediately.
+
+4. Old Key is Retired: After a safe transition period, the old Key A is removed from the public JWKS file.
+
+In order to ensure that Agent Control can always use packages from your local mirror, even after a public key rotation, your mirroring infrastructure must satisfy two fundamental requirements at the moment of verification.
```
danielorihuela:
```suggestion
In order to ensure that Agent Control can always use packages from your mirror, even after a key rotation, your mirroring infrastructure must satisfy two fundamental requirements at the moment of verification.
```

sigilioso: You found the missing local!

danielorihuela (approved):

PR #2490, opened 2026-05-05
test: windows e2e self-update

gsanchezgavier:
Adds a windows self-update e2e test. The test basically installs latest AC from the recipe and self-update to the compiled version. It leverages a fake OpAMP server to send the remote config and spawn the same local registry (but with TLS so AC compiled in release mode can connect) as the integration test to store the compiled package.

Other:
- Our `component_packages.yml` was not overriding the AC version when using a specific tag, probably because we never need it to be correct. You could see the nightlies having the cargo toml version. As for the self update to work this needed to be changed so new != current version.

Pending:
- We need AC 1.13.0 or greater (that has self-update capabilities) in order to activate this test. For now there is a special commit that forces to install 1.13.0

### DIFF
--- a/test/crates/fake-opamp-server/src/lib.rs
+++ b/test/crates/fake-opamp-server/src/lib.rs
@@ -4,10 +4,11 @@ use aws_lc_rs::rand::SystemRandom;
 use aws_lc_rs::signature::{Ed25519KeyPair, KeyPair as _};
 use base64::Engine;
 use base64::prelude::{BASE64_STANDARD, BASE64_URL_SAFE_NO_PAD};
+use opamp_client::opamp::proto::any_value::Value;
 use opamp_client::opamp::proto::{
     AgentConfigFile, AgentConfigMap, AgentDescription, AgentRemoteConfig, AgentToServer,
-    ComponentHealth, CustomMessage, EffectiveConfig, RemoteConfigStatus, ServerToAgent,
-    ServerToAgentFlags,
+    ComponentHealth, CustomMessage, EffectiveConfig, RemoteConfigStatus, RemoteConfigStatuses,
+    ServerToAgent, ServerToAgentFlags,
 };
 use opamp_client::operation::instance_uid::InstanceUid;
 use prost::Message;
@@ -249,6 +250,38 @@ impl FakeServer {
             .get(&identifier.into())
             .map(|s| s.config_status.clone())
     }
+
+    /// Returns the instance IDs of all connected agents that have an identifying attribute
+    /// matching the given key–value pair. Matches string-typed attribute values only.
+    pub fn find_agents_with_identifying_attr(&self, key: &str, value: &str) -> Vec<InstanceID> {
+        let state = self.state.lock().unwrap();
+        state
+            .agent_state
+            .iter()
+            .filter(|(_, agent_state)| has_identifying_attr(&agent_state.attributes, key, value))
+            .map(|(id, _)| id.clone())
+            .collect()
+    }
+
+    /// Returns the string value of an identifying attribute for the given agent, or `None` if the
+    /// agent is unknown or the attribute is absent or not a string.
+    pub fn get_identifying_attr_value(
+        &self,
+        identifier: impl Into<InstanceUid>,
+        key: &str,
+    ) -> Option<String> {
+        let state = self.state.lock().unwrap();
+        state
+            .agent_state
+            .get(&identifier.into())
+            .and_then(|s| find_string_identifying_attr(&s.attributes, key))
+    }
+
+    /// Returns `true` if the agent has reported a remote config status of `Applied`.
+    pub fn is_config_status_applied(&self, identifier: impl Into<InstanceUid>) -> bool {
+        self.get_remote_config_status(identifier)
+            .is_some_and(|s| s.status == RemoteConfigStatuses::Applied as i32)
+    }
 }
 
 impl Drop for FakeServer {
@@ -349,6 +382,30 @@ fn build_response(
         ..Default::default()
     }
     .encode_to_vec()
+}
+
+fn find_string_identifying_attr(description: &AgentDescription, key: &str) -> Option<String> {
+    description
+        .identifying_attributes
+        .iter()
+        .find(|kv| kv.key == key)
+        .and_then(|kv| {
+            kv.value.as_ref()?.value.as_ref().and_then(|v| match v {
+                Value::StringValue(s) => Some(s.clone()),
+                _ => None,
+            })
+        })
+}
+
+fn has_identifying_attr(description: &AgentDescription, key: &str, value: &str) -> bool {
+    description.identifying_attributes.iter().any(|kv| {
+        kv.key == key
+            && kv
+                .value
+                .as_ref()
+                .and_then(|av| av.value.as_ref())
+                .is_some_and(|v| matches!(v, Value::StringValue(s) if s == value))
+    })
 }
 
 fn generate_key_pair() -> Ed25519KeyPair {
--- a/test/crates/fake-opamp-server/src/lib.rs
+++ b/test/crates/fake-opamp-server/src/lib.rs
@@ -277,10 +277,16 @@ impl FakeServer {
             .and_then(|s| find_string_identifying_attr(&s.attributes, key))
     }
 
-    /// Returns `true` if the agent has reported a remote config status of `Applied`.
-    pub fn is_config_status_applied(&self, identifier: impl Into<InstanceUid>) -> bool {
-        self.get_remote_config_status(identifier)
-            .is_some_and(|s| s.status == RemoteConfigStatuses::Applied as i32)
+    /// Returns `Ok(())` if the agent has reported a remote config status of `Applied`.
+    pub fn is_config_status_applied(
+        &self,
+        identifier: impl Into<InstanceUid>,
+    ) -> Result<(), String> {
+        match self.get_remote_config_status(identifier) {
+            Some(s) if s.status == RemoteConfigStatuses::Applied as i32 => Ok(()),
+            Some(_) => Err("Config status is not Applied".to_string()),
+            None => Err("Remote config status not found".to_string()),
+        }
     }
 }
 
@@ -385,27 +391,19 @@ fn build_response(
 }
 
 fn find_string_identifying_attr(description: &AgentDescription, key: &str) -> Option<String> {
-    description
+    let kv = description
         .identifying_attributes
         .iter()
-        .find(|kv| kv.key == key)
-        .and_then(|kv| {
-            kv.value.as_ref()?.value.as_ref().and_then(|v| match v {
-                Value::StringValue(s) => Some(s.clone()),
-                _ => None,
-            })
-        })
+        .find(|kv| kv.key == key)?;
+
+    match kv.value.as_ref()?.value.as_ref()? {
+        Value::StringValue(s) => Some(s.clone()),
+        _ => None,
+    }
 }
 
 fn has_identifying_attr(description: &AgentDescription, key: &str, value: &str) -> bool {
-    description.identifying_attributes.iter().any(|kv| {
-        kv.key == key
-            && kv
-                .value
-                .as_ref()
-                .and_then(|av| av.value.as_ref())
-                .is_some_and(|v| matches!(v, Value::StringValue(s) if s == value))
-    })
+    find_string_identifying_attr(description, key).is_some_and(|v| v == value)
 }
 
 fn generate_key_pair() -> Ed25519KeyPair {
--- a/test/e2e-runner/src/common/oci.rs
+++ b/test/e2e-runner/src/common/oci.rs
@@ -1,11 +1,11 @@
 use crate::common::InstallationArgs;
 use crate::common::cert::SelfSignedCert;
 use crate::common::runtime::tokio_runtime;
+use crate::common::test::retry;
 use oci_client::Reference;
 use oci_test_utils::{OCISigner, PackageMediaType, PackagePublisher};
 use std::path::{Path, PathBuf};
 use std::process::{Child, Command, Stdio};
-use std::thread;
 use std::time::Duration;
 use tracing::info;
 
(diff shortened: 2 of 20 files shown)


test/crates/fake-opamp-server/src/lib.rs
```diff
@@ -351,6 +384,30 @@ fn build_response(
...
+
+fn has_identifying_attr(description: &AgentDescription, key: &str, value: &str) -> bool {
+    description.identifying_attributes.iter().any(|kv| {
+        kv.key == key
+            && kv
+                .value
+                .as_ref()
+                .and_then(|av| av.value.as_ref())
+                .is_some_and(|v| matches!(v, Value::StringValue(s) if s == value))
+    })
```
sigilioso: Nit: we could reuse the other function, couldn't we?

```suggestion
        find_string_identifying_attr(description, key).is_some_and(|v| v == value)
```

test/e2e-runner/src/common/oci.rs
```diff
@@ -0,0 +1,178 @@
...
+    let bash = PathBuf::from("bash");
+
+    let mut cmd = Command::new(bash);
+    cmd.arg(script_path);
+    cmd
+}
+
+fn wait_for_registry() {
+    let url = format!("https://localhost:{LOCAL_REGISTRY_PORT}/v2/");
+    for _ in 0..30 {
```
sigilioso: Nit: could we use the `retry` helper here?

test/crates/fake-opamp-server/src/lib.rs
```diff
@@ -351,6 +384,30 @@ fn build_response(
     .encode_to_vec()
 }
 
+fn find_string_identifying_attr(description: &AgentDescription, key: &str) -> Option<String> {
+    description
+        .identifying_attributes
+        .iter()
+        .find(|kv| kv.key == key)
+        .and_then(|kv| {
+            kv.value.as_ref()?.value.as_ref().and_then(|v| match v {
```
danielorihuela: Nit


```suggestion
            match kv.value.as_ref()?.value.as_ref()? {
```

test/crates/fake-opamp-server/src/lib.rs
```diff
@@ -249,6 +250,38 @@ impl FakeServer {
...
+            .agent_state
+            .get(&identifier.into())
+            .and_then(|s| find_string_identifying_attr(&s.attributes, key))
+    }
+
+    /// Returns `true` if the agent has reported a remote config status of `Applied`.
+    pub fn is_config_status_applied(&self, identifier: impl Into<InstanceUid>) -> bool {
+        self.get_remote_config_status(identifier)
+            .is_some_and(|s| s.status == RemoteConfigStatuses::Applied as i32)
+    }
```
sigilioso: Considering the usage, would it make sense to return `Result<(), String>`?

test/e2e-runner/src/windows/scenarios/self_update.rs
```diff
@@ -0,0 +1,142 @@
...
+        "waiting for remote config Applied status",
+        || {
+            if opamp_server.is_config_status_applied(instance_id.clone()) {
+                Ok(())
+            } else {
+                Err("remote config status not yet Applied".into())
+            }
+        },
+    );
+
```
sigilioso, line 108: It doesn't add much value to check the message sent by the 'Verify command' to the OpAMP server, right?

test/e2e-runner/src/windows/scenarios/self_update.rs
```diff
@@ -0,0 +1,142 @@
...
+            if version == new_version {
+                Ok(version)
+            } else {
+                Err(format!("expected version {new_version}, got {version}").into())
+            }
+        },
+    );
+    info!(version = new_version, "AC version updated successfully");
+
+    info!("Self-update test completed successfully");
```
sigilioso, line 129: Would checking the binary version (`<ac-path>\\newrelic-agent-control.exe version`) as well be too much?

test/e2e-runner/src/windows/scenarios/self_update.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    install_latest_agent_control(&RecipeData {
+        args: args.clone(),
+        ..Default::default()
+    });
+
+    let self_update_config = format!(
+        r#"
+fleet_control:
+  endpoint: {}
```
danielorihuela, line 36: Should we use `poll_interval: 5s` like in other tests? I don't know how long that test took and if that would help a bit reduce the time.

tools/oci-registry.sh
```diff
@@ -123,6 +123,18 @@ EOF
...
+    if [ -n "${TLS_CERT:-}" ] && [ -n "${TLS_KEY:-}" ]; then
+        http_tls_block=$(cat << EOF
+,
+        "tls": {
+            "cert": "$TLS_CERT",
+            "key": "$TLS_KEY"
+        }
+EOF
+)
+    fi
```
sigilioso, line 137: We can use the tests themselves as documentation, but I wonder if we should add some comments regarding the use of `$TLS_CERT and `$TLS_KEY`.

test/e2e-runner/src/windows/scenarios/self_update.rs
```diff
@@ -0,0 +1,142 @@
...
+        "verifying updated agent.version attribute",
+        || {
+            let Some(version) =
+                opamp_server.get_identifying_attr_value(instance_id.clone(), AGENT_VERSION_ATTR)
+            else {
+                return Err("agent.version attribute not set yet".into());
+            };
+            if version == new_version {
+                Ok(version)
+            } else {
```
danielorihuela:
```suggestion
            if version != new_version {
```

Why do we need the `Ok(version)`?

sigilioso (commented):
Thanks for addressing this! (including the tls support for the testing repository).

Everything looks good, most comments are _nits_. I have just one concerning the versions.

We are installing latest (which will be `1.13.0`) and then upgrading to the version we have just compiled, right? If we introduced a but that makes the upgrade fail (Eg: we cannot download the AC package anymore) we wouldn't catch it until the version is released.

If we swapped the versions (install the compiled one and _downgrade_ to the latest published) we would caught such issues, wouldn't we?

Ideally, we could compile twice (or tweak the package somehow) and use the compiled for both, but I don't think we are ready for that yet (maybe after improving the installation process).

test/e2e-runner/src/common/oci.rs
```diff
@@ -0,0 +1,178 @@
...
+
+pub struct OciRegistry {
+    run_process: Child,
+    _cert: SelfSignedCert,
+}
+
+impl OciRegistry {
+    pub fn start() -> Self {
+        // AC can only connect to https registries so we generate certificates to enable TLS
+        let cert = SelfSignedCert::generate();
```
danielorihuela, line 24: How is that working?
When I manually tested it, I couldn't use self-signed certificates on linux. The oci client complained with not finding the manifest.

test/e2e-runner/src/windows/install.rs
```diff
@@ -105,11 +96,58 @@ $env:NEW_RELIC_AGENT_CONTROL_SKIP_BINARY_SIGNATURE_VALIDATION='true'; `
...
 
-    // Create a temporary .ps1 file for the installation command
-    //
-    // There's an option that allows running commands directly. That is "-Command". The
-    // issue with that is that "-ExecutionPolicy" won't bypass all the checks. It seems
-    // to only work properly using a ps1 script. We are forced to create a temporary script file.
+    run_newrelic_install(install_command);
+}
+
+pub fn install_latest_agent_control(data: &RecipeData) {
```
danielorihuela, line 102: Will this disappear once `1.13` is released?

test/e2e-runner/src/windows/scenarios/self_update.rs
danielorihuela: +1

test/e2e-runner/src/windows/install.rs
danielorihuela: Ah no. We need it because of what you said on the description.

> The test basically installs latest AC from the recipe and self-update to the compiled version.

test/e2e-runner/src/windows/install.rs
```diff
@@ -105,11 +96,58 @@ $env:NEW_RELIC_AGENT_CONTROL_SKIP_BINARY_SIGNATURE_VALIDATION='true'; `
...
+$env:NEW_RELIC_AUTH_PROVISIONED_CLIENT_ID='{}'; `
+$env:NEW_RELIC_AUTH_PRIVATE_KEY_PATH='{}'; `
+$env:NEW_RELIC_REGION='{}'; `
+$env:NR_CLI_FLEET_ID='{}'; `
+$env:NEW_RELIC_AGENT_CONTROL_FLEET_ENABLED='{}'; `
+$env:NEW_RELIC_AGENT_CONTROL='true'; `
+$env:NEW_RELIC_AGENT_VERSION='1.13.0'; `
+& "C:\Program Files\New Relic\New Relic CLI\newrelic.exe" install `
+-n {}
+"#,
```
danielorihuela, line 119: This is duplicated in `install_agent_control_from_recipe`. Maybe we can have a separated function to reuse all the values and accept an input `override_version`. Or reuse one of the functions you already created. I don't know.

gsanchezgavier:
> We are installing latest (which will be `1.13.0`) and then upgrading to the version we have just compiled, right? If we introduced a but that makes the upgrade fail (Eg: we cannot download the AC package anymore) we wouldn't catch it until the version is released.
> 
> If we swapped the versions (install the compiled one and _downgrade_ to the latest published) we would caught such issues, wouldn't we?
> 

@sigilioso yes exactly , we need another test to cover the self-update from compiled version to another (which need to have the verify command in place)

> Ideally, we could compile twice (or tweak the package somehow) and use the compiled for both, but I don't think we are ready for that yet (maybe after improving the installation process).

I though about this but it takes sooooo much time to compile it that i think that waiting until we have the first packaged OCI will be much easier and  the test will cover that the current bin can execute the self update. It also will be much simpler as no local registry will be required.

test/crates/fake-opamp-server/src/lib.rs
gsanchezgavier: i can remove the previous and_then too so is consistently half functional 😆

test/e2e-runner/src/common/oci.rs
gsanchezgavier: mm haven't tested on linux, @vjripoll  was working on it. In this case is being added to the Root trust store.

test/e2e-runner/src/windows/scenarios/self_update.rs
gsanchezgavier: it could help , but given that we have a 20m compilation wf in a row i would keep it the standard as it is an e2e

test/e2e-runner/src/windows/scenarios/self_update.rs
gsanchezgavier: we are not sending any message on success are we?

test/e2e-runner/src/windows/scenarios/self_update.rs
gsanchezgavier: we don't! 👍

test/e2e-runner/src/windows/scenarios/self_update.rs
gsanchezgavier: this is exactly what the version checker is doing and attaching that into the opamp message that we check, so i would avoid that.

test/e2e-runner/src/windows/install.rs
gsanchezgavier: i did some refactor to re use all the code, just lef this outside on porpouse just to have something readable , is quite different if you see it in detail

sigilioso:
> I though about this but it takes sooooo much time to compile it that i think that waiting until we have the first packaged OCI will be much easier and the test will cover that the current bin can execute the self update. It also will be much simpler as no local registry will be required.

I was thinking about having some way to tweak the version after compiling (something _hacky_ such as taking the version from an environment variable under a particular compilation feature), but maybe it doesn't worth the effort....

test/e2e-runner/src/windows/scenarios/self_update.rs
sigilioso: I meant the message that the 'verify command' sends containing a particular non-identifying attribute, probably not needed to be checked.

test/e2e-runner/src/common/oci.rs
danielorihuela: Strange. I might have done something wrong. I remember using a self-signed certificate and adding it to the root store in linux and having problems. 🤷 

Well, if it works it works.

test/e2e-runner/src/windows/install.rs
danielorihuela: Ooh true. There's the proxy and signature validation options there too. I said nothing. Thanks!!!

test/e2e-runner/src/windows/scenarios/self_update.rs
danielorihuela: Oh, I have a different doubt then. Do we need to verify that the config status is applied at all? In the next step we check whether the `agent.version` was updated. That should be enough from my PoV. I don't see the advantage. We have different possible executions:

1. Check config is applied before pulling new config

If AC is waiting for the next pull, we are still waiting for the self-update. The previous config is applied. We know that because at line 72 we got the `initial_version`

2. Check config is applied after pulling new config

If AC already pulled the new config, then we are on the self-update path. We must wait to see if the version is updated. Am I missing something?

---

If my mental model is correct, to make sure that the config is applied after the pull, we should check it's applied at the same time that we check the version is updated or after it. I think we could remove it.

gsanchezgavier: > Do we need to verify that the config status is applied at all?

no, not required, is just that i think is important part of the flow and does not add complexity to the test.

tools/oci-registry.sh
gsanchezgavier: added

test/e2e-runner/src/windows/scenarios/self_update.rs
danielorihuela: Okey, and should we rework it? How do you know that the `Applied` status makes reference to the new config and not the one that was before? Because the pulling is done every 30 seconds.

sigilioso (approved):
LGTM

PR #2492, opened 2026-05-06
refactor: replace deprecated `serde_yaml` with `serde_saphyr`

DavSanchez:
In preparation for the [incoming `kube-rs` v4](https://github.com/kube-rs/kube/milestone/52), we can start working to remove our dependency on `serde_yaml`!

## Summary

This PR replaces the deprecated `serde_yaml` dependency with `serde_saphyr` across the entire workspace, while migrating abstract YAML DOM manipulations from `serde_yaml::Value`/`Mapping`/`Number` to `serde_json::Value`/`Map`/`Number`.

## Motivation

The `serde_yaml` crate is officially deprecated and no longer maintained. `serde_saphyr` is the recommended modern replacement that provides YAML 1.2 serialization/deserialization via `serde`.

However, `serde_saphyr` intentionally does **not** provide `Value`/`Mapping`/`Number` types. All abstract YAML tree manipulations must be expressed through `serde_json::Value`/`Map`/`Number` instead.

### Dependencies

- **Removed** `serde_yaml` from workspace `Cargo.toml`, `agent-control/Cargo.toml`, and `test/e2e-runner/Cargo.toml`
- **Added** `serde-saphyr` as the replacement for serde operations (`from_str`, `to_string`, `from_reader`, etc.)
- **Relied on existing** `serde_json` for abstract DOM types (`Value`, `Map<String, Value>`, `Number`)

### Source Migration (~66 files)

| Before (`serde_yaml`) | After (`serde_saphyr` / `serde_json`) |
|---|---|
| `serde_yaml::Value` | `serde_json::Value` |
| `serde_yaml::Mapping` | `serde_json::Map<String, serde_json::Value>` |
| `serde_yaml::Number` | `serde_json::Number` |
| `serde_yaml::from_str` | `serde_saphyr::from_str` |
| `serde_yaml::to_string` | `serde_saphyr::to_string` |
| `serde_yaml::from_reader` | `serde_saphyr::from_reader` |
| `serde_yaml::to_value` | `serde_json::to_value` |
| `serde_yaml::from_value` | `serde_json::from_value` |

### Behavioral Differences Handled

`serde_saphyr` has stricter defaults than `serde_yaml`. The following adjustments were made to preserve existing behavior:

1. **Duplicate YAML keys**: `serde_saphyr` errors by default on duplicate mapping keys, while `serde_yaml` silently accepted them. Production code paths (`FileStore::get`, `YAMLConfig::try_from`) now use `serde_saphyr::from_str_with_options` with `DuplicateKeyPolicy::LastWins` to match the previous behavior.

2. **Null → String deserialization**: `serde_yaml` deserialized `name:` (null value) into an empty `String`. `serde_saphyr` rejects this. The `AgentTypeID` deserialization was updated to use `Option<String>` fields with `.unwrap_or_default()` to handle this gracefully.

3. **Error message formats**: Several unit tests that asserted on exact `serde_yaml` error strings were updated to assert on the equivalent `serde_saphyr` error format.

4. **Block scalar serialization**: `serde_saphyr::to_string` emits multiline strings as quoted strings with `\n` escapes rather than YAML literal block scalars (`|`). One integration test's expected output was updated accordingly.

### Error Enum Refactoring

Error variants were renamed to generic names not coupled to the old dependency:

- `serde_yaml::Error` variants → `serde_json::Error` and `serde_saphyr::ser::Error` variants with names like `Serialization` / `Deserialization`
- Affected files: `agent_type/error.rs`, `k8s/error.rs`, `effective_agents_assembler.rs`, `config_migrate/migration/converter.rs`, and others

### Test Fixes

- **7 lib tests** updated to match `serde_saphyr` error message formats
- **1 integration test** (`test_store_remote_no_mocks`) updated expected YAML serialization format
- **Turbofish syntax** fixed in `custom_agent_type.rs` (`Map<String, Value>::new` → `Map::<String, Value>::new`)

## Verification

- [x] `cargo check --workspace` passes
- [x] `cargo test -p newrelic_agent_control --lib` — **863 tests pass**
- [x] `cargo test -p newrelic_agent_control --test integration_tests` — **54 tests pass** (76 ignored)
- [x] `cargo test --workspace --lib` — all workspace lib tests pass
- [x] `cargo clippy -p newrelic_agent_control` — clean (1 pre-existing warning fixed)
- [x] No remaining `serde_yaml` references in source code or tests

## Backward Compatibility

This change is intended to be **behaviorally transparent** to end users. All serialization/deserialization round-trips produce the same semantic results. The only user-visible difference is that YAML files with duplicate keys (which were previously silently accepted) are now also accepted via the `LastWins` policy.

## Related

- `serde_yaml` deprecation notice: https://github.com/dtolnay/serde_yaml/issues/388
- `serde-saphyr` crate: https://crates.io/crates/serde-saphyr
- `kube-rs` crate switch: https://github.com/kube-rs/kube/pull/1975


agent-control/src/agent_control/config.rs
```diff
@@ -197,10 +197,14 @@ const DEFAULT_SIGNATURE_VERIFICATION_ENABLED: bool = true;
 pub struct SignatureVerificationEnabled(bool);
 
 impl TryFrom<YAMLConfig> for AgentControlConfig {
-    type Error = serde_yaml::Error;
+    type Error = AgentControlConfigError;
 
     fn try_from(value: YAMLConfig) -> Result<Self, Self::Error> {
-        serde_yaml::from_value(serde_yaml::to_value(value)?)
+        serde_json::from_value(
```
danielorihuela: Why not use `serde_saphyr::from_str` and `serde_saphyr::to_str` like in other parts of this PR?

agent-control/src/agent_control/config.rs
```diff
@@ -297,9 +301,9 @@ impl TryFrom<&OpampRemoteConfig> for AgentControlDynamicConfig {
 /// Tries to append agents from a YAML value into the agents map, erroring on duplicates.
 fn try_append_agents(
     merged_agents: SubAgentsMap,
-    agents_value: serde_yaml::Value,
+    agents_value: serde_json::Value,
```
danielorihuela, line 303: Not for this PR.

I wonder if we should use `String` instead of `Value`. That way we are "decoupled" from whatever library we use (at the signature level). It's probably a lot more inconvenient to be converting stuff back and forth all the time.

agent-control/src/agent_control/config.rs
```diff
@@ -323,8 +327,8 @@ impl TryFrom<YAMLConfig> for AgentControlDynamicConfig {
     type Error = AgentControlConfigError;
 
     fn try_from(value: YAMLConfig) -> Result<Self, Self::Error> {
-        serde_yaml::from_value(
-            serde_yaml::to_value(value)
+        serde_json::from_value(
```
danielorihuela: Same as above.

> Why not use serde_saphyr::from_str and serde_saphyr::to_str like in other parts of this PR?

agent-control/tests/on_host/cli.rs
```diff
@@ -78,8 +78,8 @@ logs:
     create_temp_file(dir.path(), AUTH_PRIVATE_KEY_FILE_NAME, "dummy-key-content")?;
 
     let mut cmd = cmd_with_config_file(dir.path());
-    // cmd_assert is not made for long running programs, so we kill it anyway after 1 second
-    cmd.timeout(Duration::from_secs(1));
+    // cmd_assert is not made for long running programs, so we kill it anyway after 10 second
```
danielorihuela: Where they flaky?

agent-control/tests/on_host/config_repository.rs
```diff
@@ -80,10 +80,7 @@ const AGENT_VALUES_SINGLE_FILE: &str = r#"config_file: |
 "#;
 
 const AGENT_VALUES_SINGLE_FILE_STORED: &str = r#"config:
-  config_file: |
-    license_key: 1234567890987654321
-    log:
-      level: debug
+  config_file: "license_key: 1234567890987654321\nlog:\n  level: debug\n"
```
danielorihuela, line 83: > Block scalar serialization: serde_saphyr::to_string emits multiline strings as quoted strings with \n escapes rather than YAML literal block scalars (|). One integration test's expected output was updated accordingly.

The input of the user can still have a yaml inside yamls, right?

agent-control/tests/on_host/cli.rs
sigilioso: I actually had the same question

> 3 CLI tests (basic_startup, custom_logging_format, runs_with_no_config) had timeouts increased from 1s to 10s to prevent premature process interruption

Is this connected to the `serde_saphyr` changes somehow?

agent-control/src/agent_control/config.rs
DavSanchez: The change was mostly a S&R, I'll try this.

agent-control/tests/on_host/cli.rs
DavSanchez: they were flaky on my mac (and annoying enough) but certainly unrelated to the PR. Will remove and add it in another PR.

agent-control/tests/on_host/config_repository.rs
DavSanchez: Yes yes, but this modification involved the string value only. Not sure if the new library can be configured so the strings are printed as before (multiline).

agent-control/tests/on_host/cli.rs
DavSanchez: Moved it here https://github.com/newrelic/newrelic-agent-control/pull/2502

agent-control/src/agent_control/config.rs
DavSanchez: Looks like it works fine!

agent-control/src/agent_control/config.rs
DavSanchez: Looks like it works fine!

agent-control/src/agent_control/config.rs
DavSanchez: As this was S&R I didn't stop too much on changing types beyond the library switch. I can take a look or we can create a task for having a quick review.

danielorihuela:
From my side we could merge it. Should we mention that on a Tech Talk just in case? I mean before merging.

DavSanchez:
> From my side we could merge it. Should we mention that on a Tech Talk just in case? I mean before merging.

Added item for next tech meeting

sigilioso:
Maybe we should also remove it from the renovate rule: https://github.com/newrelic/newrelic-agent-control/blob/0837997fdb5c04564ddc74625e24e005210fb299/.github/renovate-shared-base.json#L94-L99

DavSanchez:
> Maybe we should also remove it from the renovate rule:
> 
> https://github.com/newrelic/newrelic-agent-control/blob/0837997fdb5c04564ddc74625e24e005210fb299/.github/renovate-shared-base.json#L94-L99

Good catch! https://github.com/newrelic/newrelic-agent-control/pull/2492/commits/54c8240234299101ec51d3511b07bce1e1a0c8d6

sigilioso (approved):
🚀

PR #2494, opened 2026-05-06
feat: upload agent-control to oci on publish

jcoscollanr:
Uploads the release binaries to the OCI repository after each release is done

### DIFF
--- a/.github/workflows/post-release-agent-metadata.yml
+++ b/.github/workflows/post-release-agent-metadata.yml
@@ -3,6 +3,12 @@ name: Post-Release Agent Metadata Upload
 on:
   release:
     types: [released]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Release tag to upload (e.g. 1.13.0)'
+        required: true
+        type: string
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/post-release-agent-metadata.yml
```diff
@@ -3,6 +3,12 @@ name: Post-Release Agent Metadata Upload
 on:
   release:
     types: [released]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Release tag to upload (e.g. 1.13.0)'
+        required: true
+        type: string
```
gsanchezgavier, line 11: would this override the OCI packages in case is triggered for a tag that has already been released? I wonder if we really need this , as we shouldn't need to backfill older versions of these packages (they do not support self-update)

.github/workflows/post-release-agent-metadata.yml
```diff
@@ -16,15 +22,85 @@ jobs:
...
+          echo "Downloading agent-control release artifacts for version ${VERSION}"
+          gh release download "${VERSION}" \
+            --repo newrelic/newrelic-agent-control \
+            --pattern "newrelic-agent-control_${VERSION}_linux_amd64.tar.gz" \
+            --pattern "newrelic-agent-control_${VERSION}_linux_arm64.tar.gz" \
+            --pattern "newrelic-agent-control_${VERSION}_windows_amd64.zip"
+          echo "Downloaded artifacts:"
+          ls -lh newrelic-agent-control_${VERSION}_*
+
+      - name: Repackage archives
```
gsanchezgavier, line 41: as a follow up i think we should change our release generation to build this packages and avoid this step.

gsanchezgavier (approved):
thanks! i just have some minor concern about the dispatch triggering

.github/workflows/post-release-agent-metadata.yml
jcoscollanr: Yes, it will reupload. it's mostly for the case that the pipeline fails. To not have to do hacks. The action recovers the commit of the tag, so it will upload exactly the same version

.github/workflows/post-release-agent-metadata.yml
jcoscollanr: That was done this way on purpose, so the release don't fail if for some reason the metadata action fails.

now with the self-update I imagine that is part of the release.

I still like to have it separated but I am open to ideas ^^

gsanchezgavier: i mean this https://github.com/newrelic/newrelic-agent-control/pull/2500

PR #2497, opened 2026-05-06
chore(debug): zigbuild memory leak

paologallinaharbur:
This major release of Zigbuild seems to be affected by a memory leak https://ziglang.org/news/0.16.0-released/

https://onenr.io/08jqqd7dbjl


.github/workflows/component_image.yml
gsanchezgavier, line 61: would make sense to extract this this to an actions `.github/actions/zig`

gsanchezgavier (approved):

.github/workflows/component_image.yml
paologallinaharbur: I am going to do it in a following PR together with adding some comments, I'd like to re-trigger the nightly now

PR #2501, opened 2026-05-08
fix: onhost canaries

gsanchezgavier:
issue 1
We noticed that onhost canaries were not were being created consistently. 
- This nightly worked https://github.com/newrelic/newrelic-agent-control/actions/runs/25148457592/job/73715188619
- This not https://github.com/newrelic/newrelic-agent-control/actions/runs/25485823336/job/74785949504
And bot where 🟢 
The case it was not , the ansible inventory was empty and that ended in no ansible executed and exit 0.

This pr adds a check for that in the makefile 

issue 2
Digging into that ansible inventory we noticed that the infra apply we were running on push to main was somtimes failing (not sure why because there were no logs) and no one noticed. It also has the issue that the onhost canary machine was also being recreated but not provisioned breaking the canary.

This pr moves the infra apply just before the provision , with the goal of surface any error on workflows that we are notified for and the hero take cares, and also fix the onhost broken canary in case of a machine recration. 

Extra:
Create a consistent workflows dispatch for the canaries so in case we want to manually recreate them we can.


test/onhost-canaries/Makefile
```diff
@@ -115,6 +115,16 @@ test/onhost-canaries/ansible-install:
...
+	@INVENTORY_FILE=$(ONHOST_TF_DIR)/environments/$(ENVIRONMENT)/inventory.$(ENVIRONMENT) && \
+	echo "Validating Ansible inventory contains the 2 canaries..." && \
+	HOST_COUNT=$$(ansible-inventory -i $$INVENTORY_FILE --list 2>/dev/null | jq '[.. | .hosts? // empty] | flatten | length') && \
+	if [ "$$HOST_COUNT" -ne 2 ]; then \
+		echo "ERROR: Inventory file has no hosts defined"; \
+		echo "Inventory content:"; \
+		cat $$INVENTORY_FILE; \
+		exit 1; \
+	fi && \
+	echo "✓ Inventory validation passed ($$HOST_COUNT hosts found)"
```
sigilioso, line 127: 💯  Nice!

At first as was worried of coupling the number of host in this check, but actually the whole Makefile is coupled to the hosts definition, so I'd say there is nothing to worry about.

.github/workflows/on_demand_k8s_canaries.yml
sigilioso, line 1: Nit: could we add some sort of comment to these workflow files? Just to specify that they apply both infra + AC update (same as the nightly does). I fear we can get confused because `component_xxx_canaries.yml` upgrades AC while `component_xxx_canaries_infra.yaml` upgrades Infra whereas these "on_demand" workflows do both.

sigilioso (approved):
LGTM

I liked the idea of applying the infra changes right after changing the corresponding code, but unnoticed failures are actually dangerous.

Another possibility would have been to add slack messages on the push_pr_xxx_canaries_apply.yml workflows, did you consider that?

test/onhost-canaries/Makefile
gsanchezgavier: yes i thought the same thing

gsanchezgavier:
> I liked the idea of applying the infra changes right after changing the corresponding code, but unnoticed failures are actually dangerous.
> Another possibility would have been to add slack messages on the push_pr_xxx_canaries_apply.yml workflows, did you consider that?

Not only failures but for onhost this means recreating it but not provisioning it, so there is no more AC running.

.github/workflows/on_demand_k8s_canaries.yml
gsanchezgavier: oh this is exactly why i removed the workflow dispatch of the component_ canaries. they are implementation detail. Before you needed to understand which execute whenever you want something to be updated , now it updates everything

alvarocabanas (approved):
LGTM

PR #2503, opened 2026-05-08
chore(release): 1.14.0

paologallinaharbur:
just bumping the release version

### DIFF
--- a/agent-control/Cargo.toml
+++ b/agent-control/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "newrelic_agent_control"
 description = "New Relic Agent Control"
-version = "1.13.0"
+version = "1.14.0"
 authors.workspace = true
 edition.workspace = true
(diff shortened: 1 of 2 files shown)


agent-control/Cargo.toml
```diff
@@ -45,12 +45,12 @@ self-replacer = { path = "../self-replacer" }
...
-    "ws",
-    "rustls-tls",
-    "aws-lc-rs",
+  "config",
+  "runtime",
+  "derive",
+  "ws",
+  "rustls-tls",
+  "aws-lc-rs",
 ] }
```
DavSanchez, line 54: We need to pin our formatter dependency to use it on local as well xDDD

DavSanchez (approved):

agent-control/Cargo.toml
paologallinaharbur: yes 😅

PR #2508, opened 2026-05-11
feat: alert for consistent memory growth over time

danielorihuela:
Adds new alert to detect slow memory leaks.

### Context

We detected a memory leak in k8s canaries. It was already present on onhost. However, the memory leak was so slow that the threshold wasn't reached before the new nightly version was deployed to the canaries machines.

This new alert is designed to detect memory leaks in a 24 hours time frame.

### Technical details

We use the [derivative](https://docs.newrelic.com/docs/nrql/nrql-syntax-clauses-functions/#derivative) to compute the rate of change within a 3 hour windows. The bigger the window, the better that we avoid triggering the alert because of an spike. We use a 1 hour duration for the alert. The advantage of that combination is that we could get the first alert at 09:00am CEST time, at the beginning of the day. The disadvantage is that it might get spammy with 1 alert per hour.

I think this is a good compromise, but we could increase the window or the duration or both.

### DIFF
--- a/test/onhost-canaries/terraform/main.tf
+++ b/test/onhost-canaries/terraform/main.tf
@@ -120,26 +120,56 @@ locals {
   // Platform-specific memory conditions.
   // Linux uses virtual size; Windows uses working set (physical memory committed to the process).
   memory_alert_condition_by_platform = {
-    linux = {
-      name          = "Memory usage (bytes)"
-      metric        = "max(memoryResidentSizeBytes) OR 0"
-      sample        = "ProcessSample"
-      threshold     = 42000000
-      duration      = 600
-      operator      = "above"
-      template_name = "./alert_nrql_templates/generic_metric_threshold.tftpl"
-    }
-    windows = {
-      name          = "Memory usage (bytes)"
-      # For the purpose of leak detection using memoryVirtualSizeBytes reflects better the AC memory intent of usage,
-      # as memoryResidentSizeBytes gets heavily affected by the way windows manages memory.
-      metric        = "max(memoryVirtualSizeBytes) OR 0"
-      sample        = "ProcessSample"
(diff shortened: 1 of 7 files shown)


sigilioso:
We can perfectly do it in a different PR, but would it make sense to follow the same strategy for memory usage on K8s?

test/onhost-canaries/terraform/alert_nrql_templates/derivative_metric_threshold.tftpl
DavSanchez, line 1: The name of this file can be replaced with something generic that accounts for its "time-bounded" nature, so it can be used on something else than derivatives.

Or we can add an optional template at the end of the existing file `generic_metric_threshold.tftpl` and add the whole `SINCE 10800 SECONDS AGO` to the relevant metrics.

test/onhost-canaries/terraform/main.tf
```diff
@@ -156,7 +186,7 @@ locals {
       platform = v.platform
       conditions = concat(
         local.common_alert_conditions,
-        [local.memory_alert_condition_by_platform[v.platform]]
+        local.memory_alert_condition_by_platform[v.platform]
```
DavSanchez, line 189: 👌

test/onhost-canaries/terraform/alert_nrql_templates/derivative_metric_threshold.tftpl
danielorihuela: https://github.com/newrelic/newrelic-agent-control/pull/2508/changes/364fd385dee46a8415e54480812dbd377fd24fe0

DavSanchez (dismissed):

DavSanchez (approved):

sigilioso (approved):

PR #2509, opened 2026-05-11
feat(jp): support jp endpoints

paologallinaharbur:
This PR adds support for the new jp region and removes region_parser implementation


agent-control/src/cli/common/region.rs
```diff
@@ -25,27 +29,22 @@ const OTLP_URL_US: &str = "otlp.nr-data.net";
...
-    ValueParser::from(|s: &str| -> Result<Region, String> {
-        match s.to_lowercase() {
-            s if s == "us" => Ok(Region::US),
-            s if s == "eu" => Ok(Region::EU),
-            s if s == "staging" => Ok(Region::STAGING),
-            _ => Err(format!("{s} is not a supported region")),
-        }
-    })
-}
-
```
sigilioso, line 43: Why removing this?

paologallinaharbur: why not? in the nr-auth we are solving it by clap annotations

paologallinaharbur: what am I missing?

sigilioso: We stop supporting _typos_ such as "sTaGging" or "uS" and weird things, but it is not a big deal, I just find the "ignorecase" support more natural.

I didn't realize that nr-auth resolves it differently. This is aligning the approaches, right?

agent-control/src/cli/common/region.rs
```diff
@@ -1,22 +1,26 @@
...
 const OPAMP_ENDPOINT_EU: &str = "https://opamp.service.eu.newrelic.com/v1/opamp";
+const OPAMP_ENDPOINT_JP: &str = "https://opamp.service.jp.newrelic.com/v1/opamp";
+
 const OPAMP_ENDPOINT_STAGING: &str = "https://opamp.staging-service.newrelic.com/v1/opamp";
 
 const PUBLIC_KEY_ENDPOINT_US: &str =
     "https://publickeys.newrelic.com/r/blob-management/global/agentconfiguration/jwks.json";
 const PUBLIC_KEY_ENDPOINT_EU: &str =
     "https://publickeys.eu.newrelic.com/r/blob-management/global/agentconfiguration/jwks.json";
+const PUBLIC_KEY_ENDPOINT_JP: &str =
```
sigilioso, line 17: Could we also extend the corresponding unit-tests to also consider the new endpoints? (Just to check everything is wired-up as expected)

agent-control/src/cli/common/region.rs
paologallinaharbur: yes, moreover when you use the cli it tells you which are the options, currently it only tells you "not supported"

```shell
$ cargo run  --bin newrelic-agent-control-cli -- generate-config  --region gd -h

error: invalid value 'gd' for '--region <REGION>'
  [possible values: US, EU, JP, STAGING]

```

paologallinaharbur: I am adding an "ignore_case = true" for that

paologallinaharbur: it does not exist, even the IDE seems to allucinating nowadays 🤦‍♂️

agent-control/src/cli/common/region.rs
paologallinaharbur: done!

agent-control/src/cli/common/region.rs
paologallinaharbur: it exist, but at arg level!

sigilioso (approved):
🚀

PR #2511, opened 2026-05-11
Add preload agent Linux test

yahmed153:

### DIFF
--- a/test/e2e-runner/src/linux/scenarios/preload_agent.rs
+++ b/test/e2e-runner/src/linux/scenarios/preload_agent.rs
@@ -1,0 +1,81 @@
+use crate::common::config::{DEBUG_LOGGING_CONFIG, update_config, write_agent_local_config};
+use crate::common::on_drop::CleanUp;
+use crate::common::test::retry_panic;
+use crate::common::{InstallationArgs, RecipeData};
+use crate::{
+    common::nrql,
+    linux::{
+        self,
+        install::{install_agent_control_from_recipe, tear_down_test},
+    },
+};
+use std::time::Duration;
+use tracing::info;
+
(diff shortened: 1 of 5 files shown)


test/e2e-runner/src/linux/scenarios/preload_agent.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    info!("Setup Agent Control config");
+    update_config(
+        linux::DEFAULT_AC_CONFIG_PATH,
+        format!(
+            r#"
+host_id: {test_id}
+agents:
+  nr-preload:
+    agent_type: "newrelic/com.newrelic.preload:0.1.0"
```
mvicknr, line 48: for our testing initially, will need to point to the temp .so file we have at `svcapmcontrolteam607/newrelic-agent-preload:latest`

DavSanchez: Remember that this `"newrelic/com.newrelic.preload:0.1.0"` is a reference to an *agent type* that normally maps to one of the items in the registry that currently lives in this repo ([ref](https://github.com/newrelic/newrelic-agent-control/tree/b63f22843c0a6b53059f47ad13ff0c8484a43efb/agent-control/agent-type-registry)) and is included in the AC binary. For testing purposes, definitions can live at a particular external location that AC reads and loads at startup.

We can assist you in developing this agent type definition (a YAML file) so it can be very close to a working state, but take into account that AC still lacks a functionality for running arbitrary commands after it has downloaded an OCI artifact (but this functionality is coming). As an interim solution, we could try developing something like a script that is long running (like most agents e.g. the infra-agent) that can install the preload library globally when it starts, waits and removes it when it stops or something like that.

PR #2521, opened 2026-05-13
chore(start_settings): improve handling of customCapabilities

paologallinaharbur:
I am removing "start_settings" since it was a bit redundant and was just setting defaults and we avoid doing `let Some(ref mut caps) = settings.custom_capabilities`


sigilioso (approved):
Thanks! 🚀

PR #2522, opened 2026-05-13
feat: add resolute ubuntu 26.4 to debian

vjripoll:
Add new os to packges for debian ubuntu 26.4 resolute


build/upload-schema-linux-deb.yml
```diff
@@ -15,6 +15,7 @@
         - xenial
         - bullseye
         - bookworm
+        - resolute
```
danielorihuela: Nit: Could you put it on top? Just because they are sorted.

danielorihuela (approved):

PR #2526, opened 2026-05-14
fix: impl Display for Region

DavSanchez:


test/e2e-runner/src/common/nrql.rs
```diff
@@ -23,6 +24,18 @@ pub enum Region {
     Staging,
 }
 
+impl Display for Region {
```
paologallinaharbur: I guess it was due to a test that was added after I've rebased. Intead of adding Display I was leveraging the debug :? instead.

I am fine with both, but just remove yours or mine I would say

DavSanchez: But `Debug` would/should print something like `Region::Staging` no? Different use cases for both (e.g. logging and string interpolation for a command).

Instead of `Display` I could add a different method, perhaps that way it's more clear.

DavSanchez: I see it just prints the variant and that's it. Works for the CLI commands but I see `Debug` needing to be used differently, like the CLI would stop accepting this format or the `Debug` implementation might change between Rust versions.

Instead of `Display` I have added a different method `as_cli_str`. LMKWYT!

DavSanchez: Discussed offline

vjripoll (approved):

PR #2527, opened 2026-05-14
feat: memory growth alert

danielorihuela:
Add slow memory grow alert for onhost and k8s.


gsanchezgavier (approved):
Thanks! we can try who the new ones perform , and also have the old ones

PR #2544, opened 2026-05-19
fix(host): clean fleet data on agent removal (NR-560818)

gsanchezgavier:
on-host AC was leaking sub-agent fleet data between remote agent removals, while the k8s implementation is not.

The fix introduces OnHostCleaner, a ResourceCleaner impl  that delegates per-agent fleet-data cleanup to the same two abstractions that wrote it — InstanceIDStorer and ConfigRepository — via their delete/delete_remote methods. It runs on both clean() call sites in agent_control.rs (recreate + remove).

This change is introducing another discrepancy between host/k8s, regarding the case where an `agent_type_id` is updated through the remote AC config keeping the `agent_id` (added a comment into the code) . This case will be need to be addressed whenever FC starts to dynamically change this.


agent-control/tests/on_host/scenarios/remote_agent_removal.rs
```diff
@@ -0,0 +1,195 @@
...
+use newrelic_agent_control::agent_control::agent_id::AgentID;
+use newrelic_agent_control::agent_control::defaults::AGENT_FILESYSTEM_FOLDER_NAME;
+use newrelic_agent_control::agent_control::run::BasePaths;
+use newrelic_agent_control::agent_control::run::on_host::AGENT_CONTROL_MODE_ON_HOST;
+use opamp_client::opamp::proto::RemoteConfigStatuses;
+use std::time::Duration;
+use tempfile::tempdir;
+
+/// Given a agent-control whose local configuration has no agents and then a valid remote configuration with an agent
+/// is set through OpAMP. Then the agent is removed via a new remote configuration. Finally, the agent is added again.
```
sigilioso: Nit: I'd reword a bit the comment

```suggestion
/// Check config persistence/removal on the following scenario
/// 1. AC has no agents
/// 2. One sub-agent is added (config is persisted)
/// 3. The sub-agent is removed (config is removed)
/// 4. The sub-agent is added back (config is persisted and the identifier is different from 2)
```

agent-control/tests/on_host/scenarios/remote_agent_removal.rs
```diff
@@ -0,0 +1,195 @@
...
+    });
+
+    let agent_filesystem_path = base_paths
+        .remote_dir
+        .join(AGENT_FILESYSTEM_FOLDER_NAME)
+        .join(agent_id)
+        .join(dir_entry)
+        .join(file_path);
+
+    // Check that the file was created
```
sigilioso, line 121: Should we also check the persisted instance-id or it is redundant given that we call to `get_instance_id(&AgentID::try_from(agent_id).unwrap(), base_paths.clone());` to setup the remote config?

agent-control/tests/on_host/scenarios/remote_agent_removal.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    // 3. Add agent again
+    opamp_server.set_config_response(ac_instance_id.clone(), agents_config_with_agent.as_str());
+
+    let new_subagent_instance_id =
+        get_instance_id(&AgentID::try_from(agent_id).unwrap(), base_paths.clone());
+
+    assert_ne!(
+        new_subagent_instance_id, subagent_instance_id,
+        "identifier should have been be recreated"
```
sigilioso:
```suggestion
        "identifier should have been recreated"
```

agent-control/src/agent_control.rs
```diff
@@ -519,6 +519,8 @@ where
                 Some(old_sub_agent_config) => {
                     info!(%agent_id, "Recreating SubAgent");
                     self.recreate_sub_agent(&agent_identity, running_sub_agents)?;
+                    // onhost: we are removing old sub agent fleet data
+                    // k8s: fleet data gets reused, only the CM annotation for agent_type_id is changed.
```
sigilioso, line 523: Thanks for adding the comment, should we add a TODO or track it somewhere? It needs to be clear that we need to "fix" this when agent-type updates are supported.

sigilioso (dismissed):
Thanks for addressing this! I've left some _nits_ only. LGTM

agent-control/src/agent_control.rs
gsanchezgavier: Indeed, I'm adding it as part of the remote agent types feature .

agent-control/tests/on_host/scenarios/remote_agent_removal.rs
gsanchezgavier: exactly we are getting both instance_id (old and new) and validating that they are correct when using the opamp server, this test assumes that AC is persisting them correctly.

agent-control/tests/on_host/scenarios/remote_agent_removal.rs
gsanchezgavier: the "when" and "then" is much complex, so i would prefer to keep just the Given but i will listed , is more readable

sigilioso (approved):
🚀

PR #2545, opened 2026-05-19
test: check capabilities and custom capabilities in integration tests [NR-566846]

sigilioso:
This PR extends the integration tests to check that `capabilities` and  `custom_capabilities` are reported as expected in OpAMP messages.

### DIFF
--- a/agent-control/tests/common/attributes.rs
+++ b/agent-control/tests/common/attributes.rs
@@ -53,19 +53,17 @@ pub fn check_identifying_attributes_contains_expected(
     )
     .map_err(|e| format!("Identifying attributes missing required elements: {e}"))
 }
-
-#[allow(dead_code)] // helper used on unix only
+#[cfg(unix)]
 /// Checks that the latest `CustomCapabilities` match the `expected`
 pub fn check_custom_capabilities_match(
     opamp_server: &FakeServer,
     instance_id: &InstanceID,
-    expected: Vec<String>,
+    expected: HashSet<String>,
 ) -> Result<(), String> {
     let current = opamp_server
         .get_custom_capabilities(instance_id.clone())
         .ok_or_else(|| "Custom capabilities not reported yet".to_string())?;
     let current: HashSet<String> = HashSet::from_iter(current.capabilities.clone());
-    let expected: HashSet<String> = HashSet::from_iter(expected);
     if current != expected {
         return Err(format!(
             "custom capabilities don't match expected.  Expected: {:?}. Found: {:?}",
@@ -75,7 +73,7 @@ pub fn check_custom_capabilities_match(
     Ok(())
 }
 
-#[allow(dead_code)] // helper used on unix only
+#[cfg(unix)]
 /// Checks that the latest `Capabilities` match the `expected`
 pub fn check_capabilities_match_expected(
     opamp_server: &FakeServer,
--- a/agent-control/tests/common/attributes.rs
+++ b/agent-control/tests/common/attributes.rs
@@ -1,3 +1,4 @@
+#[cfg(unix)]
 use std::collections::HashSet;
 
 use fake_opamp_server::FakeServer;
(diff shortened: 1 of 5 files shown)


agent-control/tests/k8s/scenarios/attributes.rs
```diff
@@ -117,6 +122,17 @@ agents:
...
+        check_custom_capabilities_contains(
+            &server,
+            &instance_id,
+            vec![SIGNATURE_CUSTOM_CAPABILITY.to_string()],
+        )?;
+        check_custom_capabilities_does_not_contain(
+            &server,
+            &instance_id,
+            vec![K8S_CONFIG_ONLY_AGENTS_CUSTOM_CAPABILITY.to_string()], // only when cd_enabled=false
+        )?;
```
gsanchezgavier: would make sense to change this so it only checks that the expected capability is reported and nothing more?

gsanchezgavier (dismissed):
thanks just left a minor suggestion

agent-control/tests/k8s/scenarios/attributes.rs
sigilioso: Addressed here: [b8bb2b4](https://github.com/newrelic/newrelic-agent-control/pull/2545/commits/b8bb2b417969289626771eaaee98a1c4821d2a81)

agent-control/tests/common/attributes.rs
```diff
@@ -52,6 +54,46 @@ pub fn check_identifying_attributes_contains_expected(
     .map_err(|e| format!("Identifying attributes missing required elements: {e}"))
 }
 
+#[allow(dead_code)] // helper used on unix only
```
danielorihuela: Why not `#[cfg(unix)]`?

agent-control/tests/common/attributes.rs
```diff
@@ -52,6 +54,46 @@ pub fn check_identifying_attributes_contains_expected(
...
+/// Checks that the latest `CustomCapabilities` match the `expected`
+pub fn check_custom_capabilities_match(
+    opamp_server: &FakeServer,
+    instance_id: &InstanceID,
+    expected: Vec<String>,
```
danielorihuela: Nit: Use `HashSet<String>` or `Into<HashSet<String>>` as input. It makes clear order doesn't matter from the caller.

agent-control/tests/common/attributes.rs
```diff
@@ -52,6 +54,46 @@ pub fn check_identifying_attributes_contains_expected(
...
+    instance_id: &InstanceID,
+    expected: u64,
+) -> Result<(), String> {
+    let current = opamp_server
+        .get_capabilities(instance_id.clone())
+        .ok_or_else(|| "Capabilities not reported yet".to_string())?;
+
+    if current != expected {
+        return Err(format!(
+            "capabilities don't match. Expected: {expected:#b}, Found: {current:#b}"
```
danielorihuela, line 91: Nit: Not sure how useful the binary representation is. Maybe we can add both the decimal and binary representation.

agent-control/tests/common/attributes.rs
sigilioso: Because it didn't come to mind. Changing it, thanks!

agent-control/tests/common/attributes.rs
sigilioso: I think the binary representation makes sense [in this case](https://github.com/newrelic/newrelic-opamp-rs/blob/main/proto/src/proto.rs#L82). We can always update the error message if we don't find it that useful in practice.

agent-control/tests/common/attributes.rs
sigilioso: [a280b9a](https://github.com/newrelic/newrelic-agent-control/pull/2545/commits/a280b9a78747431f8fe344bde38b3746ff85805a)

agent-control/tests/common/attributes.rs
sigilioso: [a280b9a](https://github.com/newrelic/newrelic-agent-control/pull/2545/commits/a280b9a78747431f8fe344bde38b3746ff85805a)

danielorihuela (dismissed):

danielorihuela (approved):

PR #2546, opened 2026-05-19
chore: self-update e2e test (currentToLatest)

gsanchezgavier:
- Fix the public key url and default values deserializing for self update configs
- Adds the e2e test that checks self-update from current compiled version to latest published

### DIFF
--- a/test/e2e-runner/src/common/docker_hub.rs
+++ b/test/e2e-runner/src/common/docker_hub.rs
@@ -1,0 +1,32 @@
+use crate::common::test::TestResult;
+use serde::Deserialize;
+
+const DOCKER_HUB_API: &str = "https://hub.docker.com/v2/repositories";
+
+#[derive(Deserialize)]
+struct TagsResponse {
+    results: Vec<TagEntry>,
+}
+
+#[derive(Deserialize)]
+struct TagEntry {
+    name: String,
+}
+
+pub fn latest_published_ac_tag() -> TestResult<String> {
+    latest_tag("newrelic/agent-control-artifacts")
+}
+
+/// Fetches the most recent tag published for a Docker Hub repository.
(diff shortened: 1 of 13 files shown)


test/e2e-runner/src/common/docker_hub.rs
```diff
@@ -0,0 +1,32 @@
...
+    let url = format!("{DOCKER_HUB_API}/{repository}/tags/?page_size=100");
+    let response: TagsResponse = reqwest::blocking::get(&url)?.error_for_status()?.json()?;
+
+    // look for the first non signature tag.
+    response
+        .results
+        .into_iter()
+        .map(|entry| entry.name)
+        .find(|name| !name.starts_with("sha"))
+        .ok_or_else(|| format!("tag not found in {repository}").into())
```
sigilioso: I guess this usually works but the order in the API results is undefined. Would it make sense to:

- Ignore the ones whose name starts with `"sha"`
- Sort by name (SemVer works with alphabetical order)
- Get the latest

sigilioso (commented):
I'd just double-check if we are sure of getting the latest tag from dockerhub. 

Everything else looks good to me.

test/e2e-runner/src/common/docker_hub.rs
gsanchezgavier: this was the first AI proposal, it was a bit complex and introduced a semver dependency . Is true that the API doesn't explicitly mention the order , but now i'm having second thought about this not because of the order of the API but the order we manually push the tags when backfilling or thing like that. I will revert back to the more correct way

gsanchezgavier: done! ended up quite clear, just the downside to compile that extra dep but ... 🦓

sigilioso (approved):
LGTM 🚀

PR #2549, opened 2026-05-20
ci: dhat-enabled workflow

DavSanchez:
## Summary

Adds on-demand workflows and supporting changes to produce DHAT-enabled Agent Control artifacts for heap profiling, without disturbing the regular release pipeline.

### CI — two `workflow_dispatch` workflows
- **`on_demand_dhat_heap_build.yml`** — builds `dhat-heap` binaries for Linux (`amd64`, `arm64`) and Windows (`amd64`) using the `release-debug` profile, and uploads them as artifacts tagged with the user-supplied `tag_name`. The CLI is intentionally built without the `dhat-heap` feature (the profiler is only meaningful for the long-running agent process).
- **`on_demand_dhat_heap_image.yml`** — builds and pushes a multi-arch (`linux/amd64`, `linux/arm64`) Agent Control K8s image to `ghcr.io/<org>/newrelic-agent-control-dev:dhat-heap-<image-tag>`, intended as a drop-in replacement for the standard image in a cluster.

Both are manually triggered to avoid running this on every push — DHAT-instrumented builds are only useful for targeted profiling sessions.

### Configurable DHAT output path (`agent-control/src/bin/main_k8s.rs`, `main_onhost.rs`)
The DHAT profiler now honours `DHAT_HEAP_OUTPUT_PATH` / `DHAT_AD_HOC_OUTPUT_PATH` env vars when set, falling back to the previous default behaviour otherwise. This is required for containerised runs: the profile is written on `Drop`, so without redirecting it to a mounted volume the report is lost when the pod terminates.

### `build/scripts/build_binary.sh`
- Forwards `CARGO_FEATURES` to `cargo zigbuild` so the workflows can opt into `dhat-heap` without a separate script.
- Derives `BUILD_OUT_DIR` from `BUILD_MODE` (e.g. `release-debug` → `release-debug`) instead of hard-coding `release`, so non-standard profiles land their binaries where the rest of the pipeline expects them.
- Logs the resolved profile and output dir for easier debugging.

### Docs (`docs/DEVELOPMENT.md`)
End-to-end guide for running heap profiling in Kubernetes: building the image, provisioning a `PersistentVolume` for the report, wiring up `DHAT_HEAP_OUTPUT_PATH` and volume mounts in the `agent-control-deployment` Helm values, retrieving `dhat-heap.json` via a debug pod, and a note on the memory-limit spike caused by DHAT serialising the heap on shutdown.

### DIFF
--- a/.github/workflows/on_demand_dhat_heap_build.yml
+++ b/.github/workflows/on_demand_dhat_heap_build.yml
@@ -1,0 +1,97 @@
+name: ⏯️ Build DHAT profiling binaries
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        description: 'Tag name to associate with the build'
+        required: true
+        type: string
+
+permissions:
+  contents: read
+
+jobs:
+  build-linux:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        arch: [arm64, amd64]
+    name: Build dhat-heap Linux binaries (${{ matrix.arch }})
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          fetch-depth: 0
+
+      - uses: ./.github/actions/install-rust-toolchain
+        with:
+          targets: >-
+            aarch64-unknown-linux-musl,
+            x86_64-unknown-linux-musl
+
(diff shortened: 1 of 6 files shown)


gsanchezgavier (approved):
thanks for adding this !

PR #2550, opened 2026-05-20
feat!: remove self instrumentation trace export

gsanchezgavier:
This PR removes self-instrumentation traces export feature. 

We had issues when wrongly introducing long-lived spans that accumulated events and never closed, leaking memory. We don't consider these spans useful for this codebase. 

Logs and metrics export are unchanged. All #[instrument] / info_span! instrumentation stays — it still decorates log records via the file/stderr formatter.

Changes:
  - Drop the trace exports
  - Bump the OTel crate family
  - Introduce some integration tests for the remaining sefl-instrumentation feature.

### DIFF
--- a/agent-control/src/instrumentation/config/otel.rs
+++ b/agent-control/src/instrumentation/config/otel.rs
@@ -12,9 +12,9 @@ const DEFAULT_CLIENT_TIMEOUT: Duration = Duration::from_secs(30);
 /// Default interval for exporting metrics.
 const DEFAULT_METRICS_EXPORT_INTERVAL: Duration = Duration::from_secs(60);
 /// Default maximum batch size [logs::BatchLogProcessor] for details.
-const DEFAULT_BATCH_MAX_SIZE: usize = 512;
+const DEFAULT_LOG_BATCH_MAX_SIZE: usize = 512;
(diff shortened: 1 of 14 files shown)


agent-control/src/instrumentation/config/otel.rs
```diff
@@ -12,15 +11,13 @@ use crate::http::config::ProxyConfig;
 const DEFAULT_CLIENT_TIMEOUT: Duration = Duration::from_secs(30);
 /// Default interval for exporting metrics.
 const DEFAULT_METRICS_EXPORT_INTERVAL: Duration = Duration::from_secs(60);
-/// Default maximum batch size [trace::BatchSpanProcessor] for details.
+/// Default maximum batch size [logs::BatchLogProcessor] for details.
 const DEFAULT_BATCH_MAX_SIZE: usize = 512;
-/// Default scheduled delay [trace::BatchSpanProcessor] for details.
+/// Default scheduled delay [logs::BatchLogProcessor] for details.
```
DavSanchez, line 16: Should we specify in the const name that these batches are about the logger?

DavSanchez (dismissed):
thanks!

DavSanchez (approved):
Thanks!

PR #2553, opened 2026-05-21
fix: onhost canaries staging alerts slack channel

danielorihuela:
Onhost canaries staging was using the wrong slack channel url. This PR fixes it.
Launched https://github.com/newrelic/newrelic-agent-control/actions/runs/26231951907 and it worked.


gsanchezgavier (approved):
thanks

PR #2517, opened 2026-05-12
Add preload linux scenario test

yahmed153:
Add preload agent linux scenario test, for the preload-agent

PR #2554, opened 2026-05-21
ci(apm): preload PoC basic testing improvements

DavSanchez:
## Summary

This PR builds on top of #2517 and proposes a few improvements to the preload agent E2E scenario, while keeping the changes scoped to a PoC for APM support on-host through the preload library.

### Why not commit the agent type definition

The original branch added `agent-control/agent-type-registry/newrelic/com.newrelic.preload-0.1.0.yaml` to the main codebase. Since this work is still a PoC, we shouldn't be shipping a `com.newrelic.preload` agent type as part of the registry yet. Instead, this PR:

- Removes the YAML from `agent-type-registry/`.
- Inlines a minimal version of the agent type definition inside the E2E scenario, written at runtime to the dynamic agent types directory (`/etc/newrelic-agent-control/dynamic-agent-types/preload.yaml`).

This way, the PoC scenario can exercise the full flow (Agent Control loading a custom agent type, downloading the OCI artifact, etc.) without leaking PoC artifacts into the distributed registry.

### Test scenario improvements

A few minor assertions were added to the `PreloadAgent` scenario so that failures are caught early and surface clear errors instead of failing later in opaque ways while the actual test scenario takes shape:

- Wait (with retries) until the preload OCI package is downloaded and extracted under the AC-managed `stored_packages/preload-agent` directory.
- Locate the shared library (`.so`) inside the extracted package and fail loudly if none is found.
- Append the resolved `.so` path to `/etc/ld.so.preload` (previously the path was a placeholder and the file path itself was wrong: `/ec/ld.so.preload`).

The local agent config sent to AC was also trimmed to the minimum needed for this PoC (just `version`), since the previous fields were not consumed by the agent type used here.

### CI changes

The `component_onhost_e2e.yaml` workflow was updated so the new scenario actually runs:

- Added `preload-agent` to the scenarios matrix.
- Introduced `HOST_E2E_PRELOAD_VERSION` and threaded it through as `--preload-version` to the e2e runner.


yahmed153 (approved):

test/e2e-runner/src/linux/scenarios/preload_agent.rs
```diff
@@ -35,6 +45,40 @@ pub fn test_installation_with_preload_agent(args: InstallationArgs) {
 
     let preload_agent_id = "nr-preload";
 
+    info!("Writing custom preload agent type definition");
+    let custom_agent_type_path = format!("{DYNAMIC_AGENT_TYPES_DIR}/preload.yaml");
+    let custom_agent_type = r#"namespace: newrelic
```
mvicknr, line 51: Claude found this issue:

🔴 Bug: ${{ ... }} won't be substituted by Agent Control

test/e2e-runner/src/linux/scenarios/preload_agent.rs:50-77 — the YAML is a plain r#"..."# raw string, not a format! template, so the doubled braces are not escapes — they end up literally in the file:

repository: ${{nr-var:oci.repository}}
version: ${{nr-var:version}}

Compare with agent-control/tests/on_host/scenarios/oci_auth.rs:43-54, where ${{...}} is inside format!(...) and correctly resolves to ${...}. AC's template engine expects the single-brace form
(${nr-var:version}), so the inlined agent type will fail to resolve version and the OCI download won't get a valid tag.

Fix: use single braces (${nr-var:...}) since this is a raw string, or wrap the literal in format! if you want to keep the doubled form for visual consistency with the other tests.

test/e2e-runner/src/linux/scenarios/preload_agent.rs
```diff
@@ -51,35 +95,47 @@ agents:
...
+        "preload package download assertion",
+        || assert_preload_package_downloaded(&package_dir),
+    );
 
-    linux::service::restart_service(linux::SERVICE_NAME);
+    info!("Searching for shared library inside extracted package");
+    let find_so_command = format!(r#"find {package_dir} -type f -name "*.so" | head -n 1"#);
+    let so_path = exec_bash_command(&find_so_command)
+        .unwrap_or_else(|err| panic!("Failed to find shared library in package: {err}"));
+    let so_path = so_path.lines().last().unwrap_or("").trim().to_string();
```
mvicknr: Also from Claude;

🔴 Bug: so_path extraction picks up the wrapper text, not the actual path

test/e2e-runner/src/linux/scenarios/preload_agent.rs:115-122:

let so_path = exec_bash_command(&find_so_command)...;
let so_path = so_path.lines().last().unwrap_or("").trim().to_string();

exec_bash_command (test/e2e-runner/src/linux/bash.rs:22) does not return raw stdout — it returns a wrapped, multi-line string:
command
<cmd>
success
Stdout: <stdout>
Stderr: <stderr>

The Stderr: line is always last, so so_path ends up as "Stderr:" (or empty after the colon depending on trim). The if so_path.is_empty() guard does not catch this, and the next step appends Stderr: to
/etc/ld.so.preload. None of the existing callers parse stdout from exec_bash_command — this is the first one, and it doesn't work.

Fix options (pick one):
- Add a sibling helper that returns just stdout (e.g., exec_bash_command_stdout), then use it here.
- Run find ... -print -quit on the host and have the helper write its result to a file you read_to_string.
- Parse the wrapper: extract the line starting with Stdout:  and strip the prefix — least preferred since it bakes in a fragile dependency on the wrapper format.

test/e2e-runner/src/linux/scenarios/preload_agent.rs
DavSanchez: Fixed https://github.com/newrelic/newrelic-agent-control/pull/2554/commits/e91cd32008145315cab4898dabc2400b0e0c76e6

test/e2e-runner/src/linux/scenarios/preload_agent.rs
DavSanchez: Changed the strategy used here: https://github.com/newrelic/newrelic-agent-control/pull/2554/commits/fec0466cebb9b0d2e4f0a59d2672771828b9276d

PR #2559, opened 2026-05-26
feat(pcg_config): keep same variable structure

paologallinaharbur:
PCG wants to keep the same file structure among the two agentTypes.

### DIFF
--- a/agent-control/agent-type-registry/newrelic/pipeline-control-gateway-config-0.1.0.yaml
+++ b/agent-control/agent-type-registry/newrelic/pipeline-control-gateway-config-0.1.0.yaml
@@ -3,10 +3,12 @@ name: com.newrelic.pipeline_control_gateway_config
 version: 0.1.0
 variables:
   k8s:
-    config:
-      description: "OTel Collector configuration for the Pipeline Control Gateway"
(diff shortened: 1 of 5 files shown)


agent-control/agent-type-registry/newrelic/pipeline-control-gateway-config-0.1.0.yaml
```diff
@@ -3,10 +3,12 @@ name: com.newrelic.pipeline_control_gateway_config
 version: 0.1.0
 variables:
   k8s:
-    config:
-      description: "OTel Collector configuration for the Pipeline Control Gateway"
-      type: yaml
-      required: true
+    chart_values:
```
gsanchezgavier, line 6: would make sense to documente here why ?

gsanchezgavier (approved):

agent-control/agent-type-registry/newrelic/pipeline-control-gateway-config-0.1.0.yaml
paologallinaharbur: I'll add a comment in the follow up Pr

PR #2560, opened 2026-05-26
feat!: remove common variables types [NR-568035]

gsanchezgavier:
Removes the `common` section of agent type variables 

I was planning to add a single changelog entry alongside the PR that flattens the environment to reduce the noise

### DIFF
--- a/agent-control/src/agent_type/README.md
+++ b/agent-control/src/agent_type/README.md
@@ -28,7 +28,7 @@ The `variables` section allows developers to define variables that end users can
 
 ```yaml
 variables:
-  common:
+  linux:
     config_agent:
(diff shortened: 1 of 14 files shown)


sigilioso (approved):
🧹 🎉  LGTM!

`feat!` marks this as breaking because we drop support for this `common` section in the Agent Types but the only broken scenario would be the one where a custom Agent Type defining such `common` section was used, right?

gsanchezgavier:
> 🧹 🎉 LGTM!
> 
> `feat!` marks this as breaking because we drop support for this `common` section in the Agent Types but the only broken scenario would be the one where a custom Agent Type defining such `common` section was used, right?

yes exactly

PR #2563, opened 2026-05-27
chore: move agent-type registry to its own module [NR-568043]

sigilioso:
This PR refactors the code handling the Agent Type registry by moving it to its own `registry` module under `agent_type`. It also renames some structs to make them consistent:
- `AgentRegistry` --> `AgentTypeRegistry`
- `AgentRepositoryError` --> `AgentTypeRepositoryError`

The purpose of this refactor is making room for new registry implementations without growing the current flat `agent_type` module further.

### DIFF
--- a/agent-control/src/agent_type/registry.rs
+++ b/agent-control/src/agent_type/registry.rs
@@ -18,8 +18,7 @@ pub enum AgentTypeRegistryError {
 
 /// AgentTypeRegistry stores and loads Agent types.
 pub trait AgentTypeRegistry {
-    // get returns an Agent type given a definition.
-    // TODO: evaluate if returning an owned value is needed, CoW?
+    // Returns an Agent type given a definition.
     fn get(&self, name: &str) -> Result<AgentTypeDefinition, AgentTypeRegistryError>;
 }
 
--- a/agent-control/src/agent_type/registry.rs
+++ b/agent-control/src/agent_type/registry.rs
@@ -1,9 +1,11 @@
+pub mod embedded;
+
 use thiserror::Error;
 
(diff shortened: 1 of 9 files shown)


agent-control/src/agent_type/registry.rs
```diff
@@ -14,11 +16,10 @@ pub enum AgentRepositoryError {
     ValueConversion(#[from] serde_json::Error),
 }
 
-/// AgentRegistry stores and loads Agent types.
-pub trait AgentRegistry {
-    // get returns an Agent type given a definition.
-    // TODO: evaluate if returning an owned value is needed, CoW?
```
sigilioso, line 20: I've just removed the TODO because we have only two _production_ calls:

- [EffectiveAgentAssembler](https://github.com/newrelic/newrelic-agent-control/blob/6a439079529c3fdcca1906e8b301e5fa3bbc86e1/agent-control/src/sub_agent/effective_agents_assembler.rs#L168-L178) actually needs ownership.
- [ConfigValidator](https://github.com/newrelic/newrelic-agent-control/blob/6a439079529c3fdcca1906e8b301e5fa3bbc86e1/agent-control/src/agent_control/config_validator.rs#L43-L49): it doesn't need to own the AgentTypeDefinition but it doesn't need the reference neither. If we are worried about the performance we could add an `exits` method instead.

I don't think that `CoW` helps here, but we could also consider returning the `&AgentTypeDefinition` reference and cloning it explicitly when on the `EffectiveAgentAssembler`, but considering it involves introducing some additional complexity by introducing the lifetime in the trait method. I'm not sure if its worth it.

danielorihuela (approved):

PR #2565, opened 2026-05-27
feat: extend agent-type-registry trait [NR-568043]

sigilioso:
The `RegistryDynamicConfigValidator` only needs to verify that a given agent type id exists in the registry, but it was forced to call `AgentTypeRegistry::get`, which returns (and clones) the full AgentTypeDefinition only for the result to be discarded. 

This PR adds an explicit `contains` method to the trait so callers that just need an existence check don't pay the cost of cloning the definition unnecessarily.


sigilioso:
I'm closing it for now as it actually makes the Registry extension harder. We can bring it back when the registry changes are ready.

PR #2570, opened 2026-05-29
fix: update messages for canaries alerts

vjripoll:
Improve alert messages to clarify when an alert is trigger or closed with better structure.

### DIFF
--- a/test/onhost-canaries/terraform/main.tf
+++ b/test/onhost-canaries/terraform/main.tf
@@ -391,7 +391,9 @@ resource "newrelic_notification_channel" "slack_channel" {
 
   property {
     key   = "payload"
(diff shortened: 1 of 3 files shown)


sigilioso (commented):
Thanks!

Would it make sense to use [jsonencode](https://developer.hashicorp.com/terraform/language/functions/jsonencode) to make the block easier to read/update?

Another alternative could be using a [templatefile](https://developer.hashicorp.com/terraform/language/functions/templatefile) and maybe we could even avoid repeating almost the same block in both files.

sigilioso (approved):
LGTM

Would it make sense to apply the change as soon as possible? Just to avoid potential issues in nightlies / pre-releases.

PR #2571, opened 2026-06-01
chore: agent type id type enforcement on registry

sigilioso:
The `AgentTypeRegistry` trait keyed lookups on a `&str` (the stringified fully-qualified name, e.g. `"ns/name:0.0.1"`). Every caller had to call .to_string() on an AgentTypeID it already held, only for the registry to use that string as a HashMap key. This is a stringly-typed seam in an otherwise typed path: nothing prevented passing an arbitrary, malformed string, and the round-trip through String was pure boilerplate.

This PR pushes the `AgentTypeID` type all the way through the registry, so the type system enforces that lookups are done with a valid, parsed identifier rather than a free-form string.

### DIFF
--- a/agent-control/src/agent_type/registry.rs
+++ b/agent-control/src/agent_type/registry.rs
@@ -17,9 +17,9 @@ pub enum AgentTypeRegistryError {
     ValueConversion(#[from] serde_json::Error),
 }
 
-/// Returns the corresponding [AgentTypeDefinition] given an [AgentTypeID].
+/// Defines how to return an [AgentTypeDefinition] given an identifier.
 pub trait AgentTypeRegistry {
-    // Returns an Agent type given its id.
+    /// Returns an Agent type given its id.
     fn get(
         &self,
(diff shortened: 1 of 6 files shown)


agent-control/src/agent_type/registry.rs
```diff
@@ -16,10 +17,13 @@ pub enum AgentTypeRegistryError {
...
 
-/// AgentTypeRegistry stores and loads Agent types.
+/// Defines how to return an [AgentTypeDefinition] given an identifier.
 pub trait AgentTypeRegistry {
-    // Returns an Agent type given a definition.
-    fn get(&self, name: &str) -> Result<AgentTypeDefinition, AgentTypeRegistryError>;
+    /// Returns an Agent type given its id.
+    fn get(
+        &self,
+        agent_type_id: &AgentTypeID,
```
gsanchezgavier, line 25: way cleaner!

gsanchezgavier (approved):
thanks!

PR #2572, opened 2026-06-01
feat: flatten platform on agent type definition schema (NR-568041)

gsanchezgavier:
Flattens AgentTypeDefinition so each YAML file targets a single (platform, operating_system) pair instead of nesting linux/windows/k8s subkeys under variables and deployment.

- All embedded definitions were migrated (see not below about validation).
- Embedded definitions names follow the OCI tag naming.
- Now the Embedded Registry filters definition by environment at construction.

### Some other remarkable details:
- EmbeddedRegistry::new(env, dynamic_path) — now requires an Environment. Only loads embedded YAMLs whose id maps to that env; dynamic agent types from disk are filtered the same way. Definitions whose YAML doesn't carry a valid platform are silently dropped. Validation lives in Environment::try_from(&AgentTypeID).
- Result: a binary running on Linux only sees Linux agent types in its registry; same for Windows and k8s.
- AgentControlRunner::running_mode field removed (the env is now baked into the registry at construction).
- Free function build_agent_type deleted (callers use AgentTypeDefinition::with_constraints).

Agent types where also validated with a script performing:
```
For each new file <platform>[-<os>]-<name>-<version>.yaml:

1. Parse the filename. Prefix → env_key (host-linux → linux, host-windows → windows, kubernetes → k8s). Stem <name>-<version> → look up old file <name>-<version>.yaml on main.
2. Load both files with yaml.safe_load (parses to dicts/lists, ignores comments/whitespace).
3. Compare:
  - new["namespace"] vs old["namespace"], same for name, version.
  - new["variables"] vs old["variables"][env_key] (i.e., the env subtree of the old variables map).
  - new["deployment"] vs old["deployment"][env_key].
4. Recursive deep-equal of those subtrees. Flag any key present on one side but not the other, any value mismatch, any list-length or type mismatch.
```


agent-control/src/agent_control/run/k8s.rs
```diff
@@ -198,7 +198,6 @@ impl AgentControlRunner {
             config_repository: yaml_config_repository.clone(),
             effective_agents_assembler: agents_assembler,
             sub_agent_publisher: self.sub_agent_publisher,
-            ac_running_mode: self.running_mode,
```
sigilioso, line 201: Finally! 🎉

agent-control/src/agent_type/definition/agent_type_validation_tests.rs
```diff
@@ -628,6 +628,12 @@ static AGENT_TYPE_EBPF: LazyLock<AgentTypeValuesTestCase> =
             ]),
         }
         .into(),
+        values_linux: AgentTypeValues {
+            // TODO test on linux needs to be added
```
sigilioso, line 670: Do we plan to do it as a follow-up?

agent-control/src/agent_type/definition/agent_type_validation_tests.rs
```diff
@@ -652,66 +658,57 @@ fn get_agent_type_test_cases() -> impl Iterator<Item = &'static AgentTypeValuesT
...
-    for case in get_agent_type_test_cases() {
-        assert!(
-            registry
-                .get(&AgentTypeID::try_from(case.agent_type).unwrap())
-                .is_ok(),
-            "Agent type {} not found",
-            case.agent_type
-        );
-    }
+fn registry_ids_across_envs(env: Environment) -> std::collections::HashSet<AgentTypeID> {
```
sigilioso: Is the name accurate? If it is receiving `env`, it doesn't get the ids across envs, right? It looks like a leftover, maybe the `HashSet` is not needed anymore?

agent-control/src/agent_type/registry/embedded.rs
```diff
@@ -13,17 +14,12 @@ include!(concat!(
 
 /// Defines an [AgentTypeRegistry] by keeping AgentTypeDefinitions in memory.
 ///
-/// Its default implementation, loads the AgentTypeDefinitions from yaml files which are embedded into the binary
-/// at compilation time. Check out the agent-control build script for details.
+/// The embedded YAMLs cover several environments (linux, windows, kubernetes); definitions
+/// not matching the requested [Environment] are skipped at construction time. This way the
+/// registry only contains entries the running binary can actually use.
```
sigilioso, line 19: We could go one step further and consider the _environment_ (we know it when we build the binary) and only include the target Agent Types in the `build.rs`.

However, the `build.rs` can become too complex and **it is probably not worth it**.

agent-control/src/agent_type/agent_type_id.rs
```diff
@@ -14,15 +15,49 @@ pub enum AgentTypeIDError {
...
+/// the id was loaded from, and do not participate in equality. This way an id parsed from a
+/// YAML definition matches one built from its FQN string when used as a key in a registry,
+/// and existing FQN-based references keep working unchanged.
+#[derive(Debug, Clone)]
 pub struct AgentTypeID {
     name: String,
     namespace: String,
     version: Version,
+    platform: Option<Platform>,
+    operating_system: Option<OperatingSystem>,
```
sigilioso: Have you considered two different types?

```rust
pub struct AgentTypeID {
    name: String,
    namespace: String,
    version: Version,
}

pub struct AgentTypeMetadata {
    id: AgentTypeID,
    platform: Platform,
    operating_system: OperatingSystem
}
```

Checking if metadata matches an ID is still trivial:

```rust
assert!(metadata.id == agent_type_id); 
```

But we don't need to handle the `Option` for the platform and operating_system (we know they will be there when building the struct from the AgentType's yaml)

agent-control/src/agent_type/agent_type_id.rs
```diff
@@ -109,10 +147,43 @@ impl TryFrom<&str> for AgentTypeID {
             name,
             namespace,
             version,
+            platform: None,
+            operating_system: None,
```
sigilioso: The fact of then being None for one case and necessarily Some() for another, is what makes me think they should be different types.

agent-control/src/agent_type/definition.rs
```diff
@@ -7,65 +7,101 @@
...
 use std::path::Path;
 use tracing::{debug, warn};
 
-/// AgentTypeDefinition represents the definition of an [AgentType]. It defines the variables and runtime for any supported
-/// environment.
-#[derive(Debug, PartialEq, Clone, Deserialize)]
-#[serde(deny_unknown_fields)]
+/// The agent type as parsed from a YAML file. Variables are still in their raw
+/// `VariableDefinitionTree` form. Apply [AgentTypeDefinition::with_constraints] to materialize
+/// the [AgentType] used for rendering.
```
sigilioso, line 32: **Out of scope:** considering that `with_constraints` actually adds some runtime information to the AgentType (in other words: `AgentTypeDefinition --> with_constraints(...) --> AgentType`) we could consider moving such constraints out of the `AgentType` type and store them in the component that actually used them. Then, we could keep one type only: `AgentType == AgentTypeDefinition`.

Let's consider that refactor after this is merged.

agent-control/src/agent_type/definition.rs
```diff
@@ -357,9 +368,10 @@ deployment:
 name: nrdot
 namespace: newrelic
 version: 0.0.1
+platform: host
+operating_system: linux
```
sigilioso, line 449: Should we add a test case with the supported variants of platform/operating-system and some errors? Maybe it is already covered by the current tests but specific unit tests for AgentTypeDefinition's deserialize could help use to assure that there are no gaps.

agent-control/src/agent_type/render.rs
```diff
@@ -250,19 +245,24 @@ pub(crate) mod tests {
...
+        match runtime_config.deployment {
+            rendered::Deployment::Host(on_host) => on_host,
+            rendered::Deployment::K8s(_) => unreachable!("expected host deployment"),
+        }
+    }
+
+    fn extract_k8s(runtime_config: Runtime) -> K8s {
+        match runtime_config.deployment {
+            rendered::Deployment::K8s(k8s) => k8s,
+            rendered::Deployment::Host(_) => unreachable!("expected k8s deployment"),
```
sigilioso: Super nit: we could move this to the `rendered::Deployment` testing module:

**under #cfg(test)**:

```rust

impl Deployment {
    pub fn on_host(self) -> Option<OnHost> {
        match self {
            Self::Host(on_host) => Some(on_host),
            _ => None,
        }
    }
    
    pub fn k8s(self) -> Option<K8s> {
        match self {
            Self::K8s(k8s) => Some(k8s),
            _ => None,
        }
    }
}

// Then we can use as:

fn some_test() {
    // ...
    let k8s = self.runtime.deployment.k8s().unwrap();
}
```

agent-control/src/agent_type/runtime_config.rs
```diff
@@ -17,103 +15,33 @@ use super::error::AgentTypeError;
 use super::templates::Templateable;
 use k8s::K8s;
 use on_host::OnHost;
-use serde::Deserialize;
 
-/// Strict structure that describes how to start a given agent with all needed binaries, arguments, env, etc.
-#[derive(Debug, Deserialize, Clone, PartialEq)]
-#[serde(deny_unknown_fields)]
+/// Strict structure that describes how to start a given agent with all needed binaries,
+/// arguments, env, etc.
```
sigilioso, line 20: Did we stop _denying unknown fields_ for any particular reason?

agent-control/src/agent_type/runtime_config.rs
```diff
@@ -17,103 +15,33 @@ use super::error::AgentTypeError;
...
-            })
-        }
-    }
+/// Deployment of an agent type. Each variant carries the shape-specific config for that
+/// target.
+#[derive(Debug, Clone, PartialEq)]
+#[allow(clippy::large_enum_variant)]
+pub enum Deployment {
+    Host(OnHost),
+    K8s(K8s),
```
sigilioso, line 32: Considering that we already know the variant when wiring everything up (we know the environment), I was wondering if we could ditch the enum and move the variants to "compile time". I was thinking of something similar to:

```rust
pub struct Runtime<D> {
    deployment: D
}
```

We might need to add that generic to all types using it, but we would avoid the runtime check's which should actually be unreachable.

We would use `Runtime<OnHost>` for Linux and Windows and `Runtime<K8s>` for k8s.

The idea occurred to me while reviewing the `get_on_host_config` and the `get_k8s_config` in the effective_agents_assembler:

 https://github.com/newrelic/newrelic-agent-control/blob/538aa2b3b57ee085c59f2c5093ca188c97154434/agent-control/src/sub_agent/effective_agents_assembler.rs#L57-L77

sigilioso (commented):
Looks good! 👍 

The changes not only flatten the schema definition but also improve how we handle the environment. I've left some comments / suggestions. Let me know what you think.

agent-control/src/agent_type/definition/agent_type_validation_tests.rs
gsanchezgavier: yes , too much for this PR already

agent-control/src/agent_type/definition/agent_type_validation_tests.rs
gsanchezgavier: indeed i was iterating and this is leftover, the registry could be used directly

agent-control/src/agent_type/registry/embedded.rs
gsanchezgavier: sounds something good to explore! The whole registry refactor was something i was not expecting to have in this PR but it was needed, my approach was to keep it the simplest i could for this. I will write it down in the follow ups list if that makes sense to you  (as it shouldn't change the functionality i think)

agent-control/src/agent_type/agent_type_id.rs
gsanchezgavier: I iterate over a few options but not this one, and seems better than the current! let me have a try 😃 !

agent-control/src/agent_type/definition.rs
gsanchezgavier: IIRC , long back ago we didn't like the idea to mutate a unique AgentType component while we where enriching variable's and that was one of the reasons to have both, but i could be wrong and since the Environment is not there anymore this could be a valid thing to consider! i will written it down in the follow ups

agent-control/src/agent_type/definition.rs
gsanchezgavier: yes i think is there but in a spread way, make sense to have it unified 👍

agent-control/src/agent_type/runtime_config.rs
gsanchezgavier: yes this is not being deserialised anymore. I initially iterate over the removal of this middle struct but it was increasing the size of the PR , and i wasn't sure to remove this "runtime" concept even if i don't compleatly understood it.

Anyway i will doublecheck that no functionality have been change regarding denying unknown fields, as i think i haven't checked that thank!

agent-control/src/agent_type/runtime_config.rs
gsanchezgavier: yes totally make sense ! , i would first double check if this Runtime abstraction is needed as i mentioned here https://github.com/newrelic/newrelic-agent-control/pull/2572/changes#r3346660477. 

I will add it to the follow ups

agent-control/src/agent_type/definition/agent_type_validation_tests.rs
gsanchezgavier: [8dad138](https://github.com/newrelic/newrelic-agent-control/pull/2572/commits/8dad1381ec8f8a73b03a9729fecabc59080fa622)

gsanchezgavier: oh the hashSet is use to detect the missing definitions

agent-control/src/agent_type/agent_type_id.rs
gsanchezgavier: introduced the change [b60ab1a](https://github.com/newrelic/newrelic-agent-control/pull/2572/commits/b60ab1a799e97e74629931a74d3ab4d11fa5592e), much better like this thanks!

agent-control/src/agent_type/definition.rs
gsanchezgavier: done

agent-control/src/agent_type/runtime_config.rs
gsanchezgavier: I was looking at this , and yes we are some of this checks . Adding them would involve having to change the use of flattens and some enums , the code gets uglier and i'm not sure if this is really needed 🤔 , at the end we have an exhaustive checks on deserialisation to endup with strong typed structs , so i don't think there is much room for an typo kind of error that would silently go un-noticed , i think we are going to catch that when deserialising.

agent-control/src/agent_control/run.rs
```diff
@@ -125,6 +105,7 @@ impl AgentControlRunner {
         });
 
         let agent_type_registry = Arc::new(EmbeddedRegistry::new(
+            context.running_mode,
```
danielorihuela, line 108: Nit: We use `running_mode` in multiple places for different things (`RunningMode` and `Environment`). I wonder if we should rename it, in another PR.

agent-control/src/agent_type/definition.rs
```diff
@@ -5,67 +5,179 @@
...
+            (Platform::Host, Some(OperatingSystem::Linux)) => Environment::Linux,
+            (Platform::Host, Some(OperatingSystem::Windows)) => Environment::Windows,
+            (Platform::Kubernetes, None) => Environment::K8s,
+            (Platform::Host, None) => {
+                return Err(AgentTypeMetadataError::MissingOperatingSystem);
+            }
+            (Platform::Kubernetes, Some(_)) => {
+                return Err(AgentTypeMetadataError::UnexpectedOperatingSystem);
+            }
+        };
```
danielorihuela, line 142: Nit: This makes sense as a method in `Environment`. For example, `from_platform_and_os`.

agent-control/src/agent_type/README.md
```diff
@@ -243,50 +245,49 @@ In this configuration:
 
 The Agent Control leverages [Flux](https://fluxcd.io/) to act as an operator running Helm commands (install, upgrade, delete) as needed based on the provided configurations.
 
-Then, for a Kubernetes deployment, we use the following format:
+Then, for a Kubernetes deployment (`platform: kubernetes`), we use the following format:
```
danielorihuela, line 248: Maybe we need to rework the documentation a bit. The format now is the same. However, the documentation explanation makes me thing that it's different for onhost and k8s.

Maybe we need a separated task to work on that.

agent-control/src/agent_type/runtime_config.rs
danielorihuela: The only issue I see is that if the agent type author adds a field that AC doesn't understand yet, it will just ignore it.

Imagine we add `new_field` in version 3.0.0. The user modifies it's config but forgets to update AC (he is in 2.0.0). Then AC won't complain and everything will run. With the `deny`  AC would complain that it doesn't understand the new field.

I don't know which behaviour I prefer. I like the second more, but I don't think it's that relevant. Once metadata has more capabilities, this should be handled in the UI, right?

If we want to be super strict, then we need structs for our internal representation and structs for the data we are deserializing. I feel like it's too much work. I'm with @gsanchezgavier.

danielorihuela: For custom agent types could be worth it.

agent-control/tests/on_host/scenarios/filesystem_ops.rs
```diff
@@ -38,16 +38,13 @@ fn writes_filesystem_entries() {
 namespace: test
 name: test
 version: 0.0.0
+platform: host
+operating_system: {AGENT_CONTROL_MODE_ON_HOST}
```
danielorihuela, line 42: `OS_ENVIRONMENT` and `AGENT_CONTROL_MODE_ON_HOST` seem to be the same.

agent-control/src/agent_type/definition.rs
```diff
@@ -5,67 +5,179 @@
...
+            variables: self.variables.with_config(constraints),
+            runtime_config: self.runtime_config,
+        }
+    }
+}
+
+#[derive(Error, Debug, PartialEq)]
+pub enum AgentTypeMetadataError {
+    #[error("{0}")]
+    Id(AgentTypeIDError),
```
danielorihuela: I believe this variant is not in use.

agent-control/src/agent_type/README.md
gsanchezgavier: > documentation explanation makes me thing that it's different for onhost and k8s.

The difference has not changed, only how is the environment specified. Each deployment still have the differences explained in this doc

agent-control/src/agent_type/definition.rs
gsanchezgavier: yeap, i did it this way just to simplify the errors , and because is not used anywere else , and hope it is not in the future :)

agent-control/src/agent_type/README.md
danielorihuela: But the format is the same now, isn't it?

agent-control/src/agent_type/definition.rs
gsanchezgavier: true this is a leftover thanks!

agent-control/tests/on_host/scenarios/filesystem_ops.rs
gsanchezgavier: indeed! i just reuse them on the tests thanks

agent-control/src/agent_type/README.md
gsanchezgavier: no, you have a deployment for k8s and a deployment for onhost , so all those fields needs to be documented

agent-control/src/agent_type/runtime_config.rs
sigilioso: 👍 It makes sense to me as well, I just wanted us to double check it because the behavior slightly changes.

sigilioso (approved):
LGTM

danielorihuela (approved):

agent-control/src/agent_type/registry/embedded.rs
gsanchezgavier: I was just checking this and as you said , since there is no k8s/host features , introducing this filter will require some extra complexity which for the amount of AT we have i don't think is worth.

PR #2574, opened 2026-06-01
refactor: setup agent-type-registry skeleton [NR-484348]

sigilioso:
> [!WARNING]
> On top of #2572 

## Summary

This PR reshapes the agent type registry from a single `EmbeddedRegistry` into a layered structure that leaves room for an upcoming Remote Registry. 

The new top-level Registry is a thin facade that today delegates to a single LocalRegistry, which is split into two focused submodules: embedded (compile-time definitions) and custom (on-disk definitions). No runtime behavior changes: custom still overrides embedded, file order is still sorted, and invalid files are still skipped.

## Changes:

- Introduce the new `Registry` and `LocalRegistry` structs.
- Rename `DYNAMIC_AGENT_TYPE_DIR` to `CUSTOM_AGENT_TYPE_DIR`: dynamic will be misleading when remotes are supported. However, the folder name `dynamic-agent-types` was unchanged for backwards compatibility.
- Moved the sanity-checks for Agent Type Definitions from `definition` to `registry/local` (where the embedded repository is actually built).
- `.gitignore` changed to ignore `/local` (in the repo root) instead of any `local/` folder.


agent-control/src/agent_type/registry/local/custom.rs
```diff
@@ -0,0 +1,111 @@
...
+use tracing::{debug, error};
+
+/// Reads and returns the custom (dynamic) agent type definitions found in the given directory.
+///
+/// These are the agent types a user drops in the dynamic agent types directory to override or
+/// extend the embedded ones for PoCs and testing. If there is an error reading the directory or
+/// deserializing one of its files, the error is logged and the offending entry skipped.
+pub(super) fn custom_definitions(path: PathBuf) -> Vec<AgentTypeDefinition> {
+    let Ok(dir_entries) = fs::read_dir(path.clone()).inspect_err(|err| {
+        debug!(
```
vjripoll, line 12: Maybe we can change it to warning to avoid hiding possible errors?

agent-control/src/agent_type/registry/local.rs
```diff
@@ -0,0 +1,250 @@
...
+    pub fn new(custom_agent_types_path: PathBuf) -> Self {
+        let mut registry =
+            Self::try_new(embedded_definitions()).expect("Conflicting agent type definitions");
+
+        // Custom agent types are dynamic, hence they take precedence over the embedded ones.
+        for definition in custom_definitions(custom_agent_types_path) {
+            debug!("Storing custom agent type: {}", definition.agent_type_id);
+            registry
+                .definitions
+                .insert(definition.agent_type_id.clone(), definition);
```
vjripoll: should we use here the insert unique to avoid silent rewrite?

agent-control/src/agent_type/registry/local/custom.rs
sigilioso: Why changing it now? It preserves the previous behavior: a warning could cause noise because the default folder for 'custom Agent Types' will not exist unless it is created/mounted to load custom 'custom Agent Types'.

agent-control/src/agent_type/registry/local.rs
sigilioso: I don't think so overriding the embedded (or even other custom Agent Types) is the expected behavior. I don't see the use case for insert unique here. What would you use insert unique for?

agent-control/src/agent_type/registry/local/custom.rs
vjripoll: Make sense, 👌

agent-control/src/agent_type/registry/local.rs
vjripoll: If there are duplicate IDs, the second one would overwrite the first silently. Was that the intended behavior? I don't remember if we discussed that scenario. If not, then yeah, insert_unique makes sense here.

vjripoll (dismissed):
LGTM! appart of these doubts everything is ok for me :)

agent-control/src/agent_type/registry/local.rs
sigilioso: An example would be overriding the infra-agent using Custom Agent types:

- `newrelic/com.newrelic.infrastructure:0.1.0` exists in the embedded
-  If `newrelic/com.newrelic.infrastructure:0.1.0` is also defined in the 'Custom Agent Types' folder, we need it to take precedence. In order to do that, we overwrite the existing one.

The example in the [tests](https://github.com/newrelic/newrelic-agent-control/pull/2574/changes#diff-0050ac6f73b7b1d8c63cfa7a6769396b0809160d45dfda2fb6209f0c6a73956fR177) goes one step further and also consider two Agent Types with the same id in the 'Custom Agent Types' folder, in that case the precedence is dictated by: `'Custom (sorted by file)' > Embedded`.

sigilioso:
I'll switch the base to #2572 in order to make the conflict resolution easier

alvarocabanas (dismissed):
LGTM

agent-control/src/agent_type/registry.rs
```diff
@@ -26,10 +30,46 @@ pub trait AgentTypeRegistry {
...
+/// Holds the information to initialize a [Registry].
+pub struct RegistryConfig {
+    pub custom_agent_types_path: PathBuf,
+}
+
+/// The agent type registry used across Agent Control.
+///
+/// It resolves agent type definitions delegating to the internal implementations.
+#[derive(Debug)]
+pub struct Registry {
```
gsanchezgavier, line 44: nit: should't be AgentTypeRegistry ?

agent-control/src/agent_control/defaults.rs
```diff
@@ -88,7 +88,7 @@ pub const STORE_KEY_LOCAL_DATA_CONFIG: &StoreKey = "local_config";
 /// - **k8s**: Used as the data key within the OpAMP/fleet ConfigMap.
 pub const STORE_KEY_OPAMP_DATA_CONFIG: &StoreKey = "remote_config";
 pub const STORE_KEY_INSTANCE_ID: &StoreKey = "instance_id";
-pub const DYNAMIC_AGENT_TYPE_DIR: &str = "dynamic-agent-types";
+pub const CUSTOM_AGENT_TYPES_DIR: &str = "dynamic-agent-types"; // Used `dynamic-` instead of `custom-` for backwards compatibility
```
gsanchezgavier: i'm probably forgetting something but why are we changing this . perhaps it introduce more confusion if we keep an extrnal name and call it differently inside ?

gsanchezgavier (commented):
just have two minor questions but it looks good thanks!

agent-control/src/agent_control/defaults.rs
sigilioso: I can roll it back with no issue, I was concerned about the "dynamic" term. After all remote will also be dynamically loaded.

agent-control/src/agent_type/registry.rs
sigilioso: I thought of it as `agent_type::Registry` and I used the `AgentTypeRegistry` name for the trait, but I can try to look for better names.

gsanchezgavier: oh the trait... yes makes sense then , i though you were thinking in Go again 😆

agent-control/src/agent_control/defaults.rs
gsanchezgavier: I think we have some options here:
**backward
- a)current change
**break
- pick a new name and clean up
  - b)use custom
  - c)perhaps just use something generic like `agent_types`
- Other options for sure

Since we are already introducing a breaking change for this dev/testing feature (custom agent types) I would go with b/c and clean the house

sigilioso: After discussing it _offline_ we agreed keeping the current name `dynamic-agent-types` and also keep `DYNAMIC_` as constant / variables name. I've updated it here: [03e5336](https://github.com/newrelic/newrelic-agent-control/pull/2574/commits/03e53363311cfddd8b2daf829e812ef116c5fc46)

gsanchezgavier (approved):
thanks!

PR #2577, opened 2026-06-02
feat: add agent-type registry remote config [NR-568046]

sigilioso:
## Summary

Introduces the `agent_types` block in the Agent Control local configuration, defining where agent type definitions are pulled from and how they are verified. This is the **configuration definition** step for remote agent types: it parses and exposes the config, reusing the existing global OCI mirror. It does **not** implement any remote-pull logic (that lands in a follow-up).

## Config shape

```yaml
agent_types:
  default_remote:
    repository: newrelic/agent-control-agent-types
    signature_verification_enabled: true
    public_key_url: https://publickeys.newrelic.com/g/agent-control-oci/global/agent-type/jwks.json
```

The `default_remote` block **reuses the global `oci` mirror** (`oci.registry` + `oci.auth`) rather than re-declaring a registry/auth. A customer already mirroring packages needs no extra config for agent types.
- It only adds the agent-type-specific bits: the repository, the signature toggle, and the signing key. 
- The signature fields follow the existing package convention (`signature_verification_enabled` + `public_key_url`).

All fields default, so omitting the block (or any sub-field) yields the New Relic defaults.

## Notes for reviewers
- Renamed the `AC_OCI_PACKAGE_DEFAULT_REGISTRY` constant to `AC_OCI_DEFAULT_REGISTRY`, since it backs the shared default registry host used by packages, self-update, and now agent types.
- Keep in mind the **follow-ups** in the next section.

### `custom_remotes` can be added without a breaking change

The shape is intentionally nested under `default_remote` so a future `custom_remotes` list (for customer/custom-namespace agent types) can be added as a **purely additive** `#[serde(default)]` sibling. No existing field is renamed, retyped, or made required, and the struct does not use `deny_unknown_fields`, so:
- existing configs keep parsing unchanged, and
- a newer config that already includes `custom_remotes` is silently tolerated by an older binary (forward-compatible in both directions).

A future config **could** look like:

```yaml
agent_types:
  default_remote:
    repository: newrelic/agent-control-agent-types
    signature_verification_enabled: true
    public_key_url: https://publickeys.newrelic.com/g/agent-control-oci/global/agent-type/jwks.json
  # NEW (additive, non-breaking): per-namespace custom sources
  custom_remotes:
    - namespaces: [fake_company]
      repository: agent-control-agent-types
      signature_verification_enabled: true
      public_key_url: https://my-company.example/keys/jwks.json
```

A test (`test_agent_types_tolerates_unknown_future_fields`) already locks in this guarantee by parsing a config that includes an unknown `custom_remotes` key.

### Per-registry override left out of scope (deliberately)

The `default_remote` block does **not** expose a `registry` field to override the OCI host per agent-type source. This was a conscious decision rather than an omission:

- **Consistency with packages:** there is currently no per-package registry override either — the registry host is global (`oci.registry`). Adding a per-source override only for agent types would diverge from how packages behave today.
- **Not a breaking change to add later:** introducing a per-registry override would itself be additive (a new optional field), so nothing here forecloses it.

> [!NOTE]
> If/when we want per-source registry overrides, we should design it **uniformly across all OCI usages** (packages *and* agent types) rather than bolting it onto agent types alone.

### This is only the config definition

- The implementation of the OCI registry will be addressed separately.
- Docs exposing the new configuration will be addes when the implementation is ready.

### DIFF
--- a/agent-control/src/agent_control/config.rs
+++ b/agent-control/src/agent_control/config.rs
@@ -89,7 +89,7 @@ pub struct AgentControlConfig {
     /// Configuration for the agent type registry remote source.
     /// Reuses the global `oci` registry/auth; see [AgentTypeRegistryConfig].
     #[serde(default)]
-    pub agent_types: AgentTypeRegistryConfig,
+    pub agent_types: AgentTypeConfig,
 }
 
 #[derive(Debug, Default, Deserialize, PartialEq, Clone)]
@@ -210,7 +210,7 @@ pub struct SignatureVerificationEnabled(bool);
 /// Configuration for the agent type registry's remote source(s).
 #[derive(Debug, Default, Deserialize, PartialEq, Clone)]
 #[serde(default)]
-pub struct AgentTypeRegistryConfig {
+pub struct AgentTypeConfig {
     /// The default (New Relic) remote source for agent types.
     pub default_remote: DefaultAgentTypeRemote,
 }
@@ -1383,7 +1383,7 @@ oci:
 
     #[test]
     fn test_agent_types_default_when_block_or_subfields_omitted() {
-        let expected = AgentTypeRegistryConfig::default();
+        let expected = AgentTypeConfig::default();
         for config_input in [
             "agents: {}",                                     // agent_types omitted entirely
             "agents: {}\nagent_types: {}",                    // empty agent_types block
--- a/agent-control/src/agent_control/config.rs
+++ b/agent-control/src/agent_control/config.rs
@@ -2,8 +2,8 @@ use super::agent_id::AgentID;
 use super::http_server::config::ServerConfig;
 use super::uptime_report::UptimeReportConfig;
 use crate::agent_control::defaults::{
-    AC_OCI_PACKAGE_DEFAULT_REGISTRY, AC_OCI_PACKAGE_DEFAULT_REPOSITORY,
-    AC_OCI_PACKAGE_PUBLIC_KEY_URL,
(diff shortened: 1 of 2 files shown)


agent-control/src/agent_control/config.rs
```diff
@@ -202,6 +207,42 @@ const DEFAULT_SIGNATURE_VERIFICATION_ENABLED: bool = true;
 #[wrapper_default_value(DEFAULT_SIGNATURE_VERIFICATION_ENABLED)]
 pub struct SignatureVerificationEnabled(bool);
 
+/// Configuration for the agent type registry's remote source(s).
+#[derive(Debug, Default, Deserialize, PartialEq, Clone)]
+#[serde(default)]
+pub struct AgentTypeRegistryConfig {
```
gsanchezgavier: nit: i think reflects better the key name which is open to configs related to agent types . 
```suggestion
pub struct AgentTypeConfig {
```

gsanchezgavier (dismissed):
Thanks! so my understanding is that this `default` remote will be use to resolve any namespace ,  and the if we add the `custom` that would take precedence over this . Is that correct?

sigilioso:
> Thanks! so my understanding is that this `default` remote will be use to resolve any namespace , and the if we add the `custom` that would take precedence over this . Is that correct?

Exactly! That's the idea.

agent-control/src/agent_control/config.rs
sigilioso: Renamed here: [4c4dde8](https://github.com/newrelic/newrelic-agent-control/pull/2577/commits/4c4dde80546bcaf24fcdc6d6258a1df83491f012)

gsanchezgavier (approved):
thanks!

PR #2580, opened 2026-06-03
feat: agent-type OCI downloader [NR-568047]

sigilioso:
## Summary

First step to implement support for remote registries for Agent Types. Adds the ability to fetch agent-type definitions from an OCI registry directly into memory, with optional Cosign signature verification. Keeping the artifact in memory (instead of writing to disk) means resolution does not depend on a writable filesystem location, which matters on Kubernetes (read-only root fs, ephemeral storage).

## Notes for reviewers

- The `ManifestArtifactType::AgentType` / `LayerMediaType::AgentType` constants already existed; this PR adds the consumer-side validation/extraction.
- When signature verification is enabled but no `public_key_url` is configured, the download proceeds unverified (logs a warning). This is intentional and matches the existing package downloader behavior; a test now locks in that fail-open path.
- The **in-memory** pull_blob adds a cap size to prevent downloading too big Agent Type definitions (mis-configured or malicious).

## Out of scope

- The new code isn't wire-up yet, this will be addressed in a follow-up PR, when introducing the the remote registry implementation.


sigilioso:
I've just converted it to draft because we need to consider the k8s use case and maybe relying on the filesystem is not the best approach.

sigilioso:
Now it is updated and ready to review.

agent-control/src/agent_type/oci/downloader.rs
```diff
@@ -0,0 +1,331 @@
...
+        public_key_url: &Url,
+    ) -> Result<Reference, OCIAgentTypeDownloaderError> {
+        self.client
+            .verify_signature(reference, public_key_url, &self.auth)
+            .map_err(|err| OCIAgentTypeDownloaderError(err.to_string()))
+    }
+
+    /// Pulls the manifest, validates it is an agent type artifact, pulls its single layer into
+    /// memory and returns the agent type definition it contains.
+    fn download_definition(
```
danielorihuela, line 143: I see this downloader is the equivalent of the `OCIArtifactDownloader`. The only difference is that we use `pull_blob` instead of `pull_blob_to_file`.

However, how you thought about doing a refactor after that PR or somewhere in the future?

The agent type is just another `Oci`.

```rust
pub struct Oci {
    pub repository: Repository,
    pub version: Version,
    pub public_key_url: Option<Url>,
}
```

We could have another abstraction. Imagine the following.

`OCIAgentTypeDownloader` -> `OciArtifactDownloader` -> `oci::Client`
`OCIAgentDownloader`         -> `OciArtifactDownloader` -> `oci::Client`

The lower layer implements communication with OHI.
The second layer implements all high level workflows.
The top layer is just a wrapper to make our life easier.

We could even do a separated crate for the `OciArtifactDownloader` and have the wrappers in AC. Just to be clear. I think this is all too much and not worth it, having two downloaders is easy enough.

danielorihuela: I changed my opinion, I think it's worth it if we have time.
In that extra abstraction we can test all the attacks, `toctou` and `mitm` and be sure that no downloader has vulnerabilities now or in the future.

agent-control/src/agent_type/oci/downloader.rs
```diff
@@ -0,0 +1,331 @@
...
+        ]);
+        let tar_gz = tar_gz_with_definition(&format!("{TAG}.yaml"), DEFINITION);
+        let server = FakeOciServer::new(REPOSITORY, TAG)
+            .with_artifact_type(&ManifestArtifactType::AgentType.to_string())
+            .with_layer(&tar_gz, &LayerMediaType::AgentType.to_string())
+            .with_signature(&key_pair)
+            .build();
+
+        let downloader =
+            create_downloader(server.registry(), true, Some(jwks_server.url), 1024 * 1024);
```
danielorihuela: Nit: Constant shared across tests for magic number

agent-control/src/agent_type/oci/downloader.rs
```diff
@@ -0,0 +1,331 @@
...
+    /// Downloads the agent type artifact at `<registry>/<repository>:<tag>`.
+    ///
+    /// If signature verification is enabled and a `public_key_url` is configured, it first verifies
+    /// the artifact's signature and then downloads the artifact that was verified (identified by
+    /// `digest`, to assure the artifact downloaded is the one verified).
+    ///
+    /// On failure the operation is retried as configured; if all attempts are exhausted it returns
+    /// an error.
+    fn download(&self, tag: &str) -> Result<Vec<u8>, OCIAgentTypeDownloaderError> {
+        retry(self.max_retries, self.retry_interval, || {
```
danielorihuela, line 68: Nit: A `debug` indicating here the repository and tag might help detect if we are using a mirror and which version we are trying to use.

agent-control/src/agent_type/oci/downloader.rs
```diff
@@ -0,0 +1,331 @@
...
+            signature_verification_enabled,
+            public_key_url,
+            max_retries: DEFAULT_RETRIES,
+            retry_interval: Duration::default(),
+            max_size_bytes,
+        }
+    }
+
+    /// Returns a new downloader with the provided retry configuration.
+    pub fn with_retries(self, retries: usize, retry_interval: Duration) -> Self {
```
gsanchezgavier, line 108: dead code? if we are not setting retries i would remove all this

agent-control/src/agent_type/oci/downloader.rs
```diff
@@ -0,0 +1,331 @@
...
+    fn should_verify_signature(&self) -> Option<&Url> {
+        if !self.signature_verification_enabled {
+            warn!("Signature verification is disabled, skipping");
+            return None;
+        }
+        if self.public_key_url.is_none() {
+            warn!(
+                "No public_key_url configured for agent type remote, skipping signature verification"
+            );
+        }
```
gsanchezgavier, line 125: shouldn't this fail instead? sounds like this should be required if verification is enabled

agent-control/src/agent_type/oci/downloader.rs
```diff
@@ -0,0 +1,331 @@
...
+    mock! {
+        pub OCIAgentTypeDownloader {}
+        impl OCIAgentTypeDownloader for OCIAgentTypeDownloader {
+            fn download(&self, tag: &str) -> Result<Vec<u8>, OCIAgentTypeDownloaderError>;
+        }
+    }
+
+    const DEFINITION: &[u8] = b"namespace: newrelic\nname: com.newrelic.infrastructure\n";
+    const REPOSITORY: &str = "newrelic/agent-control-agent-types";
+    const TAG: &str = "host-linux-com.newrelic.infrastructure-0.1.0";
```
gsanchezgavier: nit if possible i would use more transparent fake data like `fake-name` or `fake agent type content` instead of this. just to save time/tokens in future refactors

agent-control/src/oci/artifact_definitions.rs
```diff
@@ -156,6 +160,92 @@ impl LocalAgentPackage {
...
+        }
+        Ok(layer.clone())
+    }
+
+    /// Decompresses the gzipped tar held in memory and returns the content of the single Agent
+    /// Type definition file it contains.
+    ///
+    /// It fails if the artifact does not contain exactly one file. Deserializing the returned bytes
+    /// into an agent type definition is left to the caller.
+    pub fn extract_definition(&self) -> Result<Vec<u8>, DefinitionError> {
```
gsanchezgavier: extract_definition(self), would make sense?

agent-control/src/agent_type/oci/downloader.rs
```diff
@@ -0,0 +1,331 @@
...
+impl OCIAgentTypeArtifactDownloader {
+    /// Returns a downloader with default (no) retries.
+    pub fn new(
+        client: Client,
+        registry: Registry,
+        repository: Repository,
+        auth: Option<OciAuth>,
+        signature_verification_enabled: bool,
+        public_key_url: Option<Url>,
+        max_size_bytes: usize,
```
gsanchezgavier: would make sense just to hardcode a really high value that would allow any agent type and would prevent memory exaustion at the same time , something like 10MB ?

gsanchezgavier (commented):
Nice! i just left some minor questions and suggestions

agent-control/src/agent_type/oci/downloader.rs
sigilioso: I actually started by extending the existing `OciArtifactDownloader` instead of adding a new one, but it de-coupling it from packages was making things a bit more complex. Do you think a follow-up would be OK?

We could consider either extending the current `oci::Client` wrapper or having an extra abastraction.

agent-control/src/agent_type/oci/downloader.rs
sigilioso: The downloader is not built yet (not in the production code). Don't you think we should set up retries when wiring it up?

agent-control/src/agent_type/oci/downloader.rs
danielorihuela: Yes. I like it in another PR.
Yes, "use cases" could be implemented into the `oci::Client`.
I just throw out the idea. Do a task or whatever you think it's right.

agent-control/src/agent_type/oci/downloader.rs
gsanchezgavier: > The downloader is not built yet (not in the production code)

oh missed that :) , yes it make sense to have some retries , i just thought we where adding them but not setting them when saw that the default value was 0

agent-control/src/agent_type/oci/downloader.rs
sigilioso: I'd like to address it before closing the current story. I'm linking this comment to the task.

agent-control/src/oci/artifact_definitions.rs
sigilioso: I think it makes sense: the performance is the same (we are using the `self.blob.as_slice()` to decompress, but, by consuming we force the caller to drop the `ArtifactDefinition` when it is extracted 👍

agent-control/src/agent_type/oci/downloader.rs
sigilioso: I removed the magic number because now it is a constant instead of a paramater: [4af19af](https://github.com/newrelic/newrelic-agent-control/pull/2580/commits/4af19af11d8750ead1aad41123eb8e27ad55cd6a)

agent-control/src/agent_type/oci/downloader.rs
sigilioso: Added here: [4af19af](https://github.com/newrelic/newrelic-agent-control/pull/2580/commits/4af19af11d8750ead1aad41123eb8e27ad55cd6a)

agent-control/src/agent_type/oci/downloader.rs
danielorihuela: If we change it here, I would change it for packages too. For consistency.
However, IIRC, we did it that way for packages to avoid breaking running agents for clients.

sigilioso: Instead of failing here I have' just dropped the boolean, we can perform validation when wiring things up and make the url `None` if signature verification is disabled. I copied the approach from the packages and the settings are slightly different in this case: [4af19af](https://github.com/newrelic/newrelic-agent-control/pull/2580/commits/4af19af11d8750ead1aad41123eb8e27ad55cd6a)

agent-control/src/agent_type/oci/downloader.rs
sigilioso: Changed here: [4af19af](https://github.com/newrelic/newrelic-agent-control/pull/2580/commits/4af19af11d8750ead1aad41123eb8e27ad55cd6a)

agent-control/src/oci/artifact_definitions.rs
sigilioso: Changed here: [4af19af](https://github.com/newrelic/newrelic-agent-control/pull/2580/commits/4af19af11d8750ead1aad41123eb8e27ad55cd6a)

agent-control/src/agent_type/oci/downloader.rs
sigilioso: Hardcoded as a constant here: [4af19af](https://github.com/newrelic/newrelic-agent-control/pull/2580/commits/4af19af11d8750ead1aad41123eb8e27ad55cd6a). Maybe 10MB are too much...

danielorihuela (approved):

gsanchezgavier (approved):
🚀

PR #2588, opened 2026-06-09
feat: add new filesystem format with ephemeral and persistent folders

alvarocabanas:
Introduces ephemeral/persistent lifecycle semantics for sub-agent filesystem directories.
                                                                                                                                                                                                                                   
Replaces the flat FileSystem map with two explicit sections — ephemeral and persistent — allowing agent type authors to declare the intended lifecycle of each directory in the agent type definition.                         

- Ephemeral directories are deleted when the agent is stopped, removed, or restarted. They are cleared and repopulated on every config update, so files removed from dynamic variables (e.g. config_integrations in com.newrelic.infrastructure) are  automatically deleted from disk without any extra tracking.
- Persistent directories are only deleted when the agent is removed. They are written additively and survive agent stops,  restarts, and config updates, intended for directories owned by the sub-agent process itself (e.g. newrelic-infra/newrelic-integrations/logging).

On every write() call, a startup cleanup phase compares what's on disk against the current agent type configuration and removes stale directories that are no longer in the config. Persistent directories that are still in the config are preserved, while ephemeral directories are always cleared and recreated.


## Checklist
 -->

### DIFF
--- a/agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure-0.1.0.yaml
+++ b/agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure-0.1.0.yaml
@@ -69,11 +69,13 @@ deployment:
       - --version
     regex: \d+\.\d+\.\d+
   filesystem:
-    config:
-      newrelic-infra.yaml: |
-        ${nr-var:config_agent}
-    integrations.d: ${nr-var:config_integrations}
-    logging.d: ${nr-var:config_logging}
+    persistent:
+      config:
+        newrelic-infra.yaml: |
+          ${nr-var:config_agent}
+      logging.d: ${nr-var:config_logging}
+    ephemeral:
+      integrations.d: ${nr-var:config_integrations}
   executables:
     - id: newrelic-infra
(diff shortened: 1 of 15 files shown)


agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure-0.1.0.yaml
```diff
@@ -69,11 +69,13 @@ deployment:
       - --version
     regex: \d+\.\d+\.\d+
   filesystem:
-    config:
-      newrelic-infra.yaml: |
-        ${nr-var:config_agent}
-    integrations.d: ${nr-var:config_integrations}
-    logging.d: ${nr-var:config_logging}
+    persistent:
+      config:
```
paologallinaharbur, line 74: why the configs are under persistent?

alvarocabanas: I assumed it needs to keep there unless the agent is removed, being persistent doesn't mean it won't be overwriten, but I assume there is always a newrelic-infra.yaml if there is an infra agent.

PR #2589, opened 2026-06-09
feat: add Agent type definition `protocol_version` (NR-568042)

gsanchezgavier:
## Summary
Introduces a schema **protocol version** for agent-type definitions, decoupled from both the agent-type `version` (semver) and the Agent Control release version. 

Every agent-type file **must** now declare a quoted `protocol_version: "MAJOR.MINOR"`, which is validated at the registry ingestion boundary before the rest of the file is parsed.

Agent Control declares a single `SUPPORTED_PROTOCOL_VERSION` (the max it understands). 

## Changes
- New `agent_type/protocol_version.rs`: `ProtocolVersion` parsing + compatibility check.
- `parse_agent_type_definition` validates `protocol_version` first; an incompatible/missing/malformed version is rejected with a clear error.
- All embedded agent-type YAMLs declare `protocol_version: "0.1"`.

## Notes
- The check runs for both embedded and dynamic (custom) agent types.
- Bumping `SUPPORTED_PROTOCOL_VERSION` is the single knob for widening what AC accepts.

### DIFF
--- a/agent-control/src/agent_type/definition.rs
+++ b/agent-control/src/agent_type/definition.rs
@@ -43,8 +43,8 @@ pub struct AgentTypeDefinition {
     pub runtime_config: Runtime,
 }
 
-/// Wraps [AgentTypeDefinition] to force deserialization using [parse_agent_type_definition] which
-/// verifies the `protocol_version`.
+/// Wraps [AgentTypeDefinition] to force deserialization through [AgentTypeDefinition::from_slice],
+/// which verifies the `protocol_version`.
 struct RawAgentTypeDefinition(AgentTypeDefinition);
 
 impl<'de> Deserialize<'de> for RawAgentTypeDefinition {
@@ -84,17 +84,6 @@ impl<'de> Deserialize<'de> for RawAgentTypeDefinition {
     }
 }
 
-/// Parses raw yaml content into an [AgentTypeDefinition], validating the file's `protocol_version`
-/// before the document is converted into the definition.
-pub fn parse_agent_type_definition(
-    content: &[u8],
-) -> Result<AgentTypeDefinition, AgentTypeDefinitionParseError> {
-    let document: serde_json::Value = serde_saphyr::from_slice(content)?;
-    protocol_version::check(&document)?;
-    let definition = serde_json::from_value::<RawAgentTypeDefinition>(document)?.0;
-    Ok(definition)
-}
-
 #[derive(Error, Debug)]
 pub enum AgentTypeDefinitionParseError {
     #[error("incompatible protocol version: {0}")]
@@ -106,6 +95,15 @@ pub enum AgentTypeDefinitionParseError {
 }
 
 impl AgentTypeDefinition {
+    /// Parses raw yaml content into an [AgentTypeDefinition], validating the file's
+    /// `protocol_version` before the document is converted into the definition.
+    pub fn from_slice(content: &[u8]) -> Result<Self, AgentTypeDefinitionParseError> {
+        let document: serde_json::Value = serde_saphyr::from_slice(content)?;
+        protocol_version::check(&document)?;
+        let definition = serde_json::from_value::<RawAgentTypeDefinition>(document)?.0;
+        Ok(definition)
+    }
+
     pub fn agent_type_id(&self) -> &AgentTypeID {
         &self.metadata.id
     }
@@ -549,7 +547,7 @@ deployment:
             "protocol_version: \"{SUPPORTED_PROTOCOL_VERSION}\""
         ));
 
-        let definition = parse_agent_type_definition(yaml.as_bytes()).unwrap();
+        let definition = AgentTypeDefinition::from_slice(yaml.as_bytes()).unwrap();
 
         assert_eq!(definition.agent_type_id().namespace(), "ns");
         assert_eq!(definition.agent_type_id().name(), "test");
@@ -571,7 +569,7 @@ deployment: {{}}
 "#
         );
 
-        let definition = parse_agent_type_definition(yaml.as_bytes()).unwrap();
+        let definition = AgentTypeDefinition::from_slice(yaml.as_bytes()).unwrap();
 
         assert_eq!(definition.metadata.environment, Environment::Linux);
         assert_matches!(definition.runtime_config.deployment, Deployment::Host(_));
@@ -582,7 +580,7 @@ deployment: {{}}
         let yaml = k8s_definition_yaml("");
 
         assert_matches!(
-            parse_agent_type_definition(yaml.as_bytes()),
+            AgentTypeDefinition::from_slice(yaml.as_bytes()),
             Err(AgentTypeDefinitionParseError::ProtocolVersion(
                 ProtocolVersionError::Missing
             ))
@@ -595,7 +593,7 @@ deployment: {{}}
         let yaml = k8s_definition_yaml("protocol_version: \"99.0\"");
 
         assert_matches!(
-            parse_agent_type_definition(yaml.as_bytes()),
+            AgentTypeDefinition::from_slice(yaml.as_bytes()),
             Err(AgentTypeDefinitionParseError::ProtocolVersion(
                 ProtocolVersionError::IncompatibleMajor { .. }
             ))
@@ -608,7 +606,7 @@ deployment: {{}}
         let yaml = k8s_definition_yaml(&format!("protocol_version: {SUPPORTED_PROTOCOL_VERSION}"));
 
         assert_matches!(
-            parse_agent_type_definition(yaml.as_bytes()),
+            AgentTypeDefinition::from_slice(yaml.as_bytes()),
             Err(AgentTypeDefinitionParseError::ProtocolVersion(
                 ProtocolVersionError::InvalidFormat(_)
             ))
@@ -617,7 +615,7 @@ deployment: {{}}
 
     #[test]
     fn parse_rejects_malformed_yaml() {
-        let err = parse_agent_type_definition(b"name: [unclosed").unwrap_err();
+        let err = AgentTypeDefinition::from_slice(b"name: [unclosed").unwrap_err();
 
         assert_matches!(err, AgentTypeDefinitionParseError::Yaml(_));
     }
@@ -638,7 +636,7 @@ platform: host
         );
 
         assert_matches!(
-            parse_agent_type_definition(yaml.as_bytes()),
+            AgentTypeDefinition::from_slice(yaml.as_bytes()),
             Err(AgentTypeDefinitionParseError::Definition(_))
         );
     }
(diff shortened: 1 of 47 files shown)


agent-control/src/agent_type/protocol_version.rs
```diff
@@ -0,0 +1,196 @@
...
+//!
+//! For example, an Agent Control supporting `1.6` accepts `1.0`..=`1.6`, rejects `1.7` (too new),
+//! and rejects `0.9` and `2.0` (wrong major).
+
+use std::fmt::{self, Display};
+use std::str::FromStr;
+use thiserror::Error;
+
+/// Maximum protocol version this Agent Control understands.
+pub const SUPPORTED_PROTOCOL_VERSION: ProtocolVersion = ProtocolVersion { major: 0, minor: 1 };
```
sigilioso: Have you considered defining this in `Cargo.toml`. Something similar to:

```toml
[package.metadata]
# Max agent-type schema protocol version this Agent Control understands (MAJOR.MINOR).
agent_type_protocol_version = "0.1"
```

It would require some boilerplate to load it and keep the constant (we would need some `build.rs` magic ✨), but it could be worth it. What do you think?

agent-control/src/agent_type/protocol_version.rs
```diff
@@ -0,0 +1,196 @@
...
+        })
+    }
+}
+
+impl ProtocolVersion {
+    /// Checks this version against [SUPPORTED_PROTOCOL_VERSION].
+    /// A different major is a breaking change in either direction; a higher minor under the same major
+    /// is too new for this Agent Control. Equal or lower minors under the same major are additive and
+    /// backward-compatible.
+    pub fn is_compatible_with_supported(&self) -> Result<(), ProtocolVersionError> {
```
sigilioso:
```suggestion
    pub fn is_supported(&self) -> Result<(), ProtocolVersionError> {
```

I find it less confusing than `is_compatible_with_supported`

agent-control/src/agent_type/protocol_version.rs
```diff
@@ -0,0 +1,196 @@
...
+    fn check_reports_missing_field() {
+        assert_eq!(
+            check(&yaml_value("name: whatever")),
+            Err(ProtocolVersionError::Missing)
+        );
+    }
+
+    #[test]
+    fn check_rejects_unquoted_float() {
+        // An unquoted `0.1` is a yaml float, not the required quoted MAJOR.MINOR string.
```
sigilioso, line 177: Could make sense to also support the float representation or does it come with some downsides? Forgetting the quotes is a really easy mistake (although it is also easy to detect)

agent-control/src/agent_type/protocol_version.rs
```diff
@@ -0,0 +1,196 @@
...
+        file: ProtocolVersion,
+        supported: ProtocolVersion,
+    },
+    #[error(
+        "unsupported protocol_version {file}: newer than supported (this agent control supports up to {supported})"
+    )]
+    TooNew {
+        file: ProtocolVersion,
+        supported: ProtocolVersion,
+    },
```
sigilioso, line 49: Would the name `target` be clearer than `file`?

agent-control/src/agent_type/definition.rs
```diff
@@ -42,42 +43,68 @@ pub struct AgentTypeDefinition {
     pub runtime_config: Runtime,
 }
 
-impl<'de> Deserialize<'de> for AgentTypeDefinition {
+/// Wraps [AgentTypeDefinition] to force deserialization using [parse_agent_type_definition] which
+/// verifies the `protocol_version`.
+struct RawAgentTypeDefinition(AgentTypeDefinition);
```
sigilioso: Because this type is private we have no other alternative to use `parse_agent_type_definition`. Nice! 💯

danielorihuela, line 48: I don't understand why this wrapper is needed.

The custom deserializer behaviour remains the same. We just modified the name of some variables. `parse_agent_type_definition` is not called inside the deserializer. It's a function that we call when we need. How does this wrapper enforce the usage of `parse_agent_type_definition` then?

agent-control/src/agent_type/definition.rs
```diff
@@ -42,42 +43,68 @@ pub struct AgentTypeDefinition {
...
+/// Parses raw yaml content into an [AgentTypeDefinition], validating the file's `protocol_version`
+/// before the document is converted into the definition.
+pub fn parse_agent_type_definition(
+    content: &[u8],
+) -> Result<AgentTypeDefinition, AgentTypeDefinitionParseError> {
+    let document: serde_json::Value = serde_saphyr::from_slice(content)?;
+    protocol_version::check(&document)?;
+    let definition = serde_json::from_value::<RawAgentTypeDefinition>(document)?.0;
+    Ok(definition)
+}
```
sigilioso: Nit: we could move it to `AgentTypeDefinition`, so it mimics how serde loads from `&[u8]`:

```rust
// We could do this
let agent_type_def = AgentTypeDefinition::from_slice(some_bytes)?;
// instead of
let agent_type_def = parse_agent_type_definition(some_bytes)?;
```

sigilioso (commented):
I like the approach, thanks!

I've left some _nits_  add suggestions.

agent-control/src/agent_type/protocol_version.rs
```diff
@@ -0,0 +1,196 @@
+//! Versioning for the agent-type *schema language* itself, decoupled from both the agent type
+//! `version` (semver) and the Agent Control release version.
+//!
+//! Agent Control declares a single [SUPPORTED_PROTOCOL_VERSION] (the maximum it understands). Every
+//! agent type file must declare a `protocol_version` (a quoted `MAJOR.MINOR` string), which is
```
danielorihuela, line 5: Curious: Why only major and minor? The CDD mentions `MAJOR.MINOR.PATCH`

agent-control/src/agent_type/protocol_version.rs
gsanchezgavier: this was something that i thought it could be definitely improved, but kind of postponed until we consume this externally so I haven't though much about it... but the proposal look great ! i will try to make it work thanks!

agent-control/src/agent_type/protocol_version.rs
gsanchezgavier: thought about the same , but seemed too much for me, given that most probably no one will start these definitions from the scratch. 

> (although it is also easy to detect)

exactly, but let me know if you think it worth i can implement it

agent-control/src/agent_type/definition.rs
gsanchezgavier: yes the other clean alternative was to do this in the AgentTypeDefinition deserialization but that would have been a real pain for unit tests, or required some kind of hack

agent-control/src/agent_type/protocol_version.rs
gsanchezgavier: you might be confused with `version` metadata. Look at the `Agent Type compatibility` section, there it explains this and have a reference to the original doc

agent-control/src/agent_type/definition.rs
danielorihuela: Got it in https://github.com/newrelic/newrelic-agent-control/pull/2589#discussion_r3381865905.

gsanchezgavier: The idea is that you cannot deserialize the agent type definition without checking the protocol version in production code. So AgentTypeDefinition remains pub, but without deser, and this one is private with deser... and the parsing only can happen through `parse_agent_type_definition` which checks the protocol

agent-control/src/agent_type/protocol_version.rs
sigilioso: There is no need to support it (unless it is identified as a pain point in the future). The validator (or schema) should catch these errors easily.

agent-control/src/agent_type/protocol_version.rs
gsanchezgavier: done thanks!

agent-control/src/agent_type/protocol_version.rs
gsanchezgavier: done

agent-control/src/agent_type/registry/local/embedded.rs
sigilioso, line 1: Would it make sense extend the tests for embedded agent types to check their protocol version? It is a silly sanity check, maybe it is too much.

sigilioso (dismissed):
LGTM 👍 

I've left a comment regarding the possibility of extending the tests for embedded agents, but it can perfectly be ignored or addressed later.

agent-control/src/agent_type/registry/local/embedded.rs
gsanchezgavier: well i might be missing something but this should be already being checked as is using from_slice and that checks the protocol_version

gsanchezgavier: yes i just did a quick test and look
```

thread 'agent_type::registry::local::tests::test_default_loads_embedded_definitions' (3250613) panicked at agent-control/src/agent_type/registry/local/embedded.rs:24:18:
Invalid or incompatible embedded agent type: ProtocolVersion(TooNew { target: ProtocolVersion { major: 1, minor: 2 }, supported: ProtocolVersion { major: 1, minor: 0 } })

---- agent_type::registry::local::tests::test_new_overlays_custom_over_embedded_and_earlier_custom stdout ----

thread 'agent_type::registry::local::tests::test_new_overlays_custom_over_embedded_and_earlier_custom' (3250616) panicked at agent-control/src/agent_type/registry/local/embedded.rs:24:18:
Invalid or incompatible embedded agent type: ProtocolVersion(TooNew { target: ProtocolVersion { major: 1, minor: 2 }, supported: ProtocolVersion { major: 1, minor: 0 } })


failures:
    agent_type::registry::local::agent_type_validation_tests::all_agent_type_definitions_are_present
    agent_type::registry::local::agent_type_validation_tests::all_agent_type_definitions_are_resilient_k8s
    agent_type::registry::local::agent_type_validation_tests::all_agent_type_definitions_are_resilient_linux
    agent_type::registry::local::agent_type_validation_tests::all_agent_type_definitions_are_resilient_windows
    agent_type::registry::local::embedded::tests::embedded_definitions_count_per_environment
    agent_type::registry::local::tests::test_default_loads_embedded_definitions
```
It fails all over the place

sigilioso: I was the one missing something, thanks for clarifying 🚀

sigilioso (approved):
As long as we keep an eye on the APM test to move it back to `*` when ready, we can merge this as it is, right?

PR #2590, opened 2026-06-09
feat: introduce remote registry for Agent Types [NR-568047]

sigilioso:
## Summary

Adds a `RemoteRegistry` that resolves agent type definitions by pulling them from an OCI registry, complementing the existing embedded/local registry. It is not yet wired into the composite `Registry` (gated by an existing TODO), so this PR introduces the resolver and its test scaffolding without changing runtime resolution behavior.

## Changes

- `RemoteRegistry`: builds an environment-targeted OCI tag (`host-linux-<name>-<version>`, `host-windows-...`, or `kubernetes-...`), downloads through `OCIAgentTypeDownloader`, parses the definition, and rejects any whose platform does not match the running environment.
- `OCIAgentTypeDownloader` trait now carries an associated `Error` type so consumers and mocks aren't tied to the concrete downloader error.
- Test publisher (`oci-test-utils`) generalized behind an `ArtifactKind` trait so it can publish agent-type artifacts alongside agent packages; existing `PackageMediaType` callers are unchanged.
- Integration tests actually using the RemoteRegistry.

## Notes for reviewers

- Not yet active: nothing constructs `RemoteRegistry` in production paths, so this is additive.
- The OCI tag uses only `name` and `version`, not the agent type namespace. This is the expected behavior as `namespace` is supposed to be used to support additional remotes.

### DIFF
--- a/agent-control/src/agent_type/registry/remote.rs
+++ b/agent-control/src/agent_type/registry/remote.rs
@@ -11,15 +11,12 @@ use crate::environment::Environment;
 /// resolves them from its own running [Environment], which is used to build the OCI tag
 /// `<platform>-<operating_system>-<name>-<version>` (the operating system segment is omitted on
 /// kubernetes, matching how agent type metadata maps platform/os to [Environment]).
-// TODO: not yet wired into `Registry`; remove the `allow(dead_code)` once the composite registry
-// consumes it (see the precedence TODO in `Registry`).
-#[allow(dead_code)]
+// TODO: not yet wired into the composite `Registry` (see its precedence TODO).
 pub struct RemoteRegistry<D> {
     environment: Environment,
     downloader: D,
 }
 
-#[allow(dead_code)]
 impl<D: OCIAgentTypeDownloader> RemoteRegistry<D> {
     pub fn new(environment: Environment, downloader: D) -> Self {
         Self {
@@ -49,13 +46,24 @@ impl<D: OCIAgentTypeDownloader> AgentTypeRegistry for RemoteRegistry<D> {
         agent_type_id: &AgentTypeID,
     ) -> Result<AgentTypeDefinition, AgentTypeRegistryError> {
         let tag = self.artifact_tag(agent_type_id);
-        let definition = self
+        let raw = self
             .downloader
             .download(&tag)
             .map_err(|err| AgentTypeRegistryError::Remote(err.to_string()))?;
-        Ok(serde_saphyr::from_slice::<AgentTypeDefinition>(
-            &definition,
-        )?)
+        let definition = serde_saphyr::from_slice::<AgentTypeDefinition>(&raw)?;
+
+        // The tag targets a single environment, so a definition for a different one means the
+        // remote returned an artifact we did not ask for. Reject it rather than supervise an
+        // agent type meant for another platform.
+        let found = definition.metadata.environment;
+        if found != self.environment {
+            return Err(AgentTypeRegistryError::EnvironmentMismatch {
+                requested: self.environment,
+                found,
+            });
+        }
+
+        Ok(definition)
     }
 }
 
@@ -102,6 +110,24 @@ deployment:
         let definition = registry.get(&agent_type_id()).unwrap();
 
         assert_eq!(definition.metadata.environment, Environment::K8s);
+    }
+
+    #[test]
+    fn test_get_rejects_definition_for_a_different_environment() {
+        let mut downloader = MockOCIAgentTypeDownloader::new();
+        // The downloader returns a kubernetes definition while the registry runs on Linux.
+        downloader
+            .expect_download()
+            .returning(|_| Ok(K8S_DEFINITION.as_bytes().to_vec()));
+
+        let registry = RemoteRegistry::new(Environment::Linux, downloader);
+        assert_matches!(
+            registry.get(&agent_type_id()),
+            Err(AgentTypeRegistryError::EnvironmentMismatch {
+                requested: Environment::Linux,
+                found: Environment::K8s,
+            })
+        );
     }
 
     #[test]
--- a/agent-control/src/agent_type/registry/remote.rs
+++ b/agent-control/src/agent_type/registry/remote.rs
@@ -46,25 +46,18 @@ impl<D: OCIAgentTypeDownloader> RemoteRegistry<D> {
         &self,
         metadata: &AgentTypeMetadata,
         expected_id: &AgentTypeID,
-        tag: &str,
-    ) -> Result<(), AgentTypeRegistryError> {
+    ) -> Result<(), String> {
         if metadata.environment != self.environment {
-            return Err(AgentTypeRegistryError::MetadataMismatch {
-                tag: tag.to_string(),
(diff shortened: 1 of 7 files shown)


agent-control/src/agent_type/registry/remote.rs
```diff
@@ -0,0 +1,160 @@
...
+    fn get(
+        &self,
+        agent_type_id: &AgentTypeID,
+    ) -> Result<AgentTypeDefinition, AgentTypeRegistryError> {
+        let tag = self.artifact_tag(agent_type_id);
+        let raw = self
+            .downloader
+            .download(&tag)
+            .map_err(|err| AgentTypeRegistryError::Remote(err.to_string()))?;
+        let definition = serde_saphyr::from_slice::<AgentTypeDefinition>(&raw)?;
```
danielorihuela: I believe this is wrong. We should use the function @gsanchezgavier PR to check the protocol version.

agent-control/src/agent_type/registry/remote.rs
```diff
@@ -0,0 +1,160 @@
...
+
+        // The tag targets a single environment, so a definition for a different one means the
+        // remote returned an artifact we did not ask for. Reject it rather than supervise an
+        // agent type meant for another platform.
+        let found = definition.metadata.environment;
```
danielorihuela: Just to be clear. For this to happen, the agent owner should add a "linux" package with an agent type that says `platform: kubernetes` or something like that. Is that it?

agent-control/src/agent_type/registry/remote.rs
```diff
@@ -0,0 +1,160 @@
...
+        AgentTypeID::try_from("newrelic/com.newrelic.infrastructure:0.1.0").unwrap()
+    }
+
+    #[rstest::rstest]
+    #[case::linux(Environment::Linux, "host-linux-com.newrelic.infrastructure-0.1.0")]
+    #[case::windows(Environment::Windows, "host-windows-com.newrelic.infrastructure-0.1.0")]
+    #[case::kubernetes(Environment::K8s, "kubernetes-com.newrelic.infrastructure-0.1.0")]
+    fn test_artifact_tag(#[case] environment: Environment, #[case] expected_tag: &str) {
+        let registry = RemoteRegistry::new(environment, MockOCIAgentTypeDownloader::new());
+        assert_eq!(registry.artifact_tag(&agent_type_id()), expected_tag);
```
danielorihuela, line 130: Nit: `artifact_tag` should be a separated private function.

agent-control/src/agent_type/registry/remote.rs
sigilioso: Exactly, there should be an error on the remote (Eg: linux agent type published in the OCI repository as k8s)

agent-control/src/agent_type/registry/remote.rs
sigilioso: Well, it is right until #2589 gets merged 😉 

However, you are totally right. This is the conflicting part: if this got merged before, we would need to update it when #2589 is merged.

gsanchezgavier: good thing is that it wont compile in any case :)

agent-control/src/agent_type/registry/remote.rs
gsanchezgavier: I wonder if make sense to also check for the rest of the metadata , in other words why only environment?

gsanchezgavier (dismissed):
thanks i just left a minor question

agent-control/src/agent_type/registry/remote.rs
sigilioso: The rest of metadata is already checked when de-serializing, isn't it? We need to specifically check the environment  because we represent any environment with the enum.

agent-control/src/agent_type/registry/remote.rs
danielorihuela: AH okey, then I resolve this.

test/crates/oci-test-utils/src/publisher.rs
```diff
@@ -72,30 +98,50 @@ impl PackagePublisher {
...
+            .await
+        })
     }
 
-    async fn push_async(&self, file: &Path, media_type: PackageMediaType, tag: &str) -> Reference {
+    async fn push_async(
+        &self,
+        file: &Path,
+        layer_media_type: &str,
+        manifest_artifact_type: &str,
```
danielorihuela, line 125: I wonder if we could make this generic and pass the artifact kind directly. I had this idea, I haven't actually thought about it deeply.

danielorihuela (dismissed):

agent-control/src/agent_type/registry/remote.rs
gsanchezgavier: > The rest of metadata is already checked when de-serializing, isn't it?

the checks are similar to what the environment checks are performed. So on de-ser we check that everything is consistent. In this case you are checking that what is expected to be in the artifact is there, but only for the environment and not for the version for instance. The same way the environment could not match, could happen with the version i guess.

sigilioso: Thanks for clarifying! I'm adding the rest of validations as well.

test/crates/oci-test-utils/src/publisher.rs
sigilioso: we could, but considering that it is private and that `push_with_tag<A: ArtifactKind>` is already resolving the generic, I don't think it is worth it.

agent-control/src/agent_type/registry/remote.rs
sigilioso: I've just extended the validation here: [87db633](https://github.com/newrelic/newrelic-agent-control/pull/2590/commits/87db633da01bd0c86b3c6ff73f9a895a8cf55383)

agent-control/src/agent_type/registry/remote.rs
sigilioso: I'm leaving this unchanged for now because it is likely changing when @gsanchezgavier extends the name validation

agent-control/src/agent_type/registry/remote.rs
```diff
@@ -38,6 +38,36 @@ impl<D: OCIAgentTypeDownloader> RemoteRegistry<D> {
...
     }
+
+    /// Verifies that the downloaded definition's metadata matches what was requested: its
+    /// environment must equal the running one and its id must equal `expected_id`. A mismatch
+    /// means the remote returned an artifact we did not ask for, so it is rejected.
+    fn check_metadata(
+        &self,
+        metadata: &AgentTypeMetadata,
+        expected_id: &AgentTypeID,
+        tag: &str,
```
gsanchezgavier: couldn't we decorate the errors with tag outside and avoid passing it ? if i saw correctly is just being used for that

sigilioso: Decorating outside now, the helper returns error the details as `Err(String)` now. LMKWYT

gsanchezgavier (approved):
thanks!

PR #2591, opened 2026-06-09
chore: pin pipeline tool dependencies to explicit versions

danielorihuela:
Following a prior memory leak incident where an action was version-pinned but its internal dependencies were not, this PR audits all CI workflow files and pins every tool installed without an explicit version.

**Changes:**
- pin `cargo-xwin` to `0.22.0` and `go-winres` to `v0.3.3`
- pin `docker/setup-buildx-action` internal version to `v0.34.1`

There are other dependencies we could pin but I decided they were not important. For example, `cargo-audit`, `tilt` or `commitlint`. The reason being that these are **not** used for building the binary.


sigilioso (approved):
LGTM

PR #2594, opened 2026-06-10
refactor: use common oci fetcher for downloaders

sigilioso:
> [!NOTE]
> On top of #2590
> Addresses: https://github.com/newrelic/newrelic-agent-control/pull/2580#discussion_r3379489541

## Summary

The agent-type and package OCI downloaders each carried their own copy of the "verify signature, then pull, with retries" flow and their own string error type. This extracts that orchestration into a single `OciArtifactFetcher` in the `oci` module and folds both bespoke error types into the shared `OciClientError`, so the two downloaders now only describe how to materialize their specific artifact.

## Changes

- Add `OciArtifactFetcher` (owns the client, auth, and retry policy) and `Client::verified_reference`, which collapses the old `Some(pk)/None` signature-verification branching.
- Rename the package types for clarity now that an agent-type downloader exists: `OCIArtifactDownloader` to `OCIPackageArtifactDownloader`, trait `OCIAgentDownloader` to `OCIPackageDownloader`.
- Replace `OCIDownloaderError` and `OCIAgentTypeDownloaderError` with `OciClientError`.
-  Add `OciClientError::AttemptsExceeded` so retry exhaustion is a typed variant rather than a nested string wrap.
- Move the shared verify-then-fetch tests (TOCTOU, MITM, missing manifest, unsigned-but-verification-enabled) into the fetcher's test module, since the logic now lives there.

### DIFF
--- a/agent-control/src/agent_type/oci/downloader.rs
+++ b/agent-control/src/agent_type/oci/downloader.rs
@@ -2,24 +2,16 @@ use std::time::Duration;
 
 use oci_client::Reference;
 use oci_client::secrets::RegistryAuth;
-use thiserror::Error;
 use tracing::{debug, warn};
 use url::Url;
 
 use crate::agent_control::config::{OciAuth, Registry};
 use crate::agent_type::runtime_config::on_host::package::rendered::Repository;
-use crate::oci::Client;
 use crate::oci::artifact_definitions::LocalAgentType;
-use crate::utils::retry::retry;
-
-const DEFAULT_RETRIES: usize = 0;
+use crate::oci::{Client, OciArtifactFetcher, OciClientError};
 
 /// Maximum size of an agent type artifact blob (generous upper bound).
 const MAX_ARTIFACT_SIZE_BYTES: usize = 10 * 1024 * 1024; // 10 MiB
-
-#[derive(Debug, Error)]
-#[error("downloading agent type OCI artifact: {0}")]
-pub struct OCIAgentTypeDownloaderError(String);
 
 /// An interface for downloading Agent Type definitions from a configured OCI remote.
 pub trait OCIAgentTypeDownloader: Send + Sync {
@@ -40,18 +32,15 @@ pub trait OCIAgentTypeDownloader: Send + Sync {
 /// The artifact content is never written to disk, so resolution does not depend on a writable
 /// filesystem location (relevant on Kubernetes). Blobs larger than `max_size_bytes` are rejected.
 pub struct OCIAgentTypeArtifactDownloader {
-    client: Client,
+    fetcher: OciArtifactFetcher,
     registry: Registry,
     repository: Repository,
-    auth: RegistryAuth,
     public_key_url: Option<Url>,
-    max_retries: usize,
-    retry_interval: Duration,
     max_size_bytes: usize,
 }
 
 impl OCIAgentTypeDownloader for OCIAgentTypeArtifactDownloader {
-    type Error = OCIAgentTypeDownloaderError;
+    type Error = OciClientError;
 
     /// Downloads the agent type artifact at `<registry>/<repository>:<tag>`.
     ///
@@ -61,7 +50,7 @@ impl OCIAgentTypeDownloader for OCIAgentTypeArtifactDownloader {
     ///
     /// On failure the operation is retried as configured; if all attempts are exhausted it returns
     /// an error.
-    fn download(&self, tag: &str) -> Result<Vec<u8>, OCIAgentTypeDownloaderError> {
+    fn download(&self, tag: &str) -> Result<Vec<u8>, OciClientError> {
         let base_reference = Reference::with_tag(
             self.registry.to_string(),
             self.repository.to_string(),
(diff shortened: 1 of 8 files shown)


alvarocabanas (dismissed):
LGTM, the only point that I see questionable, is we are building OciClientError outside of the oci client in the fetcher and so instead of a custom error, it's not something I am against, but just would like to know the rationale.

sigilioso:
> LGTM, the only point that I see questionable, is we are building OciClientError outside of the oci client in the fetcher and so instead of a custom error, it's not something I am against, but just would like to know the rationale.

I tried to simplify as much as possible, the fetcher can be seen as a client _extension_ (I actually kept it in the same module for that reason). We could definitely add an additional error, but I don't think it adds value.

agent-control/src/oci.rs
```diff
@@ -168,6 +174,83 @@ impl Client {
...
             .block_on(self.verify_signature_with_public_keys(reference, &public_keys, auth))
     }
+
+    /// Returns the [Reference] to download from, verifying its signature when required.
+    ///
+    /// When `public_key_url` is `Some`, the artifact's signature is verified via
+    /// [Self::verify_signature] and the returned reference is digest-pinned (assuring the artifact
+    /// downloaded is the one verified). When it is `None`, signature verification is skipped and the
+    /// `base` reference is returned unchanged.
+    pub fn verified_reference(
```
danielorihuela: Nit: This function is only used in this file. Shouldn't be public.

agent-control/src/oci.rs
```diff
@@ -323,6 +406,169 @@ pub mod tests {
...
+            .with_signature(&key_pair);
+        let server = MockServer::start();
+        oci_mock_a.setup_mocks_on(&server);
+
+        // Verify the signature to obtain a digest-pinned reference.
+        let reference = oci_mock_a.reference_on_server(&server);
+        let fetcher = create_fetcher();
+        let verified_reference = fetcher
+            .client
+            .verified_reference(&reference, Some(&jwks_server.url), &RegistryAuth::Anonymous)
```
danielorihuela: `verified_reference` it's not 100% needed. Have you thought about inlining it?
The issue is that this test must be modified.
Something along these lines. Not sure it's worth it just to get rid of one function.

```rust
let original = fetcher
    .fetch(
        &reference,
        Some(&jwks_server.url),
        |client, reference, auth| {
            // Move the tag after verification (TOCTOU attack).
            let oci_mock_b = FakeOciServer::new("repo", "v1")
                .with_layer(MALICIOUS, "fake-media_type")
                .with_signature(&key_pair);
            server.reset();
            oci_mock_b.setup_mocks_on(&server); // new tag takes precedence
            oci_mock_a.setup_mocks_on(&server); // keep the previous digest and blobs reachable

            pull_first_layer(client, reference, auth)
        },
    )
    .unwrap();
assert_eq!(original, ORIGINAL);

let malicious = fetcher.fetch(&reference, Some(&jwks_server.url), pull_first_layer).unwrap();
assert_eq!(malicious, MALICIOUS);
```

agent-control/src/agent_type/oci/downloader.rs
```diff
@@ -40,18 +32,15 @@ pub trait OCIAgentTypeDownloader: Send + Sync {
...
-    auth: RegistryAuth,
     public_key_url: Option<Url>,
-    max_retries: usize,
-    retry_interval: Duration,
     max_size_bytes: usize,
 }
 
 impl OCIAgentTypeDownloader for OCIAgentTypeArtifactDownloader {
-    type Error = OCIAgentTypeDownloaderError;
+    type Error = OciClientError;
```
danielorihuela, line 43: The `OciClientError` is pub. The associated type is no longer needed. It was introduced because of a private error type IIRC.

sigilioso: Even if we can remove it now that we are using the `OciClientError`, I think it also simplifies the mock definition 

```rust
    /// A trivially-constructible error for the mock downloader, so consumers (e.g. the remote
    /// registry) can exercise the failure path without access to the real downloader's internals.
    #[derive(Debug, thiserror::Error)]
    #[error("{0}")]
    pub struct FakeDownloaderError(pub String);

    mock! {
        pub OCIAgentTypeDownloader {}
        impl OCIAgentTypeDownloader for OCIAgentTypeDownloader {
            type Error = FakeDownloaderError;
            fn download(&self, tag: &str) -> Result<Vec<u8>, FakeDownloaderError>;
        }
```

Do you see any downsides of keeping it?

agent-control/src/oci.rs
sigilioso: Instead of inlining it, I've moved to the _fetcher_ struct. I think that keeping the helper makes the code easier to read and now it is in the proper struct. Thanks for pointing it out! [e268e1b](https://github.com/newrelic/newrelic-agent-control/pull/2594/commits/e268e1bcc7d244bfcd26c2920f507ea66f91cbd4)

agent-control/src/oci.rs
sigilioso: Now it is private (and also moved: [e268e1b](https://github.com/newrelic/newrelic-agent-control/pull/2594/commits/e268e1bcc7d244bfcd26c2920f507ea66f91cbd4))

agent-control/src/agent_type/oci/downloader.rs
```diff
@@ -61,7 +50,7 @@ impl OCIAgentTypeDownloader for OCIAgentTypeArtifactDownloader {
     ///
     /// On failure the operation is retried as configured; if all attempts are exhausted it returns
     /// an error.
-    fn download(&self, tag: &str) -> Result<Vec<u8>, OCIAgentTypeDownloaderError> {
+    fn download(&self, tag: &str) -> Result<Vec<u8>, OciClientError> {
```
danielorihuela, line 53: This functions is very very similar to it's `OCIPackageArtifactDownloader` counterpart. I wonder if we could build an `Oci` struct first and have the same input. The idea would be to keep simplifying this wrapper types.

`should_verify_signature` behaviour is different.
`download_definition` and `download_package_artifact` are really similar. The only difference is that one creates a path and the other requires a max. An enum could help with that.

I wonder if increasing the lower layer would be worth. Maybe I can try this out next week.

agent-control/src/agent_type/oci/downloader.rs
danielorihuela: I don't understand why it's easier. You have this extra fake error.
In any case, I don't have a strong opinion. We can keep it.

danielorihuela (dismissed):

agent-control/src/agent_type/oci/downloader.rs
sigilioso: Only because the fake error is just a plain string, when using the real-one you need to choose a variant that may or may not fit the test, just semantics. I don't have a strong opinion neither!

agent-control/src/agent_type/oci/downloader.rs
sigilioso: If you address it, also take into account that the client's method they use is also different (one pulls the blob and stores it in memory, the other writes directly to a file).

danielorihuela (approved):

PR #2597, opened 2026-06-11
fix: validate version on deserialize

danielorihuela:
Validates the version on deserialize, not only when calling `from_str`. In other words, now we validate the version that comes from remote too. Tested with a regression test.

Versions are still compared based on it's string version because we are not using SemVer but the OCI format, which can include tags and/or digests.

```rust
if new_version.to_string() == AGENT_CONTROL_VERSION {
```

### DIFF
--- a/agent-control/src/agent_type/runtime_config/on_host/package/rendered.rs
+++ b/agent-control/src/agent_type/runtime_config/on_host/package/rendered.rs
@@ -109,7 +109,8 @@ fn validate_repository_char(c: char) -> Result<(), InvalidRepository> {
 #[error("{0}")]
 pub struct InvalidVersion(String);
 
-#[derive(Debug, Clone, PartialEq, Serialize)]
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+#[serde(try_from = "String")]
 pub struct Version(String);
 
 impl Display for Version {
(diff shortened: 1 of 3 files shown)


agent-control/src/agent_type/runtime_config/on_host/package/rendered.rs
```diff
@@ -109,7 +109,7 @@ fn validate_repository_char(c: char) -> Result<(), InvalidRepository> {
 #[error("{0}")]
 pub struct InvalidVersion(String);
 
-#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+#[derive(Debug, Clone, PartialEq, Serialize)]
```
sigilioso: Would it make sense to  use

```rust
#[serde(try_from = "String")]
```

and add the implementation for `TryFrom<String>` for version?

sigilioso (dismissed):
LGTM, I've left a suggestion with is actually a _nit_

agent-control/src/agent_type/runtime_config/on_host/package/rendered.rs
danielorihuela: Done. Thanks to that I fixed the regression test!!!

At first, I used `format!` and `agents: {{}}` worked as expected.
Then I changed it to just have a string and `{{}}` throws always an error. I had to change it to `{}` now.

sigilioso: I totally missed that when reviewing!

sigilioso (approved):
🚀

PR #2598, opened 2026-06-11
fix: cache windows artifacts for pipeline

danielorihuela:
Windows build aren't cached. This PR solves this. I don't think it will have a big impact, but it was easy to fix.

Captura de pantalla 2026-06-11 a las 16 03 36

Captura de pantalla 2026-06-11 a las 16 03 21


vjripoll (approved):
LGTM!

PR #2599, opened 2026-06-11
feat: add new restart policy for windows and linux

alvarocabanas:
Hardens service restart policies on Linux and Windows (systemd rate limiting: 5 restarts max in 60s) to prevent crash-looping from saturating CPU, it adds startup validation to detect misconfigured restart policies and log warnings.

### DIFF
--- a/build/package/newrelic-agent-control.service
+++ b/build/package/newrelic-agent-control.service
@@ -8,6 +8,13 @@ EnvironmentFile=/etc/newrelic-agent-control/systemd-env.conf
 ExecStart=/usr/bin/newrelic-agent-control
 KillMode=mixed
 Restart=always
+# Restart rate limiting to prevent crash-looping from saturating CPU
+# Wait 5s between attempts; tolerate up to 5 starts per rolling
+# 60s; the 6th crash inside that window stops the loop instead of spinning
+# forever. A pathological crash-loop is bounded to ~5 restarts/minute.
+RestartSec=5
+StartLimitIntervalSec=60
+StartLimitBurst=5
 Type=simple
 
 [Install]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ Remember that the keywords that you can use are the following:
 
 ## Unreleased
 
+### enhancement
+- Hardens service restart policies on Linux and Windows (systemd rate limiting: 5 restarts max in 60s) to prevent
+  crash-looping from saturating CPU.
+
 ## v1.17.0 - 2026-06-16
 
 ### 🚀 Enhancements
--- a/build/package/windows/install.ps1
+++ b/build/package/windows/install.ps1
@@ -189,8 +189,11 @@ if ($LASTEXITCODE -ne 0) {
     exit $LASTEXITCODE
 }
 
-# Configure failure actions: restart after 30 seconds; reset failure count daily
-sc.exe failure $serviceName reset= 86400 actions= restart/30000 | Out-Null
+# Configure failure actions to prevent crash-looping from saturating CPU.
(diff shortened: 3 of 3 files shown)


build/package/newrelic-agent-control.service
```diff
@@ -8,6 +8,10 @@ EnvironmentFile=/etc/newrelic-agent-control/systemd-env.conf
 ExecStart=/usr/bin/newrelic-agent-control
 KillMode=mixed
 Restart=always
+# Restart rate limiting to prevent crash-looping from saturating CPU
+RestartSec=5
+StartLimitIntervalSec=60
+StartLimitBurst=5
```
gsanchezgavier, line 17: could you add a description that reflects how this behaves?, it is a bit hard to understand just by reading it.

agent-control/src/utils/restart_policy_validator.rs
```diff
@@ -0,0 +1,212 @@
...
+                    Self-update requires Restart=always in the systemd unit. \
+                    Current setting: {}. \
+                    Self-update may not function correctly.",
+                restart_value.trim()
+            );
+        }
+    }
+}
+
+/// Validates the restart policy at startup and logs warnings if misconfigured.
```
gsanchezgavier: i might lack some context, but this sounds like a bit too complex check just for a warning log. If we are not preventing AC from run i would remove these validations

agent-control/src/utils/restart_policy_validator.rs
```diff
@@ -0,0 +1,212 @@
...
+        return;
+    }
+
+    // Try to read service properties via systemctl
+    if let Ok(output) = std::process::Command::new("systemctl")
+        .arg("show")
+        .arg(LINUX_SERVICE_NAME)
+        .arg("--property=Restart")
+        .output()
+        && output.status.success()
```
gsanchezgavier: same here, this is too much IMO just to log a warning, also this could silently fail and print nothing

agent-control/src/agent_control/run/on_host.rs
```diff
@@ -85,6 +86,8 @@ type OnHostOpAMPConsumer = EventConsumer<OpAMPEvent>;
 
 impl AgentControlRunner {
     pub fn run_onhost(self) -> Result<GracefulShutdownReason, RunError> {
+        validate_restart_policy();
```
danielorihuela: Would it make sense to make this check at build time? Maybe in another PR.

If a client gets this, this means that the version we shipped is wrong. It's not something the user did wrongly. Am I on the correct path? If that's the case, should we do this validation on build time somehow? A github pipeline that checks the systemd file and `install.ps1`?

alvarocabanas: We may remove this check entirely, Guillermo also points that it is overly complicated only for a warning.

agent-control/src/utils/restart_policy_validator.rs
alvarocabanas: Let's discuss it, we can maybe remove this.

build/package/newrelic-agent-control.service
alvarocabanas: Done in last commit

agent-control/src/utils/restart_policy_validator.rs
alvarocabanas: Removed in last commit.

agent-control/src/utils/restart_policy_validator.rs
alvarocabanas: Removed in last commit.

danielorihuela (dismissed):

danielorihuela (approved):

PR #2601, opened 2026-06-15
feat: enforce format constraints on agent type metadata (NR-568040)

gsanchezgavier:
## Summary

Agent type `name`/`namespace` and `version` allowed values (like `-` separators or arbitrary semver pre-release/build tags) that can't safely become OCI tags. This tightens validation and surfaces precise, field-specific errors at deserialization time.

## Changes

- Reject `-` in name/namespace; require start-letter, end-alphanumeric, and only `[a-z0-9._]`.
- Require plain `Major.Minor.Patch` semver and cap version length (14 chars) to bound OCI tag size.
- Centralize parsing/validation in `AgentTypeID::from_parts`, shared by `TryFrom<&str>` and `Deserialize`.
- Unify the related changelog entries to reduce noise
- Replace the AC version usage on the AC fake agent type , wich didn't make much sense and also it was causing e2e test to fail due to long version values. 
## Notes for reviewers

- All shipped agent types already comply; only internal authors of custom definitions need to migrate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,13 +16,11 @@ Remember that the keywords that you can use are the following:
 
 ### enhancement
 - Added support for JP endpoints.
-- Agent type definitions now use a flat per-platform schema (one YAML file per `(platform, operating_system)` pair). 
-  Agent type FQNs and configuration values are unchanged, so this is **not a breaking change** for end users.
-  Internal authors of custom agent type definitions need to migrate their YAMLs — see [docs/INTEGRATING_AGENTS.md](docs/INTEGRATING_AGENTS.md).
-- Agent type definitions now declare a top-level `protocol_version` (a quoted `MAJOR.MINOR` string) that versions the
-  agent-type schema language itself. It is validated against the version Agent Control supports at registry ingestion,
-  so definitions targeting an incompatible schema are rejected early. Internal authors of custom agent type definitions
-  must add this field — see [docs/INTEGRATING_AGENTS.md](docs/INTEGRATING_AGENTS.md).
+- Agent type definitions changed: flat per-platform schema (one YAML per `(platform, operating_system)` pair), a new
+  top-level `protocol_version` field (quoted `MAJOR.MINOR`, validated at registry ingestion), and stricter `name`/`namespace`
+  validation (no `-`) and `version` (plain `Major.Minor.Patch` semver). FQNs and configuration values are unchanged, so this
+  is **not a breaking change** for end users; internal authors of custom definitions must migrate — see
+  [docs/INTEGRATING_AGENTS.md](docs/INTEGRATING_AGENTS.md).
 
 ### bugfix
 - Validate version coming from remote.
--- a/agent-control/src/agent_type/agent_type_id.rs
+++ b/agent-control/src/agent_type/agent_type_id.rs
@@ -4,16 +4,39 @@ use std::fmt::{Display, Formatter};
 use thiserror::Error;
 
 const NAME_NAMESPACE_MIN_LENGTH: usize = 1;
-const NAME_NAMESPACE_MAX_LENGTH: usize = 64;
+pub(crate) const NAME_NAMESPACE_MAX_LENGTH: usize = 64;
+/// Bounds the version length so a version bump (e.g. `9999.9999.999` -> `9999.9999.1000`)
+/// cannot overflow external constraints like the OCI tag built using this metadata.
+pub(crate) const VERSION_MAX_LENGTH: usize = 14;
 
 #[derive(Error, Debug, PartialEq)]
 pub enum AgentTypeIDError {
-    #[error("invalid AgentType namespace")]
-    InvalidNamespace,
-    #[error("invalid AgentType name")]
-    InvalidName,
-    #[error("invalid AgentType version")]
-    InvalidVersion,
+    #[error("invalid namespace: {0}")]
+    InvalidNamespace(NameFormatError),
+    #[error("invalid name: {0}")]
+    InvalidName(NameFormatError),
+    #[error("invalid version: {0}")]
+    InvalidVersion(String),
+    #[error("only Major.Minor.Patch semver format is allowed")]
+    ForbiddenSemVer,
+    #[error("version must not be longer than {max} characters, but it is {length}")]
+    VersionTooLong { length: usize, max: usize },
+}
+
+#[derive(Error, Debug, PartialEq)]
+pub enum NameFormatError {
+    #[error("must not be empty")]
+    Empty,
+    #[error("must be at most {max} characters, but it is {length}")]
+    TooLong { length: usize, max: usize },
+    #[error("must start with an ASCII letter")]
+    InvalidStart,
+    #[error("must end with a letter or a digit")]
+    InvalidEnd,
+    #[error(
+        "contains invalid character '{0}', only lowercase letters, digits, '.' and '_' are allowed"
+    )]
+    InvalidCharacter(char),
 }
 
 /// Holds agent type metadata that uniquely identifies an agent type.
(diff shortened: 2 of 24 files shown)


agent-control/src/environment.rs
```diff
@@ -16,3 +16,23 @@ impl Display for Environment {
...
+        }
+
+        fn succeeding(current: &Environment) -> Option<Environment> {
+            match current {
+                Environment::Linux => Some(Environment::Windows),
+                Environment::Windows => Some(Environment::K8s),
+                Environment::K8s => None,
+            }
+        }
+    }
```
DavSanchez, line 37: Not sure I understand what's this for, but I guess we can just use a static array of all the variants and just do an `into_iter` or something like this? If we want to make sure we use all variants we can place a no-op `match` inside or something.

```rust
impl Environment {
  pub fn all() -> impl Iterator<Item = Self> {
    [Self::Linux, Self::Windows, Self::K8s].into_iter()
  }
}
```

gsanchezgavier: that was the first approach but it doesn't force you to keep it up to date, and since this is used on a test that we want to check on every variant i thought it was worth to implement something a bit more robust, that will fail to compile if a new variant is added

DavSanchez: Pity that https://doc.rust-lang.org/std/mem/fn.variant_count.html is still not stabilized, otherwise we could have the slice with a comptime guarantee that it contains the same number of elements as variants.

DavSanchez: Perhaps we might want to add a new type that implements `Iterator<Item = Environment>`?

gsanchezgavier: wound't be too much for some testing helper ?

DavSanchez: 🤷‍♂️ no strong opinion, depends on the testing approach

gsanchezgavier: i don't think it will help on readability for this case TBH

agent-control/src/agent_type/agent_type_id.rs
```diff
@@ -36,18 +59,61 @@ impl AgentTypeID {
...
-                c.eq(&'-')
-                    || c.eq(&'_')
-                    || c.eq(&'.')
-                    || c.is_ascii_digit()
-                    || c.is_ascii_lowercase()
-            })
+    /// Parses a semver version, additionally rejecting anything that isn't a digit or `.`.
+    /// This constraint relates to the OCI tag built from this metadata.
+    fn parse_version(s: &str) -> Result<Version, AgentTypeIDError> {
+        if !s.chars().all(|c| c.is_ascii_digit() || c.eq(&'.')) {
```
alvarocabanas: Can you change c.eq(&'.') for c == '.' to have better readability?

agent-control/src/environment.rs
danielorihuela: We could use `strum` but it's an extra dependency.

agent-control/src/agent_type/agent_type_id.rs
```diff
@@ -36,18 +59,61 @@ impl AgentTypeID {
...
+            });
+        }
+        Version::parse(s).map_err(|e| AgentTypeIDError::InvalidVersion(e.to_string()))
+    }
+
+    /// Limits the characters allowed so this could be used as identifier for k8s resources and other
+    /// places like OCI tags. Returns the specific [NameFormatError] on the first rule violated.
+    fn validate_format(s: &str) -> Result<(), NameFormatError> {
+        if s.len() < NAME_NAMESPACE_MIN_LENGTH {
+            return Err(NameFormatError::Empty);
```
danielorihuela, line 80: Nit: Use `TooShort` instead of `Empty`.

If the minimum length ever changes, we could end up with a confusing error. I don't know if this could ever happen. If not, NAME_NAMESPACE_MIN_LENGTH is not necessary.

agent-control/src/agent_type/oci/downloader.rs
```diff
@@ -181,6 +184,15 @@ pub mod tests {
     const REPOSITORY: &str = "my-org/agent-types-repository";
     const TAG: &str = "host-linux-some.agent.type-0.0.42";
 
+    /// An [AgentTypeTag] whose string equals [TAG], so it matches the artifact the mock server
+    /// serves under that tag.
+    fn agent_type_tag() -> AgentTypeTag {
```
paologallinaharbur: maybe it is obvious, but can't we change it to test_agent_type_tag()?

agent-control/src/agent_type/oci/downloader.rs
```diff
@@ -181,6 +184,15 @@ pub mod tests {
...
+        AgentTypeTag::new(
+            &AgentTypeID::try_from("newrelic/some.agent.type:0.0.42").unwrap(),
```
paologallinaharbur, line 191: can't we re-use TAG CONSTANT above?

agent-control/src/agent_type/agent_type_id.rs
```diff
@@ -36,18 +59,61 @@ impl AgentTypeID {
...
+            return Err(NameFormatError::InvalidStart);
+        }
+        if !s.ends_with(|c: char| c.is_ascii_alphanumeric()) {
+            return Err(NameFormatError::InvalidEnd);
+        }
+        if let Some(invalid) = s
+            .chars()
+            .find(|c| !(c.eq(&'_') || c.eq(&'.') || c.is_ascii_digit() || c.is_ascii_lowercase()))
+        {
+            return Err(NameFormatError::InvalidCharacter(invalid));
```
danielorihuela, line 98: Nit: Adding the position of the character might help debugging in the future.

For example, "n​e​w​r​e​l​i​c" contains invisible characters which are invalid like `\u{200b}`. Knowing the position might help understanding the root cause.

agent-control/src/agent_type/agent_type_id.rs
```diff
@@ -36,18 +59,61 @@ impl AgentTypeID {
...
+        }
+        Ok(())
+    }
+
+    fn from_parts(
+        name: String,
+        namespace: String,
+        version: &str,
```
danielorihuela: Nit: Take ownership of the `version`.

I think it makes sense to take ownership of the version as part of this function. Once this is called, the user shouldn't use the old version, only the new one.

agent-control/src/agent_type/agent_type_id.rs
```diff
@@ -36,18 +59,61 @@ impl AgentTypeID {
...
-            && s.chars().all(|c| {
-                c.eq(&'-')
-                    || c.eq(&'_')
-                    || c.eq(&'.')
-                    || c.is_ascii_digit()
-                    || c.is_ascii_lowercase()
-            })
+    /// Parses a semver version, additionally rejecting anything that isn't a digit or `.`.
+    /// This constraint relates to the OCI tag built from this metadata.
+    fn parse_version(s: &str) -> Result<Version, AgentTypeIDError> {
```
paologallinaharbur, line 63: "...." woun't this pass the validation?

agent-control/src/agent_type/agent_type_id.rs
```diff
@@ -86,30 +152,15 @@ impl TryFrom<&str> for AgentTypeID {
 
     fn try_from(value: &str) -> Result<Self, Self::Error> {
```
danielorihuela, line 152: Nit

```rust
let (namespace, rest) = value.split_once('/').unwrap_or_default();
let (name, version) = rest.split_once(':').unwrap_or_default();

Self::from_parts(name.to_string(), namespace.to_string(), version)
```

agent-control/src/agent_type/agent_type_id.rs
```diff
@@ -36,18 +59,61 @@ impl AgentTypeID {
...
+        if s.len() > VERSION_MAX_LENGTH {
+            return Err(AgentTypeIDError::VersionTooLong {
+                length: s.len(),
+                max: VERSION_MAX_LENGTH,
+            });
+        }
+        Version::parse(s).map_err(|e| AgentTypeIDError::InvalidVersion(e.to_string()))
+    }
+
+    /// Limits the characters allowed so this could be used as identifier for k8s resources and other
```
paologallinaharbur: why are we mentioning k8s objects?

agent-control/src/agent_type/agent_type_id.rs
paologallinaharbur: do you think that we will ever need to support some "complex" semVer "1.2.3-dev"?

agent-control/src/agent_type/agent_type_id.rs
```diff
@@ -86,30 +152,15 @@ impl TryFrom<&str> for AgentTypeID {
...
             .skip_while(|&i| i != '/')
             .skip(1)
             .take_while(|&i| i != ':')
             .collect();
-        if !AgentTypeID::has_valid_format(name.as_str()) {
-            return Err(AgentTypeIDError::InvalidName);
-        }
-
-        let version_str: String = value.chars().skip_while(|&i| i != ':').skip(1).collect();
+        let version: String = value.chars().skip_while(|&i| i != ':').skip(1).collect();
```
paologallinaharbur: I think you can simplify all this logic with something like:
```rust                                                                                                                                                                                                                                                                                                       
  fn try_from(value: &str) -> Result<Self, Self::Error> {                                                                                                                                                                                                                                                          
      let (head, version) = value.split_once(':').unwrap_or((value, ""));                                                                                                                                                                                                                                          
      let (namespace, name) = head.split_once('/').unwrap_or(("", head));    
```

CHANGELOG.md
```diff
@@ -16,13 +16,11 @@ Remember that the keywords that you can use are the following:
...
 ### enhancement
 - Added support for JP endpoints.
-- Agent type definitions now use a flat per-platform schema (one YAML file per `(platform, operating_system)` pair). 
-  Agent type FQNs and configuration values are unchanged, so this is **not a breaking change** for end users.
-  Internal authors of custom agent type definitions need to migrate their YAMLs — see [docs/INTEGRATING_AGENTS.md](docs/INTEGRATING_AGENTS.md).
-- Agent type definitions now declare a top-level `protocol_version` (a quoted `MAJOR.MINOR` string) that versions the
-  agent-type schema language itself. It is validated against the version Agent Control supports at registry ingestion,
-  so definitions targeting an incompatible schema are rejected early. Internal authors of custom agent type definitions
-  must add this field — see [docs/INTEGRATING_AGENTS.md](docs/INTEGRATING_AGENTS.md).
+- Agent type definitions changed: flat per-platform schema (one YAML per `(platform, operating_system)` pair), a new
```
paologallinaharbur, line 19: I like consolidating the changelog!

agent-control/src/agent_type/agent_type_id.rs
gsanchezgavier: I added this because i think this was the reason we added these constrains in the past and it was not documented

agent-control/src/agent_type/agent_type_id.rs
gsanchezgavier: no i don't think is needed.

agent-control/src/agent_type/agent_type_id.rs
gsanchezgavier: the error message already state the invalid chars and the max long of this is 64 , i think it will not improve this too much

agent-control/src/agent_type/oci/downloader.rs
gsanchezgavier: no, that is the tag, this is the fqn

agent-control/src/agent_type/agent_type_id.rs
gsanchezgavier: no, that would not be parsed as semver . notice that the semver parsing is still in place

agent-control/src/agent_type/agent_type_id.rs
paologallinaharbur: I think that the agentType length is used only for k8s labels, not for resources, no?

agent-control/src/agent_type/agent_type_id.rs
paologallinaharbur: ahh ok, that is what I was missing, thanks!

agent-control/src/agent_type/agent_type_id.rs
```diff
@@ -165,166 +205,64 @@ mod tests {
...
+        "ns/test_test_test_test_test_test_test_test_test_test_test_test_test_test:0.1.0";
+    const TOO_LONG_NAMESPACE_FQN: &str =
+        "test_test_test_test_test_test_test_test_test_test_test_test_test_test/nrdot:0.1.0";
+
+    /// Matches the expected [AgentTypeIDError] variant. Lets each [rstest] case carry the variant it
+    /// expects without comparing against the (semver-derived) error message.
+    type ErrorMatcher = fn(AgentTypeIDError) -> bool;
+
+    #[rstest]
+    #[case::empty_name("ns/:0.1.0", (|e| matches!(e, AgentTypeIDError::InvalidName(NameFormatError::Empty))) as ErrorMatcher)]
```
danielorihuela: Nit: Remove "as ErrorMatcher" from every test case.

agent-control/src/agent_type/oci/downloader.rs
```diff
@@ -50,11 +51,11 @@ impl OCIAgentTypeDownloader for OCIAgentTypeArtifactDownloader {
     ///
     /// On failure the operation is retried as configured; if all attempts are exhausted it returns
     /// an error.
-    fn download(&self, tag: &str) -> Result<Vec<u8>, OciClientError> {
+    fn download(&self, tag: &AgentTypeTag) -> Result<Vec<u8>, OciClientError> {
         let base_reference = Reference::with_tag(
             self.registry.to_string(),
             self.repository.to_string(),
-            tag.to_string(),
+            tag.as_str().to_string(),
```
danielorihuela: We have `Display` implemented for the `AgentTypeTag`. We can call `to_string` directly. We don't need to call `as_str` first.

agent-control/src/agent_type/agent_type_id.rs
danielorihuela: Okey from my side. No strong opinion. It just came to mind and wanted to share it, just in case.

danielorihuela (dismissed):
There are a lot of tests. I feel secure.
Good job!!! LGTM!!!

danielorihuela (approved):

PR #2602, opened 2026-06-15
feat: wire remote registry into composite

danielorihuela:
Wire the `RemoteRegistry` into the composite registry, giving AC the capability of getting an agent type from a remote OCI registry.

### Composite shape

**Generics** -> Unit testing with mocks.
**SupportedRegistry** -> Static dispatch for production code.
**Vector of registries** -> The idea is that this piece is open for extension when multiple registries need to be configured in the future.

### Behaviour around errors

Regardless of the error on a given registry, the composite registry will try getting the agent type from the next registry.

We could do something smarter. We could detect certain errors and short-circuit. I'm not sure which error should require this behaviour, so I didn't implement it.


danielorihuela:
@gsanchezgavier The changelog should be modified with the last PR right?

agent-control/src/agent_control/run.rs
```diff
@@ -141,6 +154,30 @@ impl AgentControlRunner {
     }
 }
 
+fn build_agent_type_registry(context: &RunnerContext, oci_client: oci::Client) -> Registry {
+    let default_remote = &context.bootstrap_config.agent_types.default_remote;
+    let public_key_url = if default_remote.signature_verification_enabled.into() {
+        Some(default_remote.public_key_url.clone())
```
gsanchezgavier: should we check this is not empty?

agent-control/tests/on_host/agent_type_composite_registry.rs
```diff
@@ -0,0 +1,176 @@
...
+    assert_eq!(definition.metadata.environment, Environment::K8s);
+
+    let variables = definition.variables.clone().flatten();
+    assert!(variables.contains_key("local_marker"));
+    assert!(!variables.contains_key("remote_marker"));
+}
+
+#[test]
+#[ignore = "needs oci registry (use *with_oci_registry suffix)"]
+fn test_local_miss_resolves_via_remote_with_oci_registry() {
```
gsanchezgavier: great tests, i think we could introduce also some scenario tests that runs in all environments excercising these two cases:
- remote fetch works (happy path of the remote fetch)
- embedded agent types doesn't trigger a remote fetch (ensure we are not breaking current installations)

agent-control/src/agent_type/registry.rs
```diff
@@ -43,37 +46,82 @@ pub struct RegistryConfig {
...
-    // As `LocalRegistry` implements AgentTypeRegistry a `Vec<impl AgentTypeRegistry>` could be considered.
+/// Resolves an [AgentTypeID] by walking an ordered list of inner registries: the first one that
+/// returns a definition wins. Any error from a layer is recorded and the walk continues to the
+/// next layer. If no layer succeeds, the composite returns the last error encountered.
+///
+/// `R` defaults to [SupportedRegistry] — the production composition (Local + Remote). The generic
+/// is there so unit tests can plug in mock implementations of [AgentTypeRegistry] without an OCI
+/// client.
+pub struct Registry<R: AgentTypeRegistry = SupportedRegistry> {
+    registries: Vec<R>,
```
gsanchezgavier, line 57: nit: perhaps adding the remote one was a bit simpler for our current needs? i'm not sure if this will be extended TBH, anyway the current approach is not complex and is clean so this is just a nit.

gsanchezgavier (commented):
thanks! i just left some minor comments

gsanchezgavier:
> @gsanchezgavier The changelog should be modified with the last PR right?

yes this  is the Pr that actually adds the remote agent types support for Agent control , we should doublecheck , that the upstream repo and public key are correct (even if they are not being used)

danielorihuela:
> yes this is the Pr that actually adds the remote agent types support for Agent control , we should doublecheck , that the upstream repo and public key are correct (even if they are not being used)

Shouldn't we do it in the E2E PR? Both, make sure everything is correct and modify the changelog? Otherwise, we might be adding something that's not really working on a release, if it's need to be done before the e2e.

agent-control/src/agent_control/run.rs
danielorihuela: I don't know to what are you referring to.

agent-control/tests/on_host/agent_type_composite_registry.rs
danielorihuela: But those are the e2e tests, right?
Isn't that a separate task?

agent-control/src/agent_control/run.rs
gsanchezgavier: in case the signature verification is enabled , the url should not be empty

agent-control/src/agent_type/registry.rs
danielorihuela: I had local + remote struct in my mind, but @sigilioso suggestion was to do it like that for future extensibility. I think it's a bit too much with all the generics and the enum, but it makes testing easier. Specially for the "composite" behaviour.

agent-control/tests/on_host/agent_type_composite_registry.rs
gsanchezgavier: the second one might be 🤔 in fact is already been checked , but i think that we could have scenario integration tests for the first one instead of this

gsanchezgavier:
> > yes this is the Pr that actually adds the remote agent types support for Agent control , we should doublecheck , that the upstream repo and public key are correct (even if they are not being used)
> 
> Shouldn't we do it in the E2E PR? Both, make sure everything is correct and modify the changelog? Otherwise, we might be adding something that's not really working on a release, if it's need to be done before the e2e.

I think having an integration test scenario would be ok for the current status, and give us enough confidence to state that this works. An e2e right know would make use of the same fakers (opamp + registry) as the scenario with a much higher running cost .

agent-control/src/agent_control/run.rs
danielorihuela: You are right. Solved in https://github.com/newrelic/newrelic-agent-control/pull/2602/changes/95f20de9ba4ae462b9192b2ad8e49c467a934a87.

I decided to throw an error because it's an undesired state. Having signature enabled without any public key url.

agent-control/tests/on_host/agent_type_composite_registry.rs
danielorihuela: Done. I "moved" them to scenario tests. So now, we are not just testing the isolated behaviour but that agent control is actually behaving like that.

It's a pity that now they take way longer, I tried to make them as fast as possible.

danielorihuela: https://github.com/newrelic/newrelic-agent-control/pull/2602/changes/9be5ceaa27164f82511f6c28d08d48dc98911b18

agent-control/src/agent_control/run.rs
```diff
@@ -2,16 +2,23 @@ use super::defaults::{
...
 use crate::command::RunnerContext;
 use crate::data_store::DataStore;
 use crate::event::broadcaster::unbounded::UnboundedBroadcast;
 use crate::event::{AgentControlEvent, ApplicationEvent, SubAgentEvent, channel::EventConsumer};
+use crate::oci;
 use crate::opamp::remote_config::validators::signature::validator::SignatureValidator;
 use crate::values::ConfigRepo;
+use oci_client::client::ClientConfig;
+#[cfg(debug_assertions)]
+use oci_client::client::ClientProtocol;
```
gsanchezgavier: nit: i would remove this from here and just use the entire reference when used, just to avoid the #[cfg pollution

agent-control/tests/on_host/scenarios/install_agent_type.rs
```diff
@@ -0,0 +1,271 @@
...
+    let remote_dir = tempdir().unwrap();
+
+    BasePaths {
+        local_dir: local_dir.path().to_path_buf(),
+        remote_dir: remote_dir.path().to_path_buf(),
+        log_dir: local_dir.path().to_path_buf(),
+    }
+}
+
+fn start_agent_control_for_test(
```
gsanchezgavier, line 75: nit: aren't already some similar helpers in common places that lets you inject extra ac config?

agent-control/tests/on_host/scenarios/install_agent_type.rs
```diff
@@ -0,0 +1,271 @@
...
+        &base_paths,
+        agent_type_version,
+        REMOTE_VERSION_CHECKER_OUTPUT,
+    );
+    assert_agent_control_still_running(&mut agent_control);
+}
+
+#[test]
+#[ignore = "needs oci registry (use *with_oci_registry suffix)"]
+fn test_local_agent_type_shadows_remote_registry_with_oci_registry() {
```
gsanchezgavier: wouldn't make sense to test this on k8s too?

agent-control/tests/on_host/scenarios/install_agent_type.rs
```diff
@@ -0,0 +1,271 @@
...
+                OPAMP_AGENT_VERSION_ATTRIBUTE_KEY,
+                Value::StringValue(expected_version.to_string()),
+            )]),
+        )
+        .map_err(|e| e.into())
+    });
+
+    retry(30, Duration::from_secs(1), || {
+        check_latest_health_status_was_healthy(opamp_server, &sub_agent_instance_id)
+    });
```
gsanchezgavier: i would remove this check as it is out of scope for the test , unless i'm missing something

agent-control/tests/on_host/scenarios/install_agent_type.rs
```diff
@@ -0,0 +1,271 @@
...
+fn assert_sub_agent_reports_version(
+    opamp_server: &mut FakeServer,
+    base_paths: &BasePaths,
+    agent_type_version: &str,
+    expected_version: &str,
+) {
+    let ac_instance_id = get_instance_id(&AgentID::AgentControl, base_paths.clone());
+    let agent_type_id = agent_type_id(agent_type_version);
+
+    opamp_server.set_config_response(
```
gsanchezgavier, line 139: the AC agent type validation should be triggered here and we should expect a status response ok , it might be worth to add that check, i think that is when actually the AT is downloaded

agent-control/tests/on_host/scenarios/install_agent_type.rs
```diff
@@ -0,0 +1,271 @@
...
+    let base_paths = temp_base_paths();
+    let mut agent_control = start_agent_control_for_test(&opamp_server, &signer, &base_paths);
+
+    assert_sub_agent_reports_version(
+        &mut opamp_server,
+        &base_paths,
+        agent_type_version,
+        REMOTE_VERSION_CHECKER_OUTPUT,
+    );
+    assert_agent_control_still_running(&mut agent_control);
```
gsanchezgavier, line 45: why we have this?

agent-control/tests/on_host/scenarios/install_agent_type.rs
```diff
@@ -0,0 +1,271 @@
...
+  {AGENT_ID}:
+    agent_type: "{agent_type_id}"
+"#
+        ),
+    );
+
+    let sub_agent_id = AgentID::try_from(AGENT_ID).unwrap();
+    let sub_agent_instance_id = get_instance_id(&sub_agent_id, base_paths.clone());
+
+    create_local_config(
```
gsanchezgavier: this is most probably needed now but i remember that at some point , if there was no required variables, you could launch an agent without having the config, it might worth to check it unless you already did

gsanchezgavier (commented):
thanks for adding the tests! i just left some minor questions

agent-control/src/agent_control/run.rs
danielorihuela: Isn't this needed for testing? I copy pasted it.

```rust
let oci_client_config = ClientConfig {
    #[cfg(debug_assertions)]
    protocol: ClientProtocol::HttpsExcept(vec![OCI_TEST_REGISTRY_URL.to_string()]),
    ..Default::default()
};
```

agent-control/tests/on_host/scenarios/install_agent_type.rs
danielorihuela: This function creates the local agent control config and starts it.
In other file tests, this is done directly on the test.
Also, each config is slightly different.

I think there's room for improvement, I wanted to take a look in a different PR once this is merged.

agent-control/tests/on_host/scenarios/install_agent_type.rs
danielorihuela: I thought about that, but I ended adding it to on-host because the code is the same.
I will look into replicating it in k8s if it's easy.

agent-control/src/agent_control/run.rs
gsanchezgavier: yes, i mean using 
```rust
let oci_client_config = ClientConfig {
    #[cfg(debug_assertions)]
    protocol: oci_client::client::ClientProtocol::HttpsExcept(vec![OCI_TEST_REGISTRY_URL.to_string()]),
    ..Default::default()
};
```
for instance so we don't need to include the 
```
#[cfg(debug_assertions)]
use oci_client::client::ClientProtocol;
```

gsanchezgavier: so all the hack stays together in 2 lines

agent-control/tests/on_host/scenarios/install_agent_type.rs
danielorihuela: Not needed, I copied it from somewhere else and I thought it wouldn't hurt.

agent-control/tests/on_host/scenarios/install_agent_type.rs
danielorihuela: You mean the `REMOTE_VERSION_CHECKER_OUTPUT`? This is to know which agent type I got.
Or you mean `assert_agent_control_still_running`? I thought it wouldn't hurt to double check that agent control didn't unexpectedly die. We can remove it. We are only interested in knowing which agent type we got really.

agent-control/src/agent_control/run.rs
danielorihuela: done

agent-control/tests/on_host/scenarios/install_agent_type.rs
danielorihuela: At first I didn't add it and got

```
thread 'on_host::scenarios::install_agent_type::test_local_agent_type_shadows_remote_registry_with_oci_registry' (7462051) panicked at agent-control/tests/common/retry.rs:22:39:
retry failed after 60 attempts: Identifying attributes missing required elements: Required attribute key 'agent.version' with value 'Some("1.0.0")' not found in actual attributes.
```

danielorihuela: The response I set is empty, so we end up in

```rust
Ok(None) => self.reset_to_local_config(&config, old_supervisor, opamp_client),
```

In that situation, we need the local config.

danielorihuela: We can do this instead, but it might be confusing.

```rust
opamp_server.set_config_response(
    sub_agent_instance_id.clone(),
    "{new_config: true}".to_string(),
);
```

agent-control/tests/on_host/scenarios/install_agent_type.rs
danielorihuela: done

danielorihuela: I don't think it's super important, but it doesn't hurt.

agent-control/tests/on_host/scenarios/install_agent_type.rs
danielorihuela: I went with the latter. One less thing to write in a file.

agent-control/tests/on_host/scenarios/install_agent_type.rs
danielorihuela: Removed.

agent-control/src/agent_type/registry.rs
```diff
@@ -43,37 +46,104 @@ pub struct RegistryConfig {
...
+                            debug!(
+                                agent_type_id = %agent_type_id,
+                                error = %err,
+                                "Agent type definition not found on registry layer \"{index}\"; falling through to the next layer",
+                            );
+                        }
+                        _ => warn!(
+                            agent_type_id = %agent_type_id,
+                            error = %err,
+                            "Agent type definition error on registry layer \"{index}\"; falling through to the next layer",
```
gsanchezgavier, line 126: i think the layer thing is ok to say in debug logs but in the future i would refactor this so is clear which registry is failing local or remote

gsanchezgavier (approved):
thanks! as we discussed we could iterate a bit in future PRs to improve the integration test helpers

agent-control/src/agent_type/registry.rs
danielorihuela: I thought about adding a new function to the trait of the registry that forces you to give the registry a name, but I didn't want to do it know, so I went with the index thingy.

danielorihuela: The name could be based on a url or something.

PR #2604, opened 2026-06-15
feat: add registry caching decorator (NR-568044)

gsanchezgavier:
## Summary

Adds a `CachingRegistry` decorator that memoizes successful agent type lookups for the process lifetime. Agent type artifacts are immutable per `AgentTypeID` (name+version embedded in the OCI tag), so repeated resolutions — especially remote downloads — can be served from an in-memory cache.

## Changes

- Add `CachingRegistry<R>` wrapping any `AgentTypeRegistry`, caching only successful lookups.
- Errors propagate uncached so transient failures / not-yet-published types are retried.
- Tidy `RemoteRegistry` doc comment to reference `[AgentTypeTag]`.

## Notes for reviewers

- Not yet wired into the composite `Registry` (precedence TODO still open) — standalone decorator for now.
- On top of #2602

### DIFF
--- a/agent-control/src/agent_type/registry/caching.rs
+++ b/agent-control/src/agent_type/registry/caching.rs
@@ -1,0 +1,116 @@
+use std::collections::HashMap;
+use std::sync::RwLock;
+
+use super::{AgentTypeRegistry, AgentTypeRegistryError};
+use crate::agent_type::agent_type_id::AgentTypeID;
+use crate::agent_type::definition::AgentTypeDefinition;
+
+/// An [AgentTypeRegistry] decorator that memoizes successful lookups from an inner registry.
+///
+/// Agent type artifacts are immutable for a given [AgentTypeID] (its tag embeds name and version),
+/// so a definition is cached for the process lifetime once resolved. Only successful lookups are
+/// cached: errors propagate untouched and are retried on the next call, avoiding pinning a transient
+/// failure or a not-yet-published agent type.
+pub struct CachingRegistry<R> {
+    inner: R,
+    cache: RwLock<HashMap<AgentTypeID, AgentTypeDefinition>>,
+}
+
+impl<R: AgentTypeRegistry> CachingRegistry<R> {
+    pub fn new(inner: R) -> Self {
+        Self {
+            inner,
+            cache: RwLock::new(HashMap::new()),
+        }
+    }
+}
+
+impl<R: AgentTypeRegistry> AgentTypeRegistry for CachingRegistry<R> {
+    fn get(
+        &self,
+        agent_type_id: &AgentTypeID,
+    ) -> Result<AgentTypeDefinition, AgentTypeRegistryError> {
+        if let Some(definition) = self
+            .cache
+            .read()
+            .expect("agent type cache lock poisoned")
+            .get(agent_type_id)
+        {
+            return Ok(definition.clone());
+        }
+
+        // The inner lookup runs without holding the lock so a slow source (e.g. a remote download)
+        // does not block other readers. A concurrent cold miss may fetch twice and store the same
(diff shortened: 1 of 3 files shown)


alvarocabanas (dismissed):
LGTM

agent-control/src/agent_type/registry/caching.rs
```diff
@@ -0,0 +1,116 @@
...
+use crate::agent_type::agent_type_id::AgentTypeID;
+use crate::agent_type::definition::AgentTypeDefinition;
+
+/// An [AgentTypeRegistry] decorator that memoizes successful lookups from an inner registry.
+///
+/// Agent type artifacts are immutable for a given [AgentTypeID] (its tag embeds name and version),
+/// so a definition is cached for the process lifetime once resolved. Only successful lookups are
+/// cached: errors propagate untouched and are retried on the next call, avoiding pinning a transient
+/// failure or a not-yet-published agent type.
+pub struct CachingRegistry<R> {
```
DavSanchez: What would `R` be?

DavSanchez: I ask in case we want to constrain this to `AgentTypeRegistry` as well, given the type seems `pub`lic.

gsanchezgavier: yes  makes sense to add the constrain 👍

agent-control/src/agent_type/registry/caching.rs
```diff
@@ -0,0 +1,116 @@
+use std::collections::HashMap;
+use std::sync::RwLock;
+
+use super::{AgentTypeRegistry, AgentTypeRegistryError};
+use crate::agent_type::agent_type_id::AgentTypeID;
+use crate::agent_type::definition::AgentTypeDefinition;
+
+/// An [AgentTypeRegistry] decorator that memoizes successful lookups from an inner registry.
```
danielorihuela, line 8: Have you thought of a separate `CacheRegistry`. It would do less network calls, it would be easier to extend in the future (e.g. store cache in the filesystem) and we have logic centralized.

Maybe it doesn't make sense, because I don't know what should be the behaviour if suddenly a registry with "higher" priority uploads the agent type in a different shape.

Example:

```
Local (miss), Remote 1 (miss), Remote 2 (miss), ..., Remote n (hit)
```

With the cache decorator, if we want to get an agent type from the last registry, we need to traverse the whole list and launch network calls.

With a separate`CacheRegistry`, we can avoid this network calls and centralize all the logic if we ever want to store them somewhere else (e.g. the filesystem, or k8s resources).

```
Local (miss), CacheRegistry (hit), Remote 1 (miss), Remote 2 (miss), ..., Remote n (hit)
```

It would be a matter of adding the retrieved agent type to the cache on the "composite registry". It could be a field next to `registries: Vec<R>`.

gsanchezgavier: not sure if i follow, there are two registries, the local which makes no network calls and the remote which does and is the one being cached. I don't expect we have several remote registries in line.

danielorihuela: > I don't expect we have several remote registries in line.

Okay. My question was about that. If you have thought about having a separate `Cache` as a thing, not as a decorator. Because that has some improvements on the multiple remote registries scenario.

gsanchezgavier: forgot to answer that specifically
> If you have thought about having a separate Cache as a thing

I remember discussing it with @sigilioso and IIRC we thought that a separate cache entity would required more work to populate it from other registries and since the only registry that make sense to have a cache was the remote one we lean for this simpler solution.

danielorihuela (approved):

PR #2606, opened 2026-06-16
feat: delete version command and get it from oci package for onhost

vjripoll:
## Summary

Replaces command-based agent version checking with OCI package version as the single source of truth. This eliminates fragile regex parsing and fixes incorrect version reporting in custom builds.

## Problem
Previously, Agent Control determined agent versions by executing commands and parsing output with regex:

  ```yaml
  deployment:
    version:
      path: /usr/bin/newrelic-infra
      args: ["--version"]
      regex: \d+\.\d+\.\d+
  ```

Issues:
  - Custom builds returned incorrect versions (e.g., Go version 1.25.6 instead of agent version)
  - Fragile regex parsing could break with output changes
  - Two sources of truth: OCI package version vs command output
  - Unnecessary process execution overhead

Solution
Use the OCI package version directly from packages_config - the version we downloaded is the version we report:

  ```
  let version = self.packages_config
      .values()
      .next()
      .map(|pkg| pkg.download.oci.version.to_string());
  ```

Changes

Code:
  - supervisor.rs: Replaced check_subagent_version() to read from OCI package config
  - builder.rs: Removed version_config parameter
  - on_host/rendered.rs: Removed version field from OnHost struct
  - on_host.rs: Removed version config templating logic
  - Removed unused imports: OnHostVersionConfig, OnHostAgentVersionChecker

Agent Type YAMLs (6 files):
  Removed deployment.version section from:
  - host-linux-com.newrelic.infrastructure-0.1.0.yaml
  - host-windows-com.newrelic.infrastructure-0.1.0.yaml
  - host-linux-com.newrelic.ebpf-0.1.0.yaml
  - host-linux-com.newrelic.opentelemetry.collector-0.1.0.yaml
  - host-linux-io.opentelemetry.collector-0.1.0.yaml
  - host-windows-com.newrelic.opentelemetry.collector-0.1.0.yaml


agent-control/tests/on_host/scenarios/attributes.rs
```diff
@@ -116,35 +121,74 @@ fn test_attributes_from_non_existing_agent_type() {
...
-/// the "agent.version" related with the agent type.
-#[rstest]
-#[case::with_regex(|local_dir| {CustomAgentType::default().build(local_dir)})]
-#[cfg_attr(target_family = "unix", case::without_regex(|local_dir| {CustomAgentType::default().with_version(Some(r#"{"path": "echo", "args": ["-n","1.0.0"]}"#)).build(local_dir)}))]
-#[cfg_attr(target_family = "windows", case::without_regex(|local_dir| {CustomAgentType::default().with_version(Some(r#"{"path": "cmd", "args": ["/C","set","/p=1.0.0<nul"]}"#)).build(local_dir)}))]
-fn test_attributes_from_an_existing_agent_type(#[case] get_agent_type: impl Fn(PathBuf) -> String) {
-    let opamp_server = FakeServer::start(tokio_runtime().handle());
+/// identifying and non identifying attributes are what we expect, including agent.version.
+#[test]
+#[ignore = "needs oci registry (use *with_oci_registry suffix)"]
```
danielorihuela, line 113: The ignore text is a lie haha. The test doesn't need an oci registry.

danielorihuela: That or the test is missing the `_with_oci_registry` at the end and it's not running on the CI pipeline

vjripoll: yep thats true something i left from the first iteration

agent-control/src/agent_type/runtime_config/on_host/rendered.rs
```diff
@@ -13,7 +12,6 @@ pub struct OnHost {
     pub enable_file_logging: bool,
     /// Enables and define health checks configuration.
     pub health: OnHostHealthConfig,
-    pub version: Option<OnHostVersionConfig>,
```
gsanchezgavier, line 21: could be the case of agents that doesn't have a package but somehow they compute a version?  i fear that this is a feature that we might need to add , and so I'm not sure if we should remove it. would be enough just to remove the usage from the current agent types

vjripoll: To confirm: remove version from existing agent types but keep the code infrastructure, so future agent types without OCI packages can still use it if needed?

agent-control/tests/on_host/scenarios/attributes.rs
```diff
@@ -113,31 +117,60 @@ fn test_attributes_from_non_existing_agent_type() {
...
+test-package:
+  type: {package_type}
+  download:
+    oci:
+      repository: test
+      version: ${{nr-var:version}}
+"#
+    );
+
+    let sleep_agent_type = CustomAgentType::default()
```
danielorihuela, line 178: Nit: Maybe `sleep_agent_type` is not the best name

agent-control/tests/on_host/scenarios/install_agent_type.rs
```diff
@@ -97,21 +98,67 @@ fn push_agent_type_to_registry(signer: &OCISigner) -> oci_client::Reference {
...
+        &agent_type_definition_yaml("linux", "tar"),
         &format!("{tag}.yaml"),
     );
 
     let reference = PackagePublisher::new(tokio_runtime().handle().clone(), OCI_TEST_REGISTRY_URL)
         .push_with_tag(&archive, AgentTypeArtifact, tag.as_str());
 
     signer.sign_artifact(&reference);
+
+    push_dummy_test_package(signer, REPORTED_VERSION, OCI_TEST_REGISTRY_URL);
```
danielorihuela: There's something odd here. We do a `push_with_tag` on 106 and then a `push_dummy_test_package` on 110. Why?

1. Could we remove the `REPORTED_VERSION` and use `AGENT_TYPE_VERSION` instead?
2. Could we remove `push_dummy_test_package`?
3. Could we put `#[cfg(target_family = "unix")]` on top of `TestDataHelper::compress_tar_gz(`, so that we avoid mostly copy pasting the same function for windows?

agent-control/tests/on_host/scenarios/install_agent_type.rs
```diff
@@ -97,21 +98,67 @@ fn push_agent_type_to_registry(signer: &OCISigner) -> oci_client::Reference {
...
+#[cfg(target_family = "windows")]
+fn push_agent_type_to_registry(signer: &OCISigner) -> oci_client::Reference {
+    let tag = AgentTypeTag::new(
+        &agent_type_id(AGENT_TYPE_VERSION),
+        AGENT_CONTROL_MODE_ON_HOST,
+    );
+
+    let source_dir = tempdir().unwrap();
+    let archive_dir = tempdir().unwrap();
+    let archive = archive_dir.path().join("agent-type.tar.gz");
```
danielorihuela, line 115: For windows, should this be a `zip`?

agent-control/tests/on_host/scenarios/install_agent_type.rs
```diff
@@ -168,40 +218,28 @@ fn agent_type_id(version: &str) -> AgentTypeID {
     AgentTypeID::try_from(id.as_str()).unwrap()
 }
 
-#[cfg(not(target_os = "windows"))]
-fn agent_type_definition_yaml() -> String {
+fn agent_type_definition_yaml(operating_system: &str, package_type: &str) -> String {
```
danielorihuela: Would it make sense to hide the windows and linux details here in this function?

```rust
# cfg unix
let os = linux
# cfg unix
let pkg_type = zip

# cfg windows
...
```

That way we minimize the number of places where we use compile flags. I don't know if it makes it more obscure.

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -243,27 +241,36 @@ where
...
-            // Using an enum variant that wraps a type is the same as a function taking the type.
-            // Basically, it's the same as passing "|x| SubAgentInternalEvent::AgentAttributesUpdated(x)"
-            SubAgentInternalEvent::AgentAttributesUpdated,
-        )
+        // Get version from OCI package configuration
+        let version = self
+            .packages_config
+            .values()
+            .next()
+            .map(|pkg| pkg.download.oci.version.to_string());
```
danielorihuela: Hashmap order is random. If a package has multiple binaries, the version won't be the version from the first package.

That's okay for now, but as soon as we introduce more complex OHIs, I don't think we want that.

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -243,27 +241,36 @@ where
...
-            onhost_version_checker,
-            sub_agent_internal_publisher,
-            // The below argument expects a function "UpdateAttributesMessage -> T"
-            // where T is the "event" sendable by the above publisher.
-            // Using an enum variant that wraps a type is the same as a function taking the type.
-            // Basically, it's the same as passing "|x| SubAgentInternalEvent::AgentAttributesUpdated(x)"
-            SubAgentInternalEvent::AgentAttributesUpdated,
-        )
+        // Get version from OCI package configuration
+        let version = self
```
danielorihuela: What happens if the configured version is a hash? Is it okay to show it instead of the semver version? Assuming the binary has a semver version.

vjripoll: Yes, if oci.version is a digest (e.g., sha256:...), we'll report that as agent.version. This is intentional - we're treating the configured oci.version as the source of truth. 
OpAMP accepts any string value, and digests are actually more precise than semver for tracking exact artifacts.

agent-control/tests/on_host/scenarios/install_agent_type.rs
vjripoll: No, agent types are always .tar.gz on all platforms (including Windows).

agent-control/tests/on_host/scenarios/attributes.rs
```diff
@@ -109,15 +111,46 @@ fn test_attributes_from_non_existing_agent_type() {
...
 #[rstest]
-#[case::with_regex(|local_dir| {CustomAgentType::default().build(local_dir)})]
-#[cfg_attr(target_family = "unix", case::without_regex(|local_dir| {CustomAgentType::default().with_version(Some(r#"{"path": "echo", "args": ["-n","1.0.0"]}"#)).build(local_dir)}))]
-#[cfg_attr(target_family = "windows", case::without_regex(|local_dir| {CustomAgentType::default().with_version(Some(r#"{"path": "cmd", "args": ["/C","set","/p=1.0.0<nul"]}"#)).build(local_dir)}))]
-fn test_attributes_from_an_existing_agent_type(#[case] get_agent_type: impl Fn(PathBuf) -> String) {
+#[case::with_regex(false)]
+#[cfg_attr(target_family = "unix", case::without_regex(true))]
+#[cfg_attr(target_family = "windows", case::without_regex(true))]
+#[ignore = "needs oci registry (use *with_oci_registry suffix)"]
+fn test_attributes_from_an_existing_agent_type(#[case] use_version_command: bool) {
```
danielorihuela: We lost the `_with_oci_registry`. Thist test is not being executed.

agent-control/tests/on_host/scenarios/install_agent_type.rs
```diff
@@ -50,54 +49,154 @@ fn test_local_agent_type_shadows_remote_registry_with_oci_registry() {
...
+    archive
+}
 
-    let reference = PackagePublisher::new(tokio_runtime().handle().clone(), OCI_TEST_REGISTRY_URL)
-        .push_with_tag(&archive, AgentTypeArtifact, tag.as_str());
-
-    signer.sign_artifact(&reference);
-    reference
+#[cfg(target_family = "windows")]
+fn create_agent_type_archive(
```
danielorihuela, line 108: `create_agent_type_archive` is the exact same in windows and linux. I think we can get rid of one of them.

agent-control/tests/on_host/scenarios/attributes.rs
```diff
@@ -109,15 +111,46 @@ fn test_attributes_from_non_existing_agent_type() {
...
+    let mut sleep_agent_type_builder = CustomAgentType::default().with_packages(Some(&packages));
+
+    // Optionally add version command (though it won't be used for on-host with OCI packages)
+    if use_version_command {
+        #[cfg(target_family = "unix")]
+        let version_config = r#"{"path": "echo", "args": ["-n","1.0.0"]}"#;
+        #[cfg(target_family = "windows")]
+        let version_config = r#"{"path": "cmd", "args": ["/C","set","/p=1.0.0<nul"]}"#;
+        sleep_agent_type_builder = sleep_agent_type_builder.with_version(Some(version_config));
+    }
```
danielorihuela: Shouldn't we remove the `with_version` if we are removing the version thing from the codebase?

agent-control/tests/on_host/scenarios/attributes.rs
```diff
@@ -109,15 +111,46 @@ fn test_attributes_from_non_existing_agent_type() {
...
+      repository: test
+      version: ${{{{nr-var:fake_variable}}}}
+      public_key_url: {}
+"#,
+        signer.jwks_url()
+    );
+
     // Add custom agent_type to registry
-    let sleep_agent_type = get_agent_type(dirs.local_dir());
+    let mut sleep_agent_type_builder = CustomAgentType::default().with_packages(Some(&packages));
```
danielorihuela: We can call it `custom_agent_type_builder`

agent-control/tests/on_host/scenarios/attributes.rs
```diff
@@ -109,15 +111,46 @@ fn test_attributes_from_non_existing_agent_type() {
 /// identifying and non identifying attributes are what we expect plus
 /// the "agent.version" related with the agent type.
 #[rstest]
-#[case::with_regex(|local_dir| {CustomAgentType::default().build(local_dir)})]
-#[cfg_attr(target_family = "unix", case::without_regex(|local_dir| {CustomAgentType::default().with_version(Some(r#"{"path": "echo", "args": ["-n","1.0.0"]}"#)).build(local_dir)}))]
-#[cfg_attr(target_family = "windows", case::without_regex(|local_dir| {CustomAgentType::default().with_version(Some(r#"{"path": "cmd", "args": ["/C","set","/p=1.0.0<nul"]}"#)).build(local_dir)}))]
-fn test_attributes_from_an_existing_agent_type(#[case] get_agent_type: impl Fn(PathBuf) -> String) {
+#[case::with_regex(false)]
+#[cfg_attr(target_family = "unix", case::without_regex(true))]
+#[cfg_attr(target_family = "windows", case::without_regex(true))]
```
danielorihuela: We don't have regex anymore, do we?

agent-control/tests/on_host/scenarios/attributes.rs
danielorihuela: Reuse `version` variable here

agent-control/tests/on_host/scenarios/attributes.rs
```diff
@@ -109,15 +111,46 @@ fn test_attributes_from_non_existing_agent_type() {
...
+
+    // Create packages config with public key URL
+    let packages = format!(
+        r#"
+nr-sleep-agent:
+  type: tar
+  download:
+    oci:
+      repository: test
+      version: ${{{{nr-var:fake_variable}}}}
```
danielorihuela: There's no need for a variable here, you can use `version` directly if you want to simplify the test a bit.

agent-control/tests/on_host/scenarios/install_agent_type.rs
```diff
@@ -50,54 +49,154 @@ fn test_local_agent_type_shadows_remote_registry_with_oci_registry() {
 #[test]
 #[ignore = "needs oci registry (use *with_oci_registry suffix)"]
 fn test_local_miss_resolves_via_remote_registry_with_oci_registry() {
+    eprintln!("INFO: Starting test_local_miss_resolves_via_remote_registry_with_oci_registry");
```
danielorihuela: Are these leftovers from your debugging, or you want to leave them here for the following person?

vjripoll: yep 💩 debug logs 👅

agent-control/tests/on_host/scenarios/attributes.rs
vjripoll: we have in code lot of test with sleep_ganet should we change all of them too?

danielorihuela: ah, if its something general, let's keep as it is

agent-control/tests/on_host/scenarios/attributes.rs
```diff
@@ -108,41 +108,105 @@ fn test_attributes_from_non_existing_agent_type() {
...
+        }}
+    ]"#,
+            script_path
+        )
+    };
+
+    // Add custom agent_type to registry
+    let sleep_agent_type = CustomAgentType::default()
+        .with_executables(Some(&executables))
+        .with_version(None) // Remove version checker - version comes from OCI package
```
danielorihuela: > Shouldn't we remove the with_version if we are removing the version thing from the codebase?

I said that in another comment, what I meant is if we should remove the functionality from the `CustomAgentType` struct, given that it's no longer used.

vjripoll: true 🧹

agent-control/tests/on_host/scenarios/self_instrumentation_otel.rs
```diff
@@ -65,13 +65,7 @@ self_instrumentation:
...
     // The binary runs until a timeout kills it.
     let mut cmd = cmd_with_config_file(dir.path());
-    let output = cmd.output().expect("running newrelic-agent-control binary");
-
-    let ac_stdout = String::from_utf8_lossy(&output.stdout);
-    let ac_stderr = String::from_utf8_lossy(&output.stderr);
-
-    eprintln!("--- agent-control stdout ---\n{ac_stdout}");
-    eprintln!("--- agent-control stderr ---\n{ac_stderr}");
+    let _output = cmd.output().expect("running newrelic-agent-control binary");
```
danielorihuela: This change is not from this PR, is it?

agent-control/src/agent_type/README.md
```diff
@@ -243,21 +237,28 @@ health:
...
+The resolved value of that `version` field is published as `agent.version`. When an agent type
+defines more than one package, the version of the first package (ordered by package id) is
+reported. See [On Host Packages](#on-host-packages) for the full package configuration.
 
-* `path`: Specifies the binary to run.
-* `args`: Specifies the arguments passed to the binary to get the version.
-* `regex`: Optional field that specifies the regular expression to get the version from the output. When not used, the whole output will be used as the version.
+> **Note:** Earlier releases used a `deployment.version` block with `path`/`args`/`regex` to run a
+> command and parse its output. That mechanism has been removed for on-host agents and the
+> `deployment.version` field is no longer supported.
```
danielorihuela: Nit: Given that's an internal detail, maybe there's no need to mention it.

danielorihuela (dismissed):
Left two minor comments. Maybe they don't make sense.
Other than that, great work!!!

agent-control/tests/on_host/scenarios/self_instrumentation_otel.rs
vjripoll: Good catch — that was an unrelated leftover, not part of this PR. Reverted.

agent-control/src/agent_type/README.md
vjripoll: Agreed, it's an internal detail. Removed the note.

danielorihuela (dismissed):

agent-control/src/agent_type/runtime_config/on_host/rendered.rs
paologallinaharbur: I think we decided offline not to keep the code, I am closing this

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -243,27 +241,59 @@ where
...
-            // Basically, it's the same as passing "|x| SubAgentInternalEvent::AgentAttributesUpdated(x)"
-            SubAgentInternalEvent::AgentAttributesUpdated,
-        )
+        // Get version from OCI package configuration
+        // TODO: For complex OHIs with multiple packages, we need a better strategy
+        // to determine which version to report (e.g., mark a primary package in agent type definition)
+        let version = self
+            .packages_config
+            .iter()
+            .min_by_key(|(id, _)| *id)
```
paologallinaharbur: why are we doing this? is just a way to select a random one right?

Is it the most similar thing to a [0] for an array?

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -243,27 +241,59 @@ where
...
-            onhost_version_checker,
-            sub_agent_internal_publisher,
-            // The below argument expects a function "UpdateAttributesMessage -> T"
-            // where T is the "event" sendable by the above publisher.
-            // Using an enum variant that wraps a type is the same as a function taking the type.
-            // Basically, it's the same as passing "|x| SubAgentInternalEvent::AgentAttributesUpdated(x)"
-            SubAgentInternalEvent::AgentAttributesUpdated,
-        )
+        // Get version from OCI package configuration
+        // TODO: For complex OHIs with multiple packages, we need a better strategy
```
paologallinaharbur: maybe we can add a warning for now if there is more than one package

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -243,27 +241,59 @@ where
...
+                        version.into(),
+                    )]),
+                    ..Default::default()
+                }),
+            );
+        } else {
+            warn!(
+                agent_type=%self.agent_identity.agent_type_id,
+                "Unable to determine agent version: no packages configured"
+            );
```
paologallinaharbur: Style, instead of moving around an option we could exit early at the beginning.

we could check

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -243,27 +241,59 @@ where
...
+        }
     }
```
paologallinaharbur: I believe something like this would be easier to read
```rust
        // Get version from OCI package configuration
        // TODO: For complex OHIs with multiple packages, we need a better strategy
        // to determine which version to report (e.g., mark a primary package in agent type definition)
        let Some((id, pkg)) = self.packages_config.iter().min_by_key(|(id, _)| *id) else {
            warn!(
                agent_type=%self.agent_identity.agent_type_id,
                "Unable to determine agent version: no packages configured"
            );
            return;
        };
        
        let version = pkg.download.oci.version.to_string();

        debug!(
            agent_type=%self.agent_identity.agent_type_id,
            package_id=%id,
            version=%version,
            "Found package version from OCI config"
        );

        info!(
            agent_type=%self.agent_identity.agent_type_id,
            version=%version,
            "Agent version determined from OCI package"
        );

        info!(
            agent_type=%self.agent_identity.agent_type_id,
            version=%version,
            "Publishing AgentAttributesUpdated event"
        );

        publish_update_attributes_event(
            &sub_agent_internal_publisher,
            SubAgentInternalEvent::AgentAttributesUpdated(AgentDescription {
                identifying_attributes: HashMap::from([(
                    OPAMP_AGENT_VERSION_ATTRIBUTE_KEY.to_string(),
                    version.into(),
                )]),
                ..Default::default()
            }),
        );
```

agent-control/tests/on_host/scenarios/install_agent_type.rs
```diff
@@ -160,40 +179,35 @@ fn agent_type_id(version: &str) -> AgentTypeID {
...
-    args: ["{REPORTED_VERSION}"]
-    regex: \d+\.\d+\.\d+
-"#
-    )
-}
+    #[cfg(target_family = "unix")]
+    let (operating_system, package_type) = ("linux", "tar");
+
+    #[cfg(target_family = "windows")]
+    let (operating_system, package_type) = ("windows", "zip");
```
paologallinaharbur, line 193: nice!

agent-control/tests/on_host/scenarios/attributes.rs
```diff
@@ -179,21 +244,80 @@ agents:
...
             &opamp_server,
-            &agent_control_instance_id,
+            &sleep_instance_id,
             expected_non_identifying_attributes.clone(),
         )?;
         Ok(())
     })
 }
+
+fn push_test_package_with_sleep_script(signer: &OCISigner, version: &str, registry_url: &str) {
```
paologallinaharbur: can't this be reused as well instead of push_dummy_test_package? it feels like we are duplicating code

paologallinaharbur (dismissed):
Mostly nits, maybe the most "concerning" one is the one about duplicating help functions in test code

danielorihuela (dismissed):
Looks good from my side! Would be great if @paologallinaharbur takes a look since he asked for the last few changes.

danielorihuela (approved):

PR #2610, opened 2026-06-16
feat: add retries with backoff and jitter for oci download

alvarocabanas:
Adds exponential-backoff + jitter retries to self-update OCI downloads via a configurable BackoffPolicy (self_update.download_retry). A new BackoffGate throttles upgrade attempts per target version up to a max-delay cap; once capped, Agent Control keeps retrying rather than giving up. This lets it recover automatically from transient registry outages, and reset when a new version arrives. Covered by a new integration test for outage recovery.

### DIFF
--- a/agent-control/src/agent_control/version_updater/on_host.rs
+++ b/agent-control/src/agent_control/version_updater/on_host.rs
@@ -94,13 +94,15 @@ where
     fn retry(&self) -> Result<(), UpdaterError> {
         // The gate's tracked key is the last version we tried to reach; if it is cleared there is
         // nothing pending, otherwise re-drive the normal `update` path for that version.
-        let Some(version) = self.upgrade_gate.current_key() else {
-            return Ok(());
-        };
-        self.update(&AgentControlDynamicConfig {
-            version: Some(version),
+        let version = self.upgrade_gate.current_key();
+        if version.is_some() {
+          self.update(&AgentControlDynamicConfig {
+            version,
             ..Default::default()
-        })
+          })
+        } else {
+            Ok(())
+        }
     }
 }
 
--- a/agent-control/src/agent_control/version_updater/on_host.rs
+++ b/agent-control/src/agent_control/version_updater/on_host.rs
@@ -82,13 +82,11 @@ where
         // Cooldown gate: suppress re-attempts that are still within their backoff window, until the
         // window elapses (or the desired version changes). When it permits an attempt the gate runs
         // the upgrade and tracks success/failure itself.
-        match self.upgrade_gate.guarded(new_version, || {
+        self.upgrade_gate.guarded(new_version, || {
             debug!("Starting update process");
             self.try_upgrade(new_version.clone())
-        }) {
-            Ok(result) => result,
-            Err(suppression) => Err(self.suppressed_error(new_version, suppression)),
-        }
+        })
+        .map_err(|e| self.suppressed_error(new_version, e))?
     }
 
     fn retry(&self) -> Result<(), UpdaterError> {
--- a/agent-control/src/agent_control/version_updater/on_host.rs
+++ b/agent-control/src/agent_control/version_updater/on_host.rs
@@ -1,16 +1,23 @@
 pub mod verify;
 
 use crate::agent_control::agent_id::AgentID;
-use crate::agent_control::config::{AgentControlDynamicConfig, AgentControlPackage};
+use crate::agent_control::config::{
+    AgentControlDynamicConfig, AgentControlPackage, UpgradeBackoffConfig,
+};
 use crate::agent_control::defaults::AGENT_CONTROL_VERSION;
-use crate::agent_control::version_updater::updater::{UpdaterError, VersionUpdater};
+use crate::agent_control::version_updater::updater::{
+    CooldownReason, UpdaterError, VersionUpdater,
+};
 use crate::agent_type::runtime_config::on_host::package::rendered::{Oci, Repository, Version};
 use crate::event::AgentControlInternalEvent;
 use crate::event::channel::EventPublisher;
 use crate::package::manager::{PackageData, PackageManager};
+use crate::utils::backoff_gate::{BackoffGate, GateDecision};
+use crate::utils::retry::BackoffPolicy;
+use crate::utils::time::Clock;
 use self_replacer::{BinarySelfReplacer, SelfReplacer};
 use thiserror::Error;
-use tracing::{debug, debug_span};
+use tracing::{debug, debug_span, warn};
 use url::Url;
 use verify::VerifyExecutor;
 
@@ -27,10 +34,11 @@ pub enum BuildError {
     InvalidReference(#[from] oci_client::ParseError),
 }
 
-pub struct OnHostACUpdater<P, V>
+pub struct OnHostACUpdater<P, V, C>
 where
     P: PackageManager,
     V: VerifyExecutor,
+    C: Clock,
 {
     ac_remote_update_enabled: bool,
     agent_control_internal_publisher: EventPublisher<AgentControlInternalEvent>,
@@ -38,12 +46,16 @@ where
     verify_executor: V,
     repository: Repository,
     pub_key_url: Url,
+    /// Throttles re-attempts at a failing upgrade so we don't hammer the registry every
+    /// OpAMP poll. Keyed by target [`Version`]: a new desired version resets the cooldown.
+    upgrade_gate: BackoffGate<Version, C>,
 }
 
-impl<P, V> VersionUpdater for OnHostACUpdater<P, V>
+impl<P, V, C> VersionUpdater for OnHostACUpdater<P, V, C>
 where
     P: PackageManager,
     V: VerifyExecutor,
+    C: Clock,
 {
     fn update(&self, config: &AgentControlDynamicConfig) -> Result<(), UpdaterError> {
         if !self.ac_remote_update_enabled {
@@ -65,12 +77,97 @@ where
 
         if new_version.to_string() == AGENT_CONTROL_VERSION {
             debug!("Desired version is the same as current, skipping update");
+            self.upgrade_gate.reset();
             return Ok(());
+        }
+
+        // Cooldown gate: suppress re-attempts that are still within their backoff window, or
+        // that have exhausted the consecutive-failure budget, until the desired version changes.
+        if let Some(err) = self.cooldown_error(new_version) {
+            return Err(err);
         }
 
         debug!("Starting update process");
 
-        let package_data = self.get_package_data(new_version.clone());
+        match self.try_upgrade(new_version.clone()) {
+            Ok(()) => {
+                self.upgrade_gate.reset();
+                Ok(())
+            }
+            Err(e) => {
+                self.upgrade_gate.record_failure(new_version);
+                Err(e)
+            }
+        }
+    }
+}
+
+impl<P, V, C> OnHostACUpdater<P, V, C>
+where
+    P: PackageManager,
+    V: VerifyExecutor,
+    C: Clock,
+{
+    pub fn new(
+        ac_remote_update_enabled: bool,
+        agent_control_internal_publisher: EventPublisher<AgentControlInternalEvent>,
+        package_manager: P,
+        verify_executor: V,
+        package: AgentControlPackage,
+        backoff: UpgradeBackoffConfig,
+        clock: C,
+    ) -> Self {
+        Self {
+            ac_remote_update_enabled,
+            agent_control_internal_publisher,
+            package_manager,
+            verify_executor,
+            repository: package.download.oci.repository.clone(),
+            pub_key_url: package.download.oci.public_key_url.clone(),
+            // The gate owns the exponential-backoff-plus-jitter schedule (it never sleeps; it
+            // records a "next attempt" instant checked across OpAMP polls).
+            upgrade_gate: BackoffGate::new(BackoffPolicy::from(&backoff), clock),
+        }
+    }
+
+    /// Maps the generic gate verdict for `new_version` onto the OpAMP-facing [`UpdaterError`],
+    /// returning `Some(err)` when the upgrade should be suppressed and `None` to proceed.
+    fn cooldown_error(&self, new_version: &Version) -> Option<UpdaterError> {
+        match self.upgrade_gate.check(new_version) {
+            GateDecision::Proceed => None,
+            GateDecision::CapReached {
+                consecutive_failures,
+            } => {
+                warn!(
+                    version = %new_version,
+                    consecutive_failures,
+                    "Upgrade suppressed: max consecutive failures reached. \
+                     Waiting for desired version to change before retrying.",
+                );
+                Some(UpdaterError::UpdateInCooldown {
+                    version: new_version.to_string(),
+                    reason: CooldownReason::CapReached,
+                })
+            }
+            GateDecision::InCooldown {
+                consecutive_failures,
+            } => {
+                debug!(
+                    version = %new_version,
+                    consecutive_failures,
+                    "Upgrade suppressed: in backoff cooldown window.",
+                );
+                Some(UpdaterError::UpdateInCooldown {
+                    version: new_version.to_string(),
+                    reason: CooldownReason::Backoff,
+                })
+            }
+        }
+    }
+
+    /// Performs a single upgrade attempt: install → verify → self-replace → request restart.
+    fn try_upgrade(&self, new_version: Version) -> Result<(), UpdaterError> {
+        let package_data = self.get_package_data(new_version);
 
         let new_binary_path = self
             .package_manager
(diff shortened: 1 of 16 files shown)


agent-control/src/agent_control/version_updater/updater.rs
```diff
@@ -6,6 +6,33 @@ use thiserror::Error;
...
+    /// is intentionally **stable across polls** so OpAMP `ConfigState::Failed` does not churn.
+    #[error("upgrade to {version} suppressed: {reason}")]
+    UpdateInCooldown {
+        version: String,
+        reason: CooldownReason,
+    },
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CooldownReason {
```
gsanchezgavier: IMO i would start with a plain retry leveraging the retry.rs util . But if we really want this jitter i would try to make it agnostic inside a the utils module something like 
`pub fn retry_with_jitter<F, T, E>(max_attempts: usize, ..., mut f: F) -> Result<T, E>`

agent-control/src/utils/retry.rs
```diff
@@ -27,9 +27,93 @@ where
...
+    // `0` is treated as `1`.
+    pub max_attempts: usize,
+    pub base_delay: Duration,
+    pub max_delay: Duration,
+    pub jitter: bool,
+}
+
+impl BackoffPolicy {
+    /// Returns the wait time before the `attempt`-th retry, ignoring jitter.
+    pub fn delay_for_attempt(&self, attempt: u32) -> Duration {
```
vjripoll, line 47: i saw that this function is more or less the same than the other one in on_host.rs maybe we can extract it to a helper BackoffPolicy to be used in both places

alvarocabanas: It is only defined here, maybe it was before.

agent-control/src/agent_control/version_updater/updater.rs
alvarocabanas: Modified

agent-control/src/utils/retry.rs
```diff
@@ -27,9 +27,100 @@ where
...
+    Err(last_err.expect("at least one attempt must have produced an error"))
+}
+
+/// Returns a random duration somewhere in `[0, d]`.
+///
+/// Each call to produces a different value so retries on different machines
+/// (and different attempts on the same machine) don't all happen at the same instant.
+/// current system time in nanoseconds is always advancing, and it differs across machine.
+pub fn full_jitter(d: Duration) -> Duration {
+    let cap_nanos = d.as_nanos() as u64;
```
MunhozThiago, line 107: Worth swapping for rand::thread_rng() for non-correlated values?

agent-control/src/agent_control/version_updater/on_host.rs
```diff
@@ -158,40 +232,316 @@ mod tests {
...
-        let updater = new_test_updater(true);
+        let updater = new_test_updater_with(UpgradeBackoffConfig::default(), SystemClock);
         let config = AgentControlDynamicConfig::default();
         assert!(updater.update(&config).is_ok());
     }
 
     #[test]
     fn update_is_noop_when_version_matches_current() {
-        let updater = new_test_updater(true);
+        let updater = new_test_updater_with(UpgradeBackoffConfig::default(), SystemClock);
```
MunhozThiago, line 329: If update() starts reading other config fields (registry, auth, etc.) the retry path will silently use defaults. Can we consider storing the original config or calling try_upgrade() directly?

MunhozThiago (commented):
**Low priority:**

`warn!` in suppressed_error will fire every max_delay indefinitely once the cap is hit — could get noisy in prod. Maybe degrade to info! after the first cap warning.

`BackoffPolicy.max_attempts` is reused with two different semantics (download retries vs. gate failure threshold) — a comment or dedicated field would clarify.

Everything else looks great!

agent-control/src/utils/retry.rs
alvarocabanas: I tried to do it as easy as possible, that's why I used nanoseconds since it's almost impossible for 2 machines to run this in the same nanosecond.

agent-control/src/agent_control/version_updater/on_host.rs
```diff
@@ -65,12 +75,92 @@ where
...
+    fn retry(&self) -> Result<(), UpdaterError> {
+        // The gate's tracked key is the last version we tried to reach; if it is cleared there is
+        // nothing pending, otherwise re-drive the normal `update` path for that version.
+        let Some(version) = self.upgrade_gate.current_key() else {
+            return Ok(());
+        };
+        self.update(&AgentControlDynamicConfig {
+            version: Some(version),
+            ..Default::default()
+        })
```
DavSanchez: This unwraps the option and wraps it again. Probably the compiler will optimize it or something, but also consider something like:

```suggestion
        let version = self.upgrade_gate.current_key();
        if version.is_some() {
          self.update(&AgentControlDynamicConfig {
            version,
            ..Default::default()
          })
        } else {
            Ok(())
        }
```

agent-control/src/agent_control/version_updater/on_host.rs
```diff
@@ -65,12 +75,92 @@ where
...
+        // Cooldown gate: suppress re-attempts that are still within their backoff window, until the
+        // window elapses (or the desired version changes). When it permits an attempt the gate runs
+        // the upgrade and tracks success/failure itself.
+        match self.upgrade_gate.guarded(new_version, || {
+            debug!("Starting update process");
+            self.try_upgrade(new_version.clone())
+        }) {
+            Ok(result) => result,
+            Err(suppression) => Err(self.suppressed_error(new_version, suppression)),
+        }
```
DavSanchez: Does this work?

```suggestion
        self.upgrade_gate.guarded(new_version, || {
            debug!("Starting update process");
            self.try_upgrade(new_version.clone())
        })
        .map_err(|e| self.suppressed_error(new_version, e))?
```

agent-control/src/agent_control/version_updater/on_host.rs
```diff
@@ -158,40 +232,316 @@ mod tests {
         }
     }
 
-    type TestUpdater = OnHostACUpdater<MockPackageManager, MockVerifyExecutorMock>;
+    /// Test clock backed by `Arc<Mutex<Instant>>` so the test body can advance time
+    /// deterministically while the updater holds the clock by value.
+    #[derive(Clone)]
+    struct FakeClock(Arc<std::sync::Mutex<Instant>>);
```
DavSanchez: We can import it with `use std::sync::Mutex;` no?

```suggestion
    struct FakeClock(Arc<Mutex<Instant>>);
```

agent-control/src/agent_control/version_updater/updater.rs
```diff
@@ -1,11 +1,33 @@
...
 /// Represents errors that can occur during the update process of the agent control version.
 #[derive(Debug, Error)]
 pub enum UpdaterError {
     #[error("update failed: {0}")]
     UpdateFailed(String),
+    /// The previous attempt to upgrade to this version failed; we are deliberately not hitting
+    /// the registry again until the cooldown elapses (or the version changes). The message is
+    /// derived from the [`Suppression`] *variant* only (not its failure count), so it is
+    /// intentionally **stable across polls** and OpAMP `ConfigState::Failed` does not churn.
+    #[error("upgrade to {version} suppressed: {}", cooldown_reason(reason))]
```
DavSanchez, line 14: TIL you can pass the result of functions over the error type fields to the `error` attributes! 💥

agent-control/src/agent_control/version_updater/updater.rs
```diff
@@ -20,6 +42,13 @@ pub trait VersionUpdater {
     /// Returns `Ok(())` if the desired version has been successfully communicated
     /// to the external controller, or an `UpdaterError` if the update fails.
     fn update(&self, config: &AgentControlDynamicConfig) -> Result<(), UpdaterError>;
+
+    /// Re-attempts the most recently requested upgrade, if one is still pending. Driven by a
+    /// periodic heartbeat so a transient registry outage recovers without a new desired version
+    /// being pushed. Defaults to a no-op for updaters without a retry concept.
+    fn retry(&self) -> Result<(), UpdaterError> {
+        Ok(())
+    }
```
DavSanchez: Is this gaining more complex logic later or something? In the documentation it says "if one is still pending". How is that evaluated? I assume it happens outside this function so I'm not sure if we should have this around if it's not gaining any complexity soon.

agent-control/src/utils/backoff_gate.rs
```diff
@@ -0,0 +1,307 @@
+use crate::utils::retry::BackoffPolicy;
+use crate::utils::time::{Clock, SystemClock};
+use std::sync::Mutex;
+use std::time::Instant;
+
+/// Why the gate withheld an attempt. This is the suppressed outcome of [`BackoffGate::check`]
+/// (which returns `None` to mean "proceed") and the error type of [`BackoffGate::guarded`], so a
+/// "proceed" case can never appear on the "ran the operation" path.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Suppression {
```
DavSanchez: Naming, perhaps `SuppressionReason`, `SuppressionCause` or something like that. Otherwise it doesn't convey what is the type used for.

agent-control/src/utils/backoff_gate.rs
```diff
@@ -0,0 +1,307 @@
...
+                } else {
+                    Suppression::InCooldown {
+                        consecutive_failures,
+                    }
+                },
+            );
+        }
+
+        // Proceed with the operation (download).
+        None
```
DavSanchez: Maybe a `match` is more readable for the flow?


```suggestion
        match state.next_attempt_at {
          Some(t) if self.clock.now() < t && state.consecutive_failures >= self.policy.max_attempts => Some(Suppression::CapReached { ... }),
          Some(t) if self.clock.now() < t => Some(Suppression::InCooldown { ... }),
          otherwise => None,  // Proceed with the operation (download).
        }
```

Also `filter` by `self.clock.now() < t` and then `map` (or, of course, `and_then` itself xD) to decide on the `Suppression` value can be used.

agent-control/src/utils/backoff_gate.rs
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    /// Records a failed attempt for `key`, incrementing the consecutive-failure count and
+    /// scheduling the next permitted attempt per the backoff schedule.
+    pub fn record_failure(&self, key: &K) {
+        let mut state = self.state.lock().expect("backoff-gate lock poisoned");
```
DavSanchez: I remember this "`expect` on `Mutex`es" coming up on a discussion but not sure if we made a decision against it or not!

agent-control/src/utils/backoff_gate.rs
```diff
@@ -0,0 +1,307 @@
...
+        }
+    }
+}
+
+/// A thread-safe gate that throttles repeated attempts at a fallible operation
+/// keyed by `K` (e.g. a target version
+pub struct BackoffGate<K, C = SystemClock> {
+    policy: BackoffPolicy,
+    clock: C,
+    state: Mutex<GateState<K>>,
```
DavSanchez: I'm assuming this is to be used for concurrent access right? I still need to check the PR completely, but what's the scenario for this? Isn't each thread/task creating one of those for its own or was the OCI downloader a global structure with `Arc` references from all resources?

agent-control/src/utils/backoff_gate.rs
```diff
@@ -0,0 +1,307 @@
...
+        F: FnOnce() -> Result<T, E>,
+    {
+        match self.check(key) {
+            None => {
+                let result = op();
+                if result.is_ok() {
+                    self.reset();
+                } else {
+                    self.record_failure(key);
+                }
```
DavSanchez, line 126: Both of these operations acquire the `Mutex`-ed resource. Shouldn't we be locking in this function instead so the whole thing is transactional? There's a comment in `record_failure` that is pointing to this if  I understood well. I wonder what other things can happen if there's any race here.

agent-control/src/utils/retry.rs
```diff
@@ -27,9 +27,100 @@ where
...
+/// Configuration for retrying an operation that may fail with transient errors.
+///
+/// When an attempt fails, we wait before retrying with exponential delay —
+/// `base_delay`, then `2× base_delay`, then `4×`, etc. — capped at `max_delay`.
+///
+/// `jitter = true` randomizes each wait somewhere between zero and the computed value.
+#[derive(Debug, Clone, PartialEq)]
+pub struct BackoffPolicy {
+    // `0` is treated as `1`.
+    pub max_attempts: usize,
```
DavSanchez: https://doc.rust-lang.org/stable/std/num/type.NonZeroUsize.html

agent-control/src/oci.rs
```diff
@@ -25,7 +25,15 @@ mod signature_verification;
...
-const DEFAULT_RETRIES: usize = 0;
+/// Default no-retry policy: a single attempt, no backoff.
+fn default_no_retry_policy() -> BackoffPolicy {
+    BackoffPolicy {
+        max_attempts: 1,
+        base_delay: Duration::ZERO,
+        max_delay: Duration::ZERO,
+        jitter: false,
+    }
+}
```
DavSanchez: If I'm not wrong this can be turned into a `const`, even when using [`NonZeroUsize`](https://doc.rust-lang.org/stable/std/num/type.NonZeroUsize.html).

agent-control/src/agent_control/version_updater/on_host.rs
alvarocabanas: For the moment (and probably forever) retry only needs to read the version in:

`self.update(&AgentControlDynamicConfig {
            version: Some(version),
            ..Default::default()
})
`

So there is no problem about using the defaults for the rest of the fields and adding a snapshot is extra code that doesn't give much value, and calling try_upgrade directly bypasses several needed checks.

So I've added a comment documenting the need to revisit this code if there are ever new config fields to be read by the retry path.

agent-control/src/agent_control/version_updater/on_host.rs
alvarocabanas: Done in last commit.

agent-control/src/agent_control/version_updater/updater.rs
alvarocabanas: Yes, it's really handy

agent-control/src/agent_control/version_updater/updater.rs
alvarocabanas: I modified the comment in last commit, this is the Default NoOp Updater implemented by k8s and NoOpUpdater because they don't implement retry. The one is still pending is implemented by the Updater for OnHost.

agent-control/src/utils/backoff_gate.rs
alvarocabanas: Changed to SuppresionReason

agent-control/src/utils/backoff_gate.rs
alvarocabanas: I modified it in last commit with a slight change.

agent-control/src/utils/backoff_gate.rs
alvarocabanas: We didn't decide anything AFAIK

agent-control/src/utils/backoff_gate.rs
alvarocabanas: The Mutex is there because of interior mutability, three of the gate's methods take &self but mutate the state. The alternative would be using RefCell without much benefit.

DavSanchez: So this is used from several places concurrently through an `Arc` or a reference no?

agent-control/src/utils/backoff_gate.rs
alvarocabanas: The lock only guards the tiny state changes, never the slow download in op() , so holding it across I/O would block everything and risk a self-deadlock if the function is called twice. It's safe because the gate only runs from one thread, so nothing actually races.
And as commented before we are using the mutex so the code compiles because of interior mutability and the state lock is to be able to modify the state not locking the operations since we know there are no other callers.

agent-control/src/oci.rs
alvarocabanas: Done in last commit

agent-control/src/utils/retry.rs
alvarocabanas: Done in last commit.

agent-control/src/utils/backoff_gate.rs
DavSanchez: Maybe it's out of scope, but if there are no other callers and no threads perhaps a mutable reference is the way to go as there can only be one at a time 🤔 , we can review it after it's merged.

agent-control/src/utils/backoff_gate.rs
alvarocabanas: It's not but removing the mutext doesn't compile we need interior mutability to keep VersionUpdater::update(&self) and the only option is a Mutex or RefCell I changed it to RefCell so it doesn't cause confusion

agent-control/src/utils/backoff_gate.rs
alvarocabanas: I did change it for RefCell

agent-control/src/utils/backoff_gate.rs
```diff
@@ -0,0 +1,301 @@
...
+    fn default() -> Self {
+        Self {
+            key: None,
+            consecutive_failures: 0,
+            next_attempt_at: None,
+        }
+    }
+}
+
+/// A thread-safe gate that throttles repeated attempts at a fallible operation
```
DavSanchez:
```suggestion
/// A gate that throttles repeated attempts at a fallible operation
```

agent-control/src/oci.rs
DavSanchez: I see still a function (unless GitHub is playing with me which is a possibility xD)

alvarocabanas: Sure, you are right it should be a constant, I thought you only meant the max_attempts field, I did change it into  a constant in last commit.

DavSanchez (approved):

PR #2613, opened 2026-06-17
feat: support Agent type definition unknown fields (NR-575762)

gsanchezgavier:
## Summary

Agent type definitions now tolerate unknown fields instead of failing to parse. This gives forward compatibility: a newer definition (e.g. pushed via the remote registry) that adds fields an older Agent Control doesn't understand will still load, rather than erroring out.

## Important Note
The `protocol_version` is the mechanism that controls features compatibility and this should be leverage only the feature does not fit into a `protocol_version` minor version bump.

## Changes

- Remove `#[serde(deny_unknown_fields)]` from `K8sObjectMeta`, the last struct still rejecting them.
- Add tests asserting that injecting unknown fields yields the same parse result as a clean definition, for both k8s and on-host (Linux/Windows) definitions.

### DIFF
--- a/agent-control/src/agent_type/definition.rs
+++ b/agent-control/src/agent_type/definition.rs
@@ -864,4 +864,139 @@ version: 0.0.1
             ),
         }
     }
+
+    #[test]
+    fn unknown_fields_in_k8s_definition_are_discarded() {
+        let with_unknown_fields =
+            AgentTypeDefinition::from_slice(k8s_definition(true).as_bytes()).unwrap();
+        let without_unknown_fields =
+            AgentTypeDefinition::from_slice(k8s_definition(false).as_bytes()).unwrap();
+
+        assert_eq!(with_unknown_fields, without_unknown_fields);
+    }
+
+    #[rstest]
+    #[case::linux(Environment::Linux)]
+    #[case::windows(Environment::Windows)]
+    fn unknown_fields_in_host_definition_are_discarded(#[case] expected_environment: Environment) {
+        let with_unknown_fields =
+            AgentTypeDefinition::from_slice(host_definition(expected_environment, true).as_bytes())
+                .unwrap();
+        let without_unknown_fields = AgentTypeDefinition::from_slice(
+            host_definition(expected_environment, false).as_bytes(),
+        )
+        .unwrap();
+
+        assert_eq!(with_unknown_fields, without_unknown_fields);
+    }
+
+    /// Builds a Kubernetes agent-type definition. When `with_unknown_fields` is true, an unknown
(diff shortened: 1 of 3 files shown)


agent-control/src/agent_type/definition.rs
```diff
@@ -863,4 +863,140 @@ version: 0.0.1
...
+      {unknown}
+deployment:
+  {unknown}
+  objects:
+    cr1:
+      apiVersion: fake.group/v1beta1
+      kind: FakeKind
+      metadata:
+        name: fake-object
+        namespace: fake-object-namespace
```
danielorihuela, line 926: Shouldn't we have the `{unknown}` here under `metadata` too? Isn't this the data represented by `K8sObjectMeta`?

agent-control/src/agent_type/definition.rs
```diff
@@ -863,4 +863,140 @@ version: 0.0.1
...
+          max_retries: 3
+          last_retry_interval: 30s
+          {unknown}
+  packages:
+    fake_package:
+      {unknown}
+      download:
+        {unknown}
+        oci:
+          registry: fake-registry
```
danielorihuela: `registry` field doesn't exist in `oci`

paologallinaharbur: a true unknown 🦞

paologallinaharbur (dismissed):
I am really happy to see that we were almost there, having this will be a nice tool to add small feature silently

agent-control/src/agent_type/definition.rs
gsanchezgavier: this is part of the k8s API i don't think is related.

agent-control/src/agent_type/definition.rs
gsanchezgavier: removed to avoid confusion thanks1

agent-control/src/agent_type/definition.rs
danielorihuela: We have 

```rust
pub struct K8sObject {
    #[serde(rename = "apiVersion")]
    pub api_version: String,
    pub kind: String,
    pub metadata: K8sObjectMeta,
    #[serde(default, flatten)]
    pub fields: serde_json::Map<String, serde_json::Value>,
}

pub struct K8sObjectMeta {
    #[serde(default)]
    pub labels: BTreeMap<String, String>,
    pub name: String,
    pub namespace: String,
}
```

If we don't add the `{unknown}` here in the test, we won't get an error if we ever put
`#[serde(deny_unknown_fields)]` back or something equivalent.

danielorihuela (approved):

PR #2615, opened 2026-06-17
feat: new explicit filesystem

alvarocabanas:
Replaces the implicit map-of-maps `filesystem:` shape in on-host agent-type definitions with an explicit, recursive, tagged-kind tree: every entry declares `kind: file | dir | dir_content_from_map`, and `dir` entries nest via `entries:`. Migrates all 6 host registry YAMLs and existing test fixtures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Remember that the keywords that you can use are the following:
   crash-looping from saturating CPU.
 - Added support for remote agent type retrieval.
 - add post-download script support for OCI packages.
+- Replace filesystem in on-host agent-type definitions with an explicit, recursive, tagged-kind tree: every entry declares `kind: file | dir | dir_content_from_map`, and `dir` entries nest via `entries:`.
 
 ## v1.17.0 - 2026-06-16
 
--- a/agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure-0.1.0.yaml
+++ b/agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure-0.1.0.yaml
@@ -71,10 +71,18 @@ deployment:
     regex: \d+\.\d+\.\d+
   filesystem:
     config:
-      newrelic-infra.yaml: |
-        ${nr-var:config_agent}
-    integrations.d: ${nr-var:config_integrations}
-    logging.d: ${nr-var:config_logging}
+      kind: dir
+      entries:
+        newrelic-infra.yaml:
+          kind: file
+          text: |
+            ${nr-var:config_agent}
+    integrations.d:
+      kind: dir_content_from_map
+      source: ${nr-var:config_integrations}
+    logging.d:
+      kind: dir_content_from_map
+      source: ${nr-var:config_logging}
   executables:
     - id: newrelic-infra
       path: ${nr-sub:packages.infra-agent.dir}/newrelic-infra
--- a/agent-control/agent-type-registry/newrelic/host-windows-com.newrelic.infrastructure-0.1.0.yaml
+++ b/agent-control/agent-type-registry/newrelic/host-windows-com.newrelic.infrastructure-0.1.0.yaml
@@ -67,12 +67,21 @@ deployment:
     regex: \d+\.\d+\.\d+
   filesystem:
     config:
-      newrelic-infra.yaml: |
-        ${nr-var:config_agent}
-    integrations.d: ${nr-var:config_integrations}
-    logging.d: ${nr-var:config_logging}
+      kind: dir
+      entries:
+        newrelic-infra.yaml:
+          kind: file
+          text: |
+            ${nr-var:config_agent}
+    integrations.d:
+      kind: dir_content_from_map
+      source: ${nr-var:config_integrations}
+    logging.d:
+      kind: dir_content_from_map
+      source: ${nr-var:config_logging}
     # this is needed in order to create the logging integration directory expected by fluentBit
-    newrelic-infra/newrelic-integrations/logging: { }
+    newrelic-infra/newrelic-integrations/logging:
+      kind: dir
   executables:
     - id: newrelic-infra
       path: ${nr-sub:packages.infra-agent.dir}\\newrelic-infra.exe
--- a/agent-control/agent-type-registry/newrelic/host-windows-com.newrelic.infrastructure-0.1.0.yaml
+++ b/agent-control/agent-type-registry/newrelic/host-windows-com.newrelic.infrastructure-0.1.0.yaml
@@ -80,8 +80,14 @@ deployment:
       kind: dir_content_from_map
       source: ${nr-var:config_logging}
     # this is needed in order to create the logging integration directory expected by fluentBit
-    newrelic-infra/newrelic-integrations/logging:
+    newrelic-infra:
       kind: dir
+      entries:
+        newrelic-integrations:
+          kind: dir
+          entries:
+            logging:
+              kind: dir
   executables:
     - id: newrelic-infra
       path: ${nr-sub:packages.infra-agent.dir}\\newrelic-infra.exe
--- a/agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
+++ b/agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
@@ -319,7 +319,6 @@ mod tests {
     // Multi-segment keys are rejected: nested dirs must be declared with `kind: dir` + `entries:`.
     #[case::multi_segment("agent/data", false)]
     #[case::dot_segment("agent/./data", false)]
-    #[case::windows_style_path(r"some\\windows\\style\\path", true)]
     #[case::absolute("/etc", false)]
     #[case::dotdot("agent/../escape", false)]
     fn safe_path_parsing(#[case] path: &str, #[case] should_parse: bool) {
--- a/agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
+++ b/agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
@@ -129,12 +129,12 @@ impl Templateable for FilesystemEntry {
             }
             FilesystemEntry::DirContentFromMap { source } => {
                 let map = source.template_with(variables)?;
-                let children = map
+                let files = map
                     .0
                     .into_iter()
-                    .map(|(k, content)| (PathBuf::from(k), rendered::RenderedEntry::File(content)))
+                    .map(|(k, content)| (PathBuf::from(k), content))
                     .collect();
-                Ok(rendered::RenderedEntry::Dir(children))
+                Ok(rendered::RenderedEntry::DirContentFromMap(files))
             }
         }
     }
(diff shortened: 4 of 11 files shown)


agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure-0.1.0.yaml
```diff
@@ -71,10 +71,18 @@ deployment:
...
-      newrelic-infra.yaml: |
-        ${nr-var:config_agent}
-    integrations.d: ${nr-var:config_integrations}
-    logging.d: ${nr-var:config_logging}
+      kind: dir
+      entries:
+        newrelic-infra.yaml:
+          kind: file
+          text: |
+            ${nr-var:config_agent}
```
danielorihuela, line 79: I wonder if we could get rid of `|`. The `kind` is file and the field is `text`, so we know that we have to use the yaml as string, not as a map.

agent-control/agent-type-registry/newrelic/host-windows-com.newrelic.infrastructure-0.1.0.yaml
```diff
@@ -67,12 +67,21 @@ deployment:
...
+            ${nr-var:config_agent}
+    integrations.d:
+      kind: dir_content_from_map
+      source: ${nr-var:config_integrations}
+    logging.d:
+      kind: dir_content_from_map
+      source: ${nr-var:config_logging}
     # this is needed in order to create the logging integration directory expected by fluentBit
-    newrelic-infra/newrelic-integrations/logging: { }
+    newrelic-infra/newrelic-integrations/logging:
```
paologallinaharbur: I understood we were not allowing this "magic", but we changed so many times the approach that I am no longer sure 😅

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ Remember that the keywords that you can use are the following:
 
 ## Unreleased
 
+### Breaking
```
paologallinaharbur: is it really breaking? since agentTypes are internal I believe that no customer will notice the difference.
Notice that with the release toolkit V2 would be selected automatically

docs/INTEGRATING_AGENTS.md
```diff
@@ -362,9 +362,181 @@ The files can be hardcoded, with the contents possibly containing templates, or
...
+
+| Event              | Ephemeral (`persistent: false`)           | Persistent (`persistent: true`)                        |
+|--------------------|-------------------------------------------|--------------------------------------------------------|
+| Agent start        | Reconcile (manifest diff) + write         | Reconcile (manifest diff) + write                      |
+| Agent stop         | Path deleted                              | Path kept                                              |
+| Agent restart      | Reconcile + write                         | Reconcile + write                                      |
+| Config update      | Reconcile + write                         | Reconcile + write                                      |
+| Removed from fleet | Filesystem dir deleted by ResourceCleaner | Filesystem dir deleted by ResourceCleaner              |
+
+In all the "Reconcile + write" rows, agent-process-created files survive (not in the manifest, not declared, not deleted).
```
paologallinaharbur, line 570: isn't this part of next PR?

agent-control/agent-type-registry/newrelic/host-windows-com.newrelic.infrastructure-0.1.0.yaml
paologallinaharbur: This behaviour is also not mentioned in the docs, 🤔 maybe it was not migrated to the new format by mistake

docs/INTEGRATING_AGENTS.md
```diff
@@ -362,9 +362,181 @@ The files can be hardcoded, with the contents possibly containing templates, or
...
+    entries:
+      data:
+        kind: dir
+        persistent: true
+      integrations.d:
+        kind: dir_content_from_map
+        source: ${nr-var:config_integrations}
+      newrelic-infra.yaml:
+        kind: file
+        text: ${nr-var:config_agent}
```
danielorihuela: Don't we need the `|` here?

```suggestion
        text: |
            ${nr-var:config_agent}
```

docs/INTEGRATING_AGENTS.md
alvarocabanas: Yes it is, but as I commented on the refinement, I added the documentation already in this PR

CHANGELOG.md
alvarocabanas: Well that is a good point, in that case it is not breaking.

docs/INTEGRATING_AGENTS.md
```diff
@@ -362,9 +362,181 @@ The files can be hardcoded, with the contents possibly containing templates, or
...
+    integrations:
+      - name: nri-redis
+        env:
+          HOSTNAME: localhost
+
+config_logging:
+  syslog.yaml: |
+    logs:
+      - name: syslog
+        file: /var/log/syslog
```
danielorihuela, line 480: Should we create a task to make documentation agent type agnostic?
I don't think having examples based on a particular agent type like the infra-agent is good. Every time the infra-agent type changes, we should change the examples, otherwise people reading that can get confused.

Maybe it's not a concern.

agent-control/agent-type-registry/newrelic/host-windows-com.newrelic.infrastructure-0.1.0.yaml
alvarocabanas: Ok you mean the empty dir behavior, isn't it? the thing is kind: dir defaults to empty. 
Do you prefer to force explicitely having an entries key with an empty map?
I can leave it as it is but add it to the docs or opt for that scenario.

danielorihuela: I believe @paologallinaharbur is talking about the path, not the empty directory (I'm not sure).
I remember that we agreed on having the tree even if it was more verbose.

```yaml
a/b.txt:
    kind: file
    ...

# vs

a:
    kind: dir
    entries:
        b:
            kind: file
            ...
```

For me it's okay to have the magic trick. I've seen the implementation and it's free, right? We don't have special code for it? Maybe it's enough with not showing it in the docs.

docs/INTEGRATING_AGENTS.md
danielorihuela: This should be updated too.

agent-control/agent-type-registry/newrelic/host-windows-com.newrelic.infrastructure-0.1.0.yaml
alvarocabanas: Ok I see, it shouldn't be allowed, you are right, let me look

alvarocabanas: My bad, I didn't realize I was passing this.
- I've modified the agent_types to have the new shape.
- I've added a validation to not allow single path.
- I've added an entry to the docs for this behavior

docs/INTEGRATING_AGENTS.md
alvarocabanas: I personally don't mind it, since the infra-agent is our most important agent.

docs/INTEGRATING_AGENTS.md
alvarocabanas: Added in last commit.

docs/INTEGRATING_AGENTS.md
alvarocabanas: Done in last commit

agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure-0.1.0.yaml
alvarocabanas: It's required for the yaml type.

docs/INTEGRATING_AGENTS.md
danielorihuela: Okay, let it be. Can you update the documentation to use the new filesystem? With the `kind` and all that.

agent-control/agent-type-registry/newrelic/host-windows-com.newrelic.infrastructure-0.1.0.yaml
paologallinaharbur: We can always add it back if at anypoint it is needed 😄

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -1,52 +1,66 @@
...
+                FileSystemEntriesError(format!("{} has no parent dir", path.display()))
+            })?;
+            // safe even if the dir already exists.
+            dir_manager.create(parent).map_err(|err| {
+                FileSystemEntriesError(format!("creating directory {parent:?}: {err}"))
+            })?;
+            // Will overwrite the file if it already exists.
+            file_writer
+                .write(path, content.to_owned())
+                .map_err(|err| FileSystemEntriesError(format!("creating file {path:?}: {err}")))
```
paologallinaharbur: nit: what's about having 22 funtions write_file_entry and write_dir_entry?

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -1,52 +1,66 @@
...
+        }
+        RenderedEntry::Dir(children) => {
+            dir_manager.create(path).map_err(|err| {
+                FileSystemEntriesError(format!("creating directory {path:?}: {err}"))
+            })?;
+            for (sub_path, child) in children {
+                write_entry(file_writer, dir_manager, &path.join(sub_path), child)?;
+            }
+            Ok(())
```
paologallinaharbur, line 85: I would add a trace log to show the recursions.
Moreover I would add a log saying that we are creating a directory

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -27,46 +29,44 @@ use serde::de::Error;
...
-/// The files can be hardcoded, with the contents possibly containing templates, or the whole set of
-/// files can be templated so a directory contains an arbitrary number of files (a place to use a
-/// `map[string]yaml` variable type). **The paths cannot be templated.**
-///
-/// See [`AgentDirectoryEntry`] and [`DirEntriesType`] for more details.
+/// Every entry is tagged with a `kind:`. `dir` entries may contain further entries under
+/// `entries:`, recursively.
 #[derive(Debug, Default, Deserialize, Clone, PartialEq)]
-pub struct FileSystem(HashMap<SafePath, DirEntriesType>);
+pub struct FileSystem(HashMap<SafePath, FilesystemEntry>);
```
paologallinaharbur, line 38: dumb question, we do not want SafePath to be a TemplateableValue, right? I guess there is no use case for now

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -362,207 +314,210 @@ mod tests {
...
-    fn file_entry_parsing(
-        #[case] path: &str,
-        #[case] validation: impl Fn(Result<DirEntriesType, serde_saphyr::Error>) -> bool,
-    ) {
-        let yaml = format!("\"{path}\": \"some random content\"");
-        let parsed = serde_saphyr::from_str::<DirEntriesType>(&yaml);
-        let parsed_display = format!("{parsed:?}");
-        assert!(validation(parsed), "input: {yaml}, parsed:{parsed_display}");
+    #[case::single_segment("config", true)]
+    #[case::single_segment_curdir("./config", true)]
```
paologallinaharbur: do you think that it would make sense to deny as well "./config"?

paologallinaharbur (dismissed):

docs/INTEGRATING_AGENTS.md
danielorihuela: @alvarocabanas Isn't this outdated still? I see `filesystem` on 264, but I don't see the new fields like `kind` in there.

danielorihuela: In line 222, it's outdated too.

danielorihuela (dismissed):

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
alvarocabanas: I think for the moment we want to keep filesystem entry names fixed in the agent type and validated up front, if we want something themplateable we can still use dir_content_from_map

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
alvarocabanas: Changed in last commit

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
alvarocabanas: Changed in last commit

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
alvarocabanas: Changed in last commit.

docs/INTEGRATING_AGENTS.md
alvarocabanas: Changed in last commit.

danielorihuela (approved):

paologallinaharbur (approved):

PR #2617, opened 2026-06-18
refactor: config build

danielorihuela:
Simplify agent control config creation in onhost integration tests.

Other details:

* `test_ac_log_to_file_as_root`, `onhost_ac_multiconfig_agents_append_fails` and `self_instrumentation_otel_exports_logs_and_metrics_as_root` don't use it because I judged these are not so common configs, and adding support for them would imply modifications in all other tests. So I preferred to keep a more opinionated struct instead of more general.


gsanchezgavier (approved):
awesome!! . I hope one day we do something similar to the k8s part and we get rid of those folder per test thing :seed

PR #2618, opened 2026-06-19
Feat/filesystem persistence

alvarocabanas:
Implements the `persistent:` filesystem feature, every `file`/`dir`/`dir_content_from_map` entry now accepts `persistent: TemplateableValue<bool>` (default `false`), and the  supervisor runs `delete_ephemeral` on sub-agent at startup (before writing the tree) and stop so ephemeral entries are wiped while persistent ones survive.                                                                                                               
Adds a sidecar manifest at `${base_dir}/.ac-managed-paths.json` that records the paths Agent Control wrote on each successful write; the next write diffs prev vs current declared set and deletes the difference, so removing an entry from the agent type (or a key from a `dir_content_from_map` source) cleans up its on-disk path while agent-process-created files survive because they were never in the manifest. Manifest read failures degrade safely to "delete nothing."                                                                                                                                                                                                               
Wires fleet-removal cleanup through `OnHostCleaner` (now extended with `agent_filesystem_base` + `Arc<DirectoryManager>`) and adds a `purge_stale_agent_filesystems` startup pass for agents removed while AC was stopped.       

## Checklist
 -->

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,6 @@ coverage/
 
 # zig compiler
 .zig-cache/
+
+# Sidecar manifest written by Agent Control's filesystem rendering.
+.ac-managed-paths.json
--- a/.gitignore
+++ b/.gitignore
@@ -60,5 +60,5 @@ coverage/
 # zig compiler
 .zig-cache/
 
-# Sidecar manifest written by Agent Control's filesystem rendering.
+# manifest written by Agent Control's filesystem rendering.
 .ac-managed-paths.json
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Remember that the keywords that you can use are the following:
 - Agent type definitions now tolerate unknown fields for forward compatibility.
 - On-host agents: removed command-based version checking. Agent version is now determined from OCI package metadata, eliminating the need for `deployment.version` configuration in agent type definitions.
 - Replace filesystem in on-host agent-type definitions with an explicit, recursive, tagged-kind tree: every entry declares `kind: file | dir | dir_content_from_map`, and `dir` entries nest via `entries:`.
+- Add new persistence behavior for filesystems in the agent type allowing to set a folder or file as persistent to have different behaviors on stop/start/restart of an agent, it also adds resource cleaning after an agent is removed.
 
 ## v1.17.0 - 2026-06-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ Remember that the keywords that you can use are the following:
 - Agent type definitions now tolerate unknown fields for forward compatibility.
 - On-host agents: removed command-based version checking. Agent version is now determined from OCI package metadata, eliminating the need for `deployment.version` configuration in agent type definitions.
 - Replace filesystem in on-host agent-type definitions with an explicit, recursive, tagged-kind tree: every entry declares `kind: file | dir | dir_content_from_map`, and `dir` entries nest via `entries:`.
-- On-host filesystem entries now accept a `persistent` flag (default `false`): ephemeral entries are deleted on sub-agent stop, persistent entries survive until the agent is removed from the fleet. Reconciliation across writes is driven by a sidecar `.ac-managed-paths.json` manifest (reserved filename — agent types must not declare it) so paths Agent Control no longer owns are deleted while sub-agent-created files are preserved.
+- On-host filesystem entries now accept a `persistent` flag (default `false`): ephemeral entries are deleted on sub-agent stop, persistent entries survive until the agent is removed from the fleet. Reconciliation across writes is driven by a `.ac-managed-paths.json` manifest (reserved filename — agent types must not declare it) so paths Agent Control no longer owns are deleted while sub-agent-created files are preserved.
 
 ## v1.17.0 - 2026-06-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ Remember that the keywords that you can use are the following:
 - Agent type definitions now tolerate unknown fields for forward compatibility.
 - On-host agents: removed command-based version checking. Agent version is now determined from OCI package metadata, eliminating the need for `deployment.version` configuration in agent type definitions.
 - Replace filesystem in on-host agent-type definitions with an explicit, recursive, tagged-kind tree: every entry declares `kind: file | dir | dir_content_from_map`, and `dir` entries nest via `entries:`.
-- Add new persistence behavior for filesystems in the agent type allowing to set a folder or file as persistent to have different behaviors on stop/start/restart of an agent, it also adds resource cleaning after an agent is removed.
+- On-host filesystem entries now accept a `persistent` flag (default `false`): ephemeral entries are deleted on sub-agent stop, persistent entries survive until the agent is removed from the fleet. Reconciliation across writes is driven by a sidecar `.ac-managed-paths.json` manifest (reserved filename — agent types must not declare it) so paths Agent Control no longer owns are deleted while sub-agent-created files are preserved.
 
 ## v1.17.0 - 2026-06-16
 
--- a/agent-control/src/agent_control/resource_cleaner/on_host.rs
+++ b/agent-control/src/agent_control/resource_cleaner/on_host.rs
@@ -2,13 +2,14 @@
 
 use fs::directory_manager::DirectoryManager;
 use std::collections::HashSet;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 use thiserror::Error;
-use tracing::{debug, instrument};
+use tracing::{debug, instrument, warn};
 
 use crate::agent_control::agent_id::AgentID;
+use crate::agent_control::defaults::RESERVED_AGENT_IDS;
 use crate::agent_type::agent_type_id::AgentTypeID;
 use crate::opamp::instance_id::storer::{InstanceIDStorer, StorerError};
 use crate::values::config_repository::{ConfigRepository, ConfigRepositoryError};
@@ -19,6 +20,8 @@ use super::{ResourceCleaner, ResourceCleanerError};
 /// delegating to the same storers that wrote it, also recursively deletes the sub-agent's
 /// dedicated filesystem directory, the `persistent` flag is bypassed here because the agent
 /// has been removed from the fleet.
+/// The same removal logic is reused at startup by [`Self::purge_stale_agents`] to reclaim the
+/// resources of agents removed from the fleet config while Agent Control was stopped.
 pub struct OnHostCleaner<S, C, D>
 where
     S: InstanceIDStorer,
@@ -28,6 +31,7 @@ where
     instance_id_storer: Arc<S>,
     config_repo: Arc<C>,
     agent_filesystem_base: PathBuf,
+    fleet_data_base: PathBuf,
     dir_manager: Arc<D>,
 }
 
@@ -42,13 +46,80 @@ where
         instance_id_storer: Arc<S>,
         config_repo: Arc<C>,
         agent_filesystem_base: PathBuf,
+        fleet_data_base: PathBuf,
         dir_manager: Arc<D>,
     ) -> Self {
         Self {
             instance_id_storer,
             config_repo,
             agent_filesystem_base,
+            fleet_data_base,
             dir_manager,
+        }
+    }
+
+    /// Deletes all on-disk resources Agent Control owns for `agent_id`: its stored remote config,
+    /// its OpAMP instance id, and its dedicated filesystem directory.
+    fn remove_agent_resources(&self, agent_id: &AgentID) -> Result<(), OnHostCleanerError> {
+        debug!(%agent_id, "Cleaning remote config data");
+        self.config_repo
+            .delete_remote(agent_id)
+            .map_err(OnHostCleanerError::RemoteConfig)?;
+
+        debug!(%agent_id, "Cleaning opamp identifier data");
+        self.instance_id_storer
+            .delete(agent_id)
+            .map_err(OnHostCleanerError::InstanceId)?;
+
+        let fs_dir = self.agent_filesystem_base.join(agent_id);
+        debug!(%agent_id, path = ?fs_dir, "Cleaning agent filesystem directory");
+        self.dir_manager
+            .delete(&fs_dir)
+            .map_err(|err| OnHostCleanerError::Filesystem {
+                path: fs_dir,
+                source: err,
+            })
+    }
+
+    /// At startup, reclaims the resources of any agent that is no longer in the agents config.
+    pub fn purge_stale_agents<'a>(&self, configured_agent_ids: impl IntoIterator<Item = &'a str>) {
+        let mut skip: HashSet<String> = configured_agent_ids.into_iter().map(String::from).collect();
+        skip.extend(RESERVED_AGENT_IDS.iter().map(|id| id.to_string()));
+
+        let mut names: HashSet<String> = HashSet::new();
+        names.extend(self.agent_dir_names(&self.agent_filesystem_base));
+        names.extend(self.agent_dir_names(&self.fleet_data_base));
+
+        for name in names {
+            if skip.contains(&name) {
+                continue;
+            }
+            let agent_id = match AgentID::try_from(name.as_str()) {
+                Ok(id) => id,
+                Err(err) => {
+                    warn!(?err, name, "skipping stale directory with invalid agent id");
+                    continue;
+                }
+            };
+            tracing::info!(%agent_id, "reclaiming resources of agent no longer in fleet config");
+            if let Err(err) = self.remove_agent_resources(&agent_id) {
+                warn!(?err, %agent_id, "failed to reclaim stale agent resources");
+            }
+        }
+    }
+
+    /// Lists the immediate child directory names under `base` (the per-agent subdirectories).
+    /// A missing `base` yields no names; a listing error is logged and treated as empty.
+    fn agent_dir_names(&self, base: &Path) -> Vec<String> {
+        match self.dir_manager.list(base) {
+            Ok(entries) => entries
+                .iter()
+                .filter_map(|p| p.file_name().and_then(|n| n.to_str()).map(String::from))
+                .collect(),
+            Err(err) => {
+                warn!(?err, ?base, "cannot list agent directory for stale cleanup");
+                Vec::new()
+            }
         }
     }
 }
@@ -68,26 +139,7 @@ where
         if agent_id == &AgentID::AgentControl {
             return Err(OnHostCleanerError::AgentControlId.into());
         }
-
-        debug!(%agent_id, "Cleaning remote config data");
-        self.config_repo
-            .delete_remote(agent_id)
-            .map_err(OnHostCleanerError::RemoteConfig)?;
-
-        debug!(%agent_id, "Cleaning opamp identifier data");
-        self.instance_id_storer
-            .delete(agent_id)
-            .map_err(OnHostCleanerError::InstanceId)?;
-
-        let fs_dir = self.agent_filesystem_base.join(agent_id);
-        debug!(%agent_id, path = ?fs_dir, "Cleaning agent filesystem directory");
-        self.dir_manager
-            .delete(&fs_dir)
-            .map_err(|err| OnHostCleanerError::Filesystem {
-                path: fs_dir,
-                source: err,
-            })?;
-
+        self.remove_agent_resources(agent_id)?;
         Ok(())
     }
 }
@@ -118,48 +170,10 @@ impl From<OnHostCleanerError> for ResourceCleanerError {
     }
 }
 
-/// Removes filesystem directories under `agent_filesystem_base` whose name does not match a
-/// currently-configured agent ID. This handles "agent removed from fleet config while Agent
-/// Control was stopped": on next start, the orphaned directory is reclaimed.
-pub fn purge_stale_agent_filesystems<'a>(
-    dir_manager: &impl DirectoryManager,
-    agent_filesystem_base: &std::path::Path,
-    configured_agent_ids: impl IntoIterator<Item = &'a str>,
-) {
-    let configured: HashSet<&str> = configured_agent_ids.into_iter().collect();
-
-    let entries = match dir_manager.list(agent_filesystem_base) {
-        Ok(entries) => entries,
-        Err(err) => {
-            tracing::warn!(
-                ?err,
-                path = ?agent_filesystem_base,
-                "skipping stale-agent filesystem cleanup: cannot read base dir"
-            );
-            return;
-        }
-    };
-
-    for path in entries {
-        let Some(name) = path.file_name().and_then(|n| n.to_str()) else {
-            continue;
-        };
-        if configured.contains(name) {
-            continue;
-        }
-        tracing::info!(
-            path = ?path,
-            "purging filesystem directory of agent no longer in fleet config"
-        );
-        if let Err(err) = dir_manager.delete(&path) {
-            tracing::warn!(?err, ?path, "failed to purge stale agent filesystem dir");
-        }
-    }
-}
-
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::agent_control::defaults::AGENT_CONTROL_ID;
     use crate::opamp::instance_id::storer::tests::MockInstanceIDStorer;
     use crate::values::config_repository::tests::MockConfigRepository;
     use ::fs::directory_manager::mock::MockDirectoryManager;
@@ -176,6 +190,10 @@ mod tests {
 
     fn fs_base() -> PathBuf {
         PathBuf::from("/var/lib/newrelic-agent-control/filesystem")
+    }
+
+    fn fleet_base() -> PathBuf {
+        PathBuf::from("/var/lib/newrelic-agent-control/fleet-data")
     }
 
     #[test]
@@ -204,6 +222,7 @@ mod tests {
             Arc::new(instance_id_storer),
             Arc::new(config_repo),
             fs_base(),
+            fleet_base(),
             Arc::new(dir_manager),
         );
 
@@ -231,6 +250,7 @@ mod tests {
             Arc::new(instance_id_storer),
             Arc::new(config_repo),
             fs_base(),
+            fleet_base(),
             Arc::new(dir_manager),
         );
 
@@ -251,6 +271,7 @@ mod tests {
             Arc::new(instance_id_storer),
             Arc::new(config_repo),
             PathBuf::new(),
+            PathBuf::new(),
             Arc::new(dir_manager),
         );
 
@@ -259,41 +280,136 @@ mod tests {
         assert!(result.is_err());
     }
 
-    /// Subdirs under the filesystem base whose names are not in the configured-agents set are
-    /// removed; configured ones survive.
-    #[test]
-    fn purge_stale_agent_filesystems_removes_only_orphans() {
-        let kept_path = fs_base().join("kept");
-        let orphan_path = fs_base().join("orphan");
-
-        let mut dir_manager = MockDirectoryManager::new();
-        dir_manager.should_list(&fs_base(), vec![kept_path.clone(), orphan_path.clone()]);
-        dir_manager.should_delete(&orphan_path);
-        // `kept` is in the configured set → no `delete` call expected for it.
-
-        purge_stale_agent_filesystems(&dir_manager, &fs_base(), ["kept"]);
+    fn cleaner(
+        instance_id_storer: MockInstanceIDStorer,
+        config_repo: MockConfigRepository,
+        dir_manager: MockDirectoryManager,
+    ) -> OnHostCleaner<MockInstanceIDStorer, MockConfigRepository, MockDirectoryManager> {
+        OnHostCleaner::new(
+            Arc::new(instance_id_storer),
+            Arc::new(config_repo),
+            fs_base(),
+            fleet_base(),
+            Arc::new(dir_manager),
+        )
     }
 
+    /// Orphans (agents no longer configured) are fully reclaimed — remote config, instance id and
+    /// filesystem dir — and discovered from BOTH the filesystem and fleet-data bases. Configured
+    /// agents survive.
     #[test]
-    fn purge_stale_agent_filesystems_is_noop_on_empty_base() {
+    fn purge_reclaims_orphans_from_filesystem_and_fleet_data() {
+        let orphan_fs = agent_id("orphan-fs");
+        let orphan_fleet = agent_id("orphan-fleet");
+
+        let mut dir_manager = MockDirectoryManager::new();
+        dir_manager.should_list(
+            &fs_base(),
+            vec![fs_base().join("kept"), fs_base().join("orphan-fs")],
+        );
+        dir_manager.should_list(&fleet_base(), vec![fleet_base().join("orphan-fleet")]);
+        // `remove_agent_resources` always deletes the agent's filesystem dir (idempotent).
+        dir_manager.should_delete(&fs_base().join("orphan-fs"));
+        dir_manager.should_delete(&fs_base().join("orphan-fleet"));
+
+        let mut config_repo = MockConfigRepository::new();
+        config_repo
+            .expect_delete_remote()
+            .with(predicate::eq(orphan_fs.clone()))
+            .once()
+            .returning(|_| Ok(()));
+        config_repo
+            .expect_delete_remote()
+            .with(predicate::eq(orphan_fleet.clone()))
+            .once()
+            .returning(|_| Ok(()));
+
+        let mut instance_id_storer = MockInstanceIDStorer::new();
+        instance_id_storer
+            .expect_delete()
+            .with(predicate::eq(orphan_fs))
+            .once()
+            .returning(|_| Ok(()));
+        instance_id_storer
+            .expect_delete()
+            .with(predicate::eq(orphan_fleet))
+            .once()
+            .returning(|_| Ok(()));
+
+        cleaner(instance_id_storer, config_repo, dir_manager).purge_stale_agents(["kept"]);
+    }
+
+    /// Agent Control's own directory (a reserved ID) is never reclaimed.
+    #[test]
+    fn purge_skips_agent_control_dir() {
+        let orphan = agent_id("orphan");
+
         let mut dir_manager = MockDirectoryManager::new();
         dir_manager.should_list(&fs_base(), vec![]);
-        dir_manager.expect_delete().never();
+        dir_manager.should_list(
+            &fleet_base(),
+            vec![fleet_base().join(AGENT_CONTROL_ID), fleet_base().join("orphan")],
+        );
+        dir_manager.should_delete(&fs_base().join("orphan"));
 
-        purge_stale_agent_filesystems(&dir_manager, &fs_base(), ["any"]);
+        let mut config_repo = MockConfigRepository::new();
+        config_repo
+            .expect_delete_remote()
+            .with(predicate::eq(orphan.clone()))
+            .once()
+            .returning(|_| Ok(()));
+
+        let mut instance_id_storer = MockInstanceIDStorer::new();
+        instance_id_storer
+            .expect_delete()
+            .with(predicate::eq(orphan))
+            .once()
+            .returning(|_| Ok(()));
+
+        cleaner(instance_id_storer, config_repo, dir_manager).purge_stale_agents([]);
     }
 
-    /// If listing the base dir fails, the helper logs and returns without attempting any
-    /// deletes. Stale dirs survive until the next AC start.
     #[test]
-    fn purge_stale_agent_filesystems_skips_when_list_fails() {
+    fn purge_is_noop_when_there_are_no_orphans() {
+        let mut dir_manager = MockDirectoryManager::new();
+        dir_manager.should_list(&fs_base(), vec![]);
+        dir_manager.should_list(&fleet_base(), vec![]);
+        dir_manager.expect_delete().never();
+
+        let mut config_repo = MockConfigRepository::new();
+        config_repo.expect_delete_remote().never();
+        let mut instance_id_storer = MockInstanceIDStorer::new();
+        instance_id_storer.expect_delete().never();
+
+        cleaner(instance_id_storer, config_repo, dir_manager).purge_stale_agents(["any"]);
+    }
+
+    /// If listing one base fails, the helper logs and still reclaims orphans found in the other.
+    #[test]
+    fn purge_continues_when_one_base_listing_fails() {
+        let orphan = agent_id("orphan");
+
         let mut dir_manager = MockDirectoryManager::new();
         dir_manager
             .expect_list()
             .with(predicate::eq(fs_base()))
             .return_once(|_: &Path| Err(std::io::Error::other("boom")));
-        dir_manager.expect_delete().never();
+        dir_manager.should_list(&fleet_base(), vec![fleet_base().join("orphan")]);
+        dir_manager.should_delete(&fs_base().join("orphan"));
 
-        purge_stale_agent_filesystems(&dir_manager, &fs_base(), ["any"]);
+        let mut config_repo = MockConfigRepository::new();
+        config_repo
+            .expect_delete_remote()
+            .with(predicate::eq(orphan.clone()))
+            .once()
+            .returning(|_| Ok(()));
+        let mut instance_id_storer = MockInstanceIDStorer::new();
+        instance_id_storer
+            .expect_delete()
+            .with(predicate::eq(orphan))
+            .once()
+            .returning(|_| Ok(()));
+
+        cleaner(instance_id_storer, config_repo, dir_manager).purge_stale_agents([]);
     }
 }
--- a/agent-control/src/agent_control/resource_cleaner/on_host.rs
+++ b/agent-control/src/agent_control/resource_cleaner/on_host.rs
@@ -155,9 +155,12 @@ pub enum OnHostCleanerError {
     /// Failed to delete the stored remote configuration.
     #[error("failed to delete stored remote config: {0}")]
     RemoteConfig(#[source] ConfigRepositoryError),
+    /// Failed to delete agent filesystem directory.
     #[error("failed to delete agent filesystem directory {path:?}: {source}")]
     Filesystem {
+        /// The path in the filesystem that couldn't be deleted.
         path: PathBuf,
+        /// The io error.
         #[source]
         source: std::io::Error,
     },
--- a/agent-control/src/agent_control/resource_cleaner/on_host.rs
+++ b/agent-control/src/agent_control/resource_cleaner/on_host.rs
@@ -210,8 +210,8 @@ mod tests {
         assert!(cleaner.clean(&id, &any_type_id()).is_ok());
     }
 
-    /// `delete` on a missing dir is a no-op per the trait contract — `OnHostCleaner` doesn't
-    /// pre-check existence; it just delegates.
+    /// When the directory manager's `delete` fails, `clean` propagates the error rather than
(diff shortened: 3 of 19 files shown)


agent-control/src/agent_control/resource_cleaner/on_host.rs
```diff
@@ -111,23 +190,103 @@ mod tests {
...
+            Arc::new(dir_manager),
+        );
 
         assert!(cleaner.clean(&id, &any_type_id()).is_ok());
     }
 
+    /// `delete` on a missing dir is a no-op per the trait contract — `OnHostCleaner` doesn't
+    /// pre-check existence; it just delegates.
+    #[test]
+    fn clean_propagates_directory_manager_delete_error() {
```
DavSanchez: Comment does not fully match test no?

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -1,33 +1,180 @@
...
+
+    /// Reconciles the on-disk state under `base_dir` against the current declared tree, then
+    /// writes the declared tree, then updates the sidecar manifest.
     pub fn write(
         &self,
         file_writer: &impl FileWriter,
         dir_manager: &impl DirectoryManager,
     ) -> Result<(), FileSystemEntriesError> {
-        for (path, entry) in &self.0 {
+        // `FileSystem::default()` produces an empty `base_dir` so we don't want to write a manifest
```
DavSanchez:
```rust
        // `FileSystem::default()` produces an empty `base_dir`
```

Should we try to guarantee this does not happen? Not sure if it's possible since we have the debug-builds capable of modifying where the base dirs are, but might be something to consider.

agent-control/src/sub_agent/on_host/supervisor.rs
DavSanchez, line 167: This performs the cleanup even if the `filesystem` value is dropped above no?

DavSanchez (commented):
Looking great so far! I have some questions

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -1,33 +1,180 @@
...
-use tracing::trace;
+use tracing::{trace, warn};
+
+/// Filename of the sidecar manifest Agent Control writes inside each sub-agent's filesystem dir.
+/// The manifest records the absolute paths AC wrote on the previous successful write event so
+/// the next write can compute "previously managed but no longer declared → delete".
+///
+/// Reserved name: agent-type definitions cannot declare a path with this exact filename at any
+/// level. The filesystem parser validates this at parse time.
+pub const MANAGED_PATHS_MANIFEST_FILENAME: &str = ".ac-managed-paths.json";
```
danielorihuela, line 15: > The filesystem parser validates this at parse time.

Is this true? I don't see where it's this validated. Anyways, even if we validated it, this file could be modified by an agent type post download script.

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -1,33 +1,180 @@
...
+        let manifest_path = self.manifest_path();
+        let prev_declared = read_manifest(&manifest_path);
+        let curr_declared = self.collect_declared_paths();
+
+        // Reconcile: delete paths AC owned previously but no longer declares.
+        let mut stale: Vec<&PathBuf> = prev_declared
+            .iter()
+            .filter(|p| !curr_declared.contains(*p))
+            .collect();
+        // Sort by depth descending so we delete leaves first; saves directory-not-empty churn.
```
danielorihuela: I don't understand what's "directory-not-empty" churn.

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -520,4 +554,367 @@ foo:
...
+            ),
+        ]);
+
+        serde_saphyr::from_str::<FileSystem>(first_yaml)
+            .unwrap()
+            .template_with(&variables_first)
+            .unwrap()
+            .write(&LocalFile, &DirectoryManagerFs)
+            .unwrap();
+
```
danielorihuela, line 700: I think at this point we should assert that files from the first manifest exist in the file system.

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -520,4 +554,367 @@ foo:
...
+            "agent-created file inside dir_content_from_map should survive"
+        );
+        assert_eq!(
+            std::fs::read_to_string(&runtime_in_projected).unwrap(),
+            "state"
+        );
+    }
+
+    /// `delete_ephemeral` removes ephemeral entries' on-disk paths but leaves persistent ones
+    /// alone. A persistent child of an ephemeral parent is wiped along with the parent.
```
danielorihuela: This test doesn't have an ephemeral folder with a persistent child. The comment doesn't match the test.

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -520,4 +554,367 @@ foo:
...
+        templated.delete_ephemeral().unwrap();
+
+        assert!(!tmp_dir.path().join("ephemeral.txt").exists());
+        assert!(!tmp_dir.path().join("ephemeral-dir").exists());
+        assert!(tmp_dir.path().join("persistent.txt").exists());
+        assert!(tmp_dir.path().join("persistent-dir").is_dir());
+    }
+
+    #[test]
+    fn removed_parent_dir_takes_agent_created_descendants_with_it() {
```
danielorihuela, line 837: Isn't this tested in `reconciles_against_current_declared_set`?

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -123,13 +124,15 @@ where
...
-        // Reuse supervisor inner fields
         let Self {
             agent_identity,
             package_manager,
             internal_publisher,
             thread_contexts,
             logging_path,
+            // Dropped on purpose; listed by name (not via `..`) so any future field
+            // added to the struct fails to compile here until it's explicitly threaded through.
+            filesystem: _,
```
danielorihuela: Nit: Maybe we can just write `..`

This is not really dropping the value, is it? We can access it through `self.filesystem`?

agent-control/tests/on_host/scenarios/stale_agent_filesystem_cleanup.rs
```diff
@@ -0,0 +1,45 @@
...
+    AgentControlConfigBuilder::basic(opamp_server.endpoint(), opamp_server.jwks_endpoint())
+        .write(dirs.local_dir());
+
+    let _agent_control =
+        start_agent_control_with_custom_config(dirs.base_paths(), AGENT_CONTROL_MODE_ON_HOST);
+
+    // The orphan dir must be gone after AC startup. Use a small retry to allow the start-up
+    // tasks to land before asserting.
+    crate::common::retry::retry(30, Duration::from_secs(1), || {
+        if orphan_dir.exists() {
```
danielorihuela, line 45: Isn't the `kept_dir` removed too? The manifest is not created, is it?

agent-control/src/agent_control/resource_cleaner/on_host.rs
alvarocabanas: Fixed in last commit.

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
alvarocabanas: I've removed default that was only used on tests and created a specific function for tests.

agent-control/src/sub_agent/on_host/supervisor.rs
alvarocabanas: No, there is no filesystem cleanup. It will be the next write that will reconcile the on-disk state.

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
alvarocabanas: Yes, now it's implemented, and also there is a new check to ensure the files in the manifest are in the agent's base dir or they are ignroed.

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
alvarocabanas: I changed the comment for a more meaninfull one

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
alvarocabanas: Done in last commit.

agent-control/src/sub_agent/on_host/supervisor.rs
alvarocabanas: I put .. as you suggested. I also removed the missleading comment.

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
alvarocabanas: True, I changed the comment.

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
alvarocabanas: No, Both simulate the agent wrote a file AC didn't declare, but one is under a dir that survives vs. under a dir that's about to disappear.
So the assertions are the opposite.

agent-control/tests/on_host/scenarios/stale_agent_filesystem_cleanup.rs
alvarocabanas: Done in last commit

agent-control/src/agent_control/resource_cleaner/on_host.rs
```diff
@@ -111,23 +190,103 @@ mod tests {
...
         let mut config_repo = MockConfigRepository::new();
         config_repo.expect_delete_remote().never();
+        let mut dir_manager = MockDirectoryManager::new();
+        dir_manager.expect_delete().never();
 
-        let cleaner = OnHostCleaner::new(Arc::new(instance_id_storer), Arc::new(config_repo));
+        let cleaner = OnHostCleaner::new(
+            Arc::new(instance_id_storer),
+            Arc::new(config_repo),
+            fs_base(),
```
danielorihuela: Nit: Use `PathBuf::new` to make it super clear that the path is not relevant when the agent id is refused.

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -1,34 +1,205 @@
...
+        Self { base_dir, entries }
+    }
+
+    #[cfg(test)]
+    pub(crate) fn test_empty() -> Self {
+        let base_dir = tempfile::tempdir()
+            .expect("create temp dir for test FileSystem")
+            .keep();
+        Self::new(base_dir, HashMap::new())
+    }
```
danielorihuela: Nit: Move this into the test module.

We can have an `impl Filesystem` at the bottom, with the tests, and avoid cluttering the production code.

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -1,34 +1,205 @@
...
+        let manifest_path = self.manifest_path();
+        let prev_declared = read_manifest(&manifest_path, &self.base_dir);
+        let curr_declared = self.collect_declared_paths();
+
+        // Reconcile: delete paths AC owned previously but no longer declares.
+        let mut stale: Vec<&PathBuf> = prev_declared
+            .iter()
+            .filter(|p| !curr_declared.contains(*p))
+            .collect();
+        // Deepest first, so a dir's stale children are gone before we remove the dir.
```
danielorihuela, line 187: Curious: What's the reason for that? It's just to avoid errors when deleting a file if it was already deleted as part of a dir, or is there something else?

alvarocabanas: Since there is persistence granularity, we need to go from leaf to branch, maybe a file is not persistent but the dir is.

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -1,34 +1,205 @@
...
+    managed_paths
+        .into_iter()
+        .filter(|p| {
+            // Never let the manifest mark itself for deletion.
+            if p == path {
+                return false;
+            }
+            let within = is_within_base(p, base_dir);
+            if !within {
+                warn!(?p, ?base_dir, "ignoring out-of-base managed-paths entry");
```
danielorihuela: Does it make sense to change the message so that's clear that someone or something tried to do something nasty, and that this is not an Agent Control issue per se?

"Manifest contains out-of-base entry. An agent is not allowed to modify files outside it's isolated filesystem. Ignoring path."

It's a bit of a lie, because a post-install script can modify it. However, I think this makes it clear that something went wrong. With the original message, it seems like agent control did something wrong.

agent-control/src/agent_control/resource_cleaner/on_host.rs
paologallinaharbur, line 1: In the PR description you say:

> and the supervisor runs delete_ephemeral on sub-agent stop so ephemeral entries are wiped while persistent ones survive.

Shouldn't we execute it also at startup time?

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ Remember that the keywords that you can use are the following:
 - Added support for remote agent type retrieval.
 - add post-download script support for OCI packages.
 - Replace filesystem in on-host agent-type definitions with an explicit, recursive, tagged-kind tree: every entry declares `kind: file | dir | dir_content_from_map`, and `dir` entries nest via `entries:`.
+- On-host filesystem entries now accept a `persistent` flag (default `false`): ephemeral entries are deleted on sub-agent stop, persistent entries survive until the agent is removed from the fleet. Reconciliation across writes is driven by a sidecar `.ac-managed-paths.json` manifest (reserved filename — agent types must not declare it) so paths Agent Control no longer owns are deleted while sub-agent-created files are preserved.
```
paologallinaharbur: sidecar?

paologallinaharbur: why are not we deleting at sub-agent start as well? if AC is interrupted and than restarted we might have the files still there.

(now that I think about it it could be a edge case)

.gitignore
```diff
@@ -59,3 +59,6 @@ coverage/
 
 # zig compiler
 .zig-cache/
+
+# Sidecar manifest written by Agent Control's filesystem rendering.
+.ac-managed-paths.json
```
paologallinaharbur, line 64: This is the linux convention to create hidden files, are we creating "real" hidden files in windows as well?

docs/INTEGRATING_AGENTS.md
paologallinaharbur, line 1: There is an other file of the docs to be updated, it is `AC_repositories_and_files.md` to specify the location of the new file

fs/src/directory_manager.rs
```diff
@@ -53,6 +57,18 @@ impl DirectoryManager for DirectoryManagerFs {
...
         remove_dir_all(path)
     }
+
+    #[instrument(skip_all, fields(path = %path.display()))]
+    fn list(&self, path: &Path) -> io::Result<Vec<PathBuf>> {
+        validate_path(path)?;
+
+        let read = match std::fs::read_dir(path) {
+            Ok(r) => r,
+            Err(err) if err.kind() == io::ErrorKind::NotFound => return Ok(Vec::new()),
```
paologallinaharbur, line 70: should we add a debug log? is it something that will happen often?

paologallinaharbur (dismissed):
If we merge this aren't we deleting the data from the
```
NRIA_AGENT_DIR: "${nr-sub:filesystem_agent_dir}/newrelic-infra"
```

Why can't I see such folder from the list in the fylesystem?

Should we create a hierarchy now that we can with newrelic-infra and inside integrations.d?


Also `NRIA_SAFE_BIN_DIR: "${nr-sub:packages.infra-agent.dir}/integrations"` could be creating issues

fs/src/win_permissions.rs
```diff
@@ -40,7 +41,8 @@ fn get_administrator_sid() -> Result<Vec<u8>, PermissionError> {
 }
 
 /// set_file_permissions_for_administrator removes any other ACL from a file only granting
-/// read and write to Administrators.
+/// read, write, and delete to Administrators. DELETE is needed so the same Administrator that
```
paologallinaharbur, line 45: without the delete we cannot delete the file later on? 🤔

agent-control/src/agent_type/runtime_config/on_host/rendered.rs
```diff
@@ -7,7 +7,7 @@ use crate::agent_type::runtime_config::{
 };
 use std::collections::HashMap;
 
-#[derive(Debug, Default, Clone, PartialEq)]
```
paologallinaharbur, line 11: isn't this a big change? or it is not since now agentType are "splitted"?

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -176,7 +177,11 @@ where
     }
 
     fn stop(self) -> Result<(), ThreadContextStopperError> {
-        stop_supervisor_threads(self.thread_contexts)
+        let stop_result = stop_supervisor_threads(self.thread_contexts);
```
paologallinaharbur, line 202: Shoun't we delete files after the apply as well? Also in that case we are calling "stop_supervisor_threads"

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -768,7 +846,7 @@ pub mod tests {
             MockPackageManager::new_arc(),
             false,
             PathBuf::default(),
-            FileSystem::default(),
+            FileSystem::test_empty(),
```
paologallinaharbur, line 950: why are we creating a folder in all tests instead of leaving             FileSystem::default(),?

agent-control/src/agent_control/resource_cleaner/on_host.rs
```diff
@@ -79,12 +111,53 @@ impl From<OnHostCleanerError> for ResourceCleanerError {
     }
 }
 
+/// Removes filesystem directories under `agent_filesystem_base` whose name does not match a
+/// currently-configured agent ID. This handles "agent removed from fleet config while Agent
+/// Control was stopped": on next start, the orphaned directory is reclaimed.
+pub fn purge_stale_agent_filesystems<'a>(
```
paologallinaharbur: Are we not cleaning as well downloaded packages?

paologallinaharbur: could we call it as well after applying a new list of agents or the code is shared with k8s?

agent-control/src/agent_control/resource_cleaner/on_host.rs
```diff
@@ -11,33 +14,47 @@ use crate::values::config_repository::{ConfigRepository, ConfigRepositoryError};
 use super::{ResourceCleaner, ResourceCleanerError};
 
 /// On-host implementation of [`ResourceCleaner`] that wipes a sub-agent's fleet data by
-/// delegating to the same storers that wrote it.
-pub struct OnHostCleaner<S, C>
+/// delegating to the same storers that wrote it, also recursively deletes the sub-agent's
+/// dedicated filesystem directory, the `persistent` flag is bypassed here because fleet removal
```
paologallinaharbur: "fleet removal is final"? What does it mean?

agent-control/src/agent_control/resource_cleaner/on_host.rs
paologallinaharbur: There is the resource cleaner for it

paologallinaharbur: why are we not re-using the resource cleaner as well?
if feels like we are duplicating the code, no?
Moreover, we are deleting less resources

agent-control/src/agent_control/resource_cleaner/on_host.rs
```diff
@@ -59,6 +76,15 @@ where
             .delete(agent_id)
             .map_err(OnHostCleanerError::InstanceId)?;
 
+        let fs_dir = self.agent_filesystem_base.join(agent_id);
```
paologallinaharbur, line 74: should we remove oci packages here?

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -1,34 +1,205 @@
...
+    }
+    let mut managed_paths: Vec<PathBuf> = declared.iter().cloned().collect();
+    managed_paths.sort();
+    let body = serde_json::to_vec(&ManagedPathsManifest { managed_paths })
+        .map_err(|err| std::io::Error::new(std::io::ErrorKind::InvalidData, err))?;
+    std::fs::write(path, body)
+}
+
+fn delete_path(path: &Path) -> Result<(), FileSystemEntriesError> {
+    trace!("Deleting stale managed path {}", path.display());
```
danielorihuela: This function is not only called to delete stale managed paths.

agent-control/src/agent_control/resource_cleaner/on_host.rs
alvarocabanas: Declared ephemeral files are rewritten on startup with freshly-rendered content, so they're effectively reset.
Regarding directories content for an ephemeral dir, if there are agent created files inside it not declared in the agent type they will be carried over, but I thought it was not an issue, but it's true is a weird behavior so I'll modify it.

agent-control/src/agent_control/resource_cleaner/on_host.rs
alvarocabanas: Done in last commit

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
alvarocabanas: Done in last commit

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
alvarocabanas: Done in last commit

agent-control/src/agent_control/resource_cleaner/on_host.rs
alvarocabanas: Done in last commit

CHANGELOG.md
alvarocabanas: I've removed the term sidecar in the last commit since it's confusing.

alvarocabanas: I have added it in last commit, it's indeed an edge case but it's cheap and it's more consistent to cover it.

.gitignore
alvarocabanas: No it's not a hidden file in Windows.

fs/src/directory_manager.rs
alvarocabanas: I think it would be noise, we are here just checking if there is anything to purge.

.gitignore
paologallinaharbur: Should it be? Or should be not hidden in linux?

docs/INTEGRATING_AGENTS.md
alvarocabanas: Done in last commit

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -1,34 +1,197 @@
...
+            let within = is_within_base(p, base_dir);
+            if !within {
+                warn!(?p, ?base_dir, "An agent is not allowed to modify files outside it's isolated filesystem. Ignoring path.");
+            }
+            within
+        })
+        .collect()
+}
+
+fn is_within_base(path: &Path, base_dir: &Path) -> bool {
```
paologallinaharbur, line 261: can you add a description of this method?

.gitignore
alvarocabanas: Is it a must?

agent-control/src/agent_control/resource_cleaner/on_host.rs
alvarocabanas: It means that persistence is not taking in consideration and everything is deleted, I will remove that if it's confusing

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -1,34 +1,197 @@
...
+            for sub in files.keys() {
+                declared.insert(path.join(sub));
+            }
+        }
+    }
+}
+
+/// Reads the managed-paths manifest at `path`, returning only entries that are safe to act on:
+/// paths genuinely contained in `base_dir`.
+fn read_manifest(path: &Path, base_dir: &Path) -> HashSet<PathBuf> {
```
paologallinaharbur: if we have a ManagedPathsManifest why we return HashSet<PathBuf> instead?

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -1,34 +1,197 @@
...
+    fn collect_declared_paths(&self) -> HashSet<PathBuf> {
+        let mut declared = HashSet::new();
+        for (path, entry) in &self.entries {
+            collect_recursive(path, entry, &mut declared);
+        }
+        declared
+    }
+}
+
+fn collect_recursive(path: &Path, entry: &RenderedEntry, declared: &mut HashSet<PathBuf>) {
```
paologallinaharbur: same here with HashSet<PathBuf>

agent-control/src/agent_control/resource_cleaner/on_host.rs
alvarocabanas: Removed in last commit

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -1,34 +1,197 @@
...
+    !has_escape && path.is_absolute() && path.starts_with(base_dir)
+}
+
+fn write_manifest(path: &Path, declared: &HashSet<PathBuf>) -> Result<(), std::io::Error> {
+    if let Some(parent) = path.parent() {
+        std::fs::create_dir_all(parent)?;
+    }
+    let mut managed_paths: Vec<PathBuf> = declared.iter().cloned().collect();
+    managed_paths.sort();
+    let body = serde_json::to_vec(&ManagedPathsManifest { managed_paths })
```
paologallinaharbur: It seems that we use the real type jyst at the end

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -1,34 +1,197 @@
...
+        .map_err(|err| std::io::Error::new(std::io::ErrorKind::InvalidData, err))?;
+    std::fs::write(path, body)
+}
```
paologallinaharbur: should they be under impl ManagedPathsManifest{}?

agent-control/src/sub_agent/on_host/supervisor.rs
alvarocabanas: Because I removed Default on purpose, Default was only called from the tests and was setting a filesystem in the root and creating the manifest. So I opted to remove Default and have this helper function instead.

danielorihuela:
From the spike we had

Event | Ephemeral (persistent: false) | Persistent (persistent: true)
-- | -- | --
Agent start | Wipe + recreate + write managed files | mkdir -p if absent; write managed files
Agent stop | Directory deleted | Directory kept
Agent restart | Wipe + recreate + write managed files | Reconcile managed files; agent-created files preserved
Config update | Wipe + recreate + write managed files | Overwrite managed files; agent-created files preserved
Removed from fleet | Directory deleted (already gone) | Directory deleted by ResourceCleaner

And from the code we have


Event | Function calls
-- | --
Agent start | `delete_ephemeral`, `write`
Agent apply | `delete_ephemeral`, `write`
Agent stop | `delete_ephemeral`
Removed from fleet (happy path) | `self.dir_manager.delete(&fs_dir)`
Removed from fleet (crash path) | `purge_stale_agent_filesystems`

We always delete what is declared as ephemeral in the current agent type.
When removed from fleet, we remove the whole dir.
On a crash, we purge folders whose name don't appear in the list of declared agent ids.
On start and apply, we also delete whatever got removed from the agent type.

I think it makes sense.

agent-control/src/agent_control/resource_cleaner/on_host.rs
alvarocabanas: I have changed the code in last commit to reuse the cleaner, plus I also added now the deletion of the rest of the resources the cleaner deletes.

agent-control/src/agent_type/runtime_config/on_host/rendered.rs
alvarocabanas: No, since now Filesystem doesn't have default it was also removed from here, it's not used anywhere so there is no issue.

agent-control/src/agent_control/resource_cleaner/on_host.rs
alvarocabanas: We are not, we were also not doing it with the Resource Cleaner, I agree we need to do it, but I would leave it for a second PR to not make this one even bigger.

agent-control/src/agent_control/resource_cleaner/on_host.rs
alvarocabanas: Same as before, I would leave it for another PR.

alvarocabanas:
> If we merge this aren't we deleting the data from the
> 
> ```
> NRIA_AGENT_DIR: "${nr-sub:filesystem_agent_dir}/newrelic-infra"
> ```
> 
> Why can't I see such folder from the list in the fylesystem?
> 
> Should we create a hierarchy now that we can with newrelic-infra and inside integrations.d?
> 
> Also `NRIA_SAFE_BIN_DIR: "${nr-sub:packages.infra-agent.dir}/integrations"` could be creating issues

Regarding ${nr-sub:packages.infra-agent.dir}/integrations" could be creating issues, the infra-agent won't use the binary unless the configuration is there, so in this case, I think is enough with having the binary cleaning in the ResourceCleaner as i said I was going to do in another pR.

You are right about the newrelic-infra folder, it was actually a bug and I have changed it in last commit.

.gitignore
paologallinaharbur: I just find a bit weird that we use a linux notation for a windows file.
But it is just a nit, just ignore this

agent-control/src/agent_control/resource_cleaner/on_host.rs
paologallinaharbur: we can also have a task for it 😄

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
alvarocabanas: Added in last commit

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
alvarocabanas: ManagedPathsManifest is the serializer only used to load or save the entries but then we iterate the hashSet

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
alvarocabanas: same

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
alvarocabanas: same

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
alvarocabanas: I refactorized it in last commit.

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
alvarocabanas: Change in last commit

agent-control/src/agent_control/resource_cleaner/on_host.rs
alvarocabanas: The runtime apply path already does the equivalent per-agent

alvarocabanas: I'll add the task

alvarocabanas: Done

fs/src/win_permissions.rs
alvarocabanas: Yes, because of Windows ACL, without the DELETE, a later remove_file would fail with ERROR_ACCESS_DENIED, even it's the same Administrator that wrote it

agent-control/src/sub_agent/on_host/supervisor.rs
alvarocabanas: A file eliminated between the old and new config is removed by write's manifest reconciliation, and ephemeral state is freshened by delete_ephemeral, so there is not need to do what you suggest.

agent-control/src/agent_control/resource_cleaner/on_host.rs
```diff
@@ -1,43 +1,132 @@
...
+                path: fs_dir,
+                source: err,
+            })
+    }
+
+    /// At startup, reclaims the resources of any agent that is no longer in the agents config.
+    pub fn purge_stale_agents<'a>(&self, configured_agent_ids: impl IntoIterator<Item = &'a str>) {
+        let mut skip: HashSet<String> =
+            configured_agent_ids.into_iter().map(String::from).collect();
+        skip.extend(RESERVED_AGENT_IDS.iter().map(|id| id.to_string()));
```
sigilioso: Nit: we can avoid or skip the `mut`:

We can chain the iterators to avoid it:

```rust
        let skip: HashSet<String> = configured_agent_ids
            .into_iter()
            .map(String::from)
            .chain(RESERVED_AGENT_IDS.iter().map(|id| id.to_string()))
            .collect();

```

We can explicitly say when the variable is mutable:

```rust
        let skip = {
            let mut skip: HashSet<String> =
                configured_agent_ids.into_iter().map(String::from).collect();
            skip.extend(RESERVED_AGENT_IDS.iter().map(|id| id.to_string()));
            skip
        };
```

agent-control/src/agent_control/resource_cleaner/on_host.rs
```diff
@@ -1,43 +1,132 @@
...
+    }
+
+    /// Lists the immediate child directory names under `base` (the per-agent subdirectories).
+    /// A missing `base` yields no names; a listing error is logged and treated as empty.
+    fn agent_dir_names(&self, base: &Path) -> Vec<String> {
+        match self.dir_manager.list(base) {
+            Ok(entries) => entries
+                .iter()
+                .filter_map(|p| p.file_name().and_then(|n| n.to_str()).map(String::from))
+                .collect(),
```
sigilioso: Nit: do we need to `.collect()` here? Considering that we use the result to extend a `HashSet` maybe we could return the iterator directly.

agent-control/src/agent_control/run/on_host.rs
```diff
@@ -107,8 +108,23 @@ impl AgentControlRunner {
...
+            fleet_data_base,
+            dir_manager,
+        );
+        resource_cleaner.purge_stale_agents(
+            agent_control_config
+                .dynamic
+                .agents
+                .keys()
+                .map(|id| id.as_str()),
+        );
```
sigilioso, line 137: Nit: should we add a debug log line before executing `purge_stale_agents`? If stale agents are found the first log line will be:

```rust
        debug!(%agent_id, "Cleaning remote config data");
```

from the cleaner implementation. Connecting it to the initial cleanup could help understanding the logs.

agent-control/src/sub_agent/on_host/supervisor.rs
sigilioso: Would it make sense to say this explicitly in a comment? I actually had the same doubt when going through the code.

docs/INTEGRATING_AGENTS.md
```diff
@@ -482,7 +482,7 @@ config_logging:
 
 The runtime produces the following on disk under `${nr-sub:filesystem_agent_dir}`. Each kind is shown in isolation.
 
-**`kind: file`**: single file rendered from the templated `text:` field. `persistent: true` means it survives `ResourceCleaner` on teardown.
+**`kind: file`**: single file rendered from the templated `text:` field. `persistent: true` keeps it across sub-agent stop and restarts.
```
sigilioso: We mean sub-agent stop and Agent Control restarts, right? I think _restarts_ can be a bit ambiguous here.

fs/src/directory_manager.rs
```diff
@@ -53,6 +57,18 @@ impl DirectoryManager for DirectoryManagerFs {
         }
         remove_dir_all(path)
     }
+
+    #[instrument(skip_all, fields(path = %path.display()))]
+    fn list(&self, path: &Path) -> io::Result<Vec<PathBuf>> {
```
sigilioso, line 65: Shouldn't this `list` new function have its own unit tests?

alvarocabanas: Done in last commit

agent-control/src/agent_control/run/on_host.rs
alvarocabanas: Done in last commit

agent-control/src/agent_control/resource_cleaner/on_host.rs
alvarocabanas: Done in last commit

agent-control/src/agent_control/resource_cleaner/on_host.rs
alvarocabanas: Done in last commit

docs/INTEGRATING_AGENTS.md
alvarocabanas: I changed it for agent control, it's true that it was confusing the user.

agent-control/src/sub_agent/on_host/supervisor.rs
alvarocabanas: Done in last commit

docs/INTEGRATING_AGENTS.md
```diff
@@ -520,54 +520,96 @@ agent/integrations.d/
...
 
 The flag does **not** shield the entry from intentional removal: if you delete an entry from the agent type (or remove a key from a `dir_content_from_map` source map), the manifest diff catches it and the on-disk path is deleted on the next write event.
 
-###### Sidecar manifest
+**`persistent` applies per entry and does not cascade to children.** When cleaning (on stop, and before each (re)write), cleanup walks the declared tree: a persistent entry is kept and the walk descends into its children, while an ephemeral entry is deleted together with its entire on-disk subtree (a recursive `remove_dir_all`, which stops the walk there). So a nested path survives cleanup only if **every declared node on the path is `persistent: true`**.
+
+**`dir_content_from_map` has no `persistent` flag.** Agent Control owns and re-renders the projected files on every write, so it is always ephemeral. A `persistent:` key left in the YAML is silently ignored, so older configs still parse.
```
sigilioso, line 551: Nice explanation! 💯

agent-control/src/agent_control/resource_cleaner/on_host.rs
```diff
@@ -1,46 +1,133 @@
...
+            if skip.contains(&name) {
+                continue;
+            }
+            let agent_id = match AgentID::try_from(name.as_str()) {
+                Ok(id) => id,
+                Err(err) => {
+                    warn!(?err, name, "skipping stale directory with invalid agent id");
+                    continue;
+                }
+            };
```
gsanchezgavier, line 106: nit: i thing this is already been done in the k8s GC and probably when processing a AC remote config, maybe we could have a fn for it

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -1,40 +1,276 @@
...
+
+#[derive(Debug, Default, Serialize, Deserialize)]
+struct ManagedPathsManifest {
+    managed_paths: Vec<PathBuf>,
+}
+
+impl ManagedPathsManifest {
+    /// Loads the manifest at `path`. A missing or malformed file creates an empty manifest (logged)
+    fn load(path: &Path) -> Self {
+        let raw = match std::fs::read(path) {
```
sigilioso, line 137: I've just noticed that some operations use the `fs` (internal crate) abstractions and some others use `std::fs` directly. Is there a specific reason why? I'd try to be consistent and use the internal abstraction when possible (even if the unit-tests use the real filesystem instead of mocks).

This could be addressed in a follow-up PR with no issue.

gsanchezgavier, line 137: we should use fs reader from our crate right?

sigilioso (approved):
I've left a comment regarding the usage of `fs` vs. `std::fs`, but it can be addressed in a follow-up PR after merging this one.

Everything else looks good to me.

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -1,40 +1,276 @@
...
+        }
+
+        for (path, entry) in &self.entries {
+            entry.write(path, file_writer, dir_manager)?;
+        }
+
+        let mut managed_paths: Vec<PathBuf> = curr_declared.iter().cloned().collect();
+        managed_paths.sort();
+        let manifest = ManagedPathsManifest { managed_paths };
+        if let Err(err) = manifest.save(&manifest_path) {
```
gsanchezgavier, line 204: are we calling this from multiple threads?

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -231,6 +259,24 @@ fn check_single_segment(path: &Path) -> Result<(), String> {
     ))
 }
 
+/// Rejects the reserved manifest filename at any level. Agent Control writes its
+/// managed-paths manifest at `<base_dir>/.ac-managed-paths.json`; an entry declaring that name
```
gsanchezgavier, line 263: is this needed? isn't this manifest file outside of the fs tree?

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
alvarocabanas: Yes you are right, and I can tackle it in a new PR

agent-control/src/agent_control/resource_cleaner/on_host.rs
alvarocabanas: Same as the FS, I can tackle it in a newer PR

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
alvarocabanas: See the comment above

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
alvarocabanas: No, there is one manifest per agent thread

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
alvarocabanas: No, it leaves on each filesystem for each agent

PR #2626, opened 2026-06-22
fix(self-update): treat empty version string as absent (no-op)

danielorihuela:
Pushing `version: ""` from Fleet Control caused Agent Control to silently pull `docker.io/.../manifests/latest`, risking an unversioned upgrade across the entire fleet.

## Context

Fleet Control can push a remote config with a `version` field to trigger an on-host self-update. The bug was that an empty string `""` and an absent field had different behaviors:

- Absent `version` → no update (correct)
- `version: ""` → pull `:latest` tag (wrong)

An absent field correctly short-circuits in `on_host.rs::update()` via `let Some(new_version) = &config.version`. But an empty string deserialized to `Some(Version(""))`, passed the guard, and `Oci::to_reference` fell into a `(None, None)` arm that silently defaulted to `:latest`.

## Technical details

**Fix is applied at two layers:**

**Layer 1 — `Version::from_str` (`rendered.rs`):** Empty strings are now rejected at the type level (`Err(InvalidVersion("version cannot be empty"))`). This enforces the invariant that a `Version` value always contains a non-empty string.

**Layer 2 — `deserialize_version` custom deserializer (`config.rs`):** A `deserialize_with` function on `AgentControlDynamicConfig::version` maps `""` → `None` _before_ calling `Version::from_str`. This means Fleet Control pushing `version: ""` is silently treated as absent — a no-op with no error logged and no config rejection.

The two layers are intentionally separate: `Version` enforces a type invariant (non-empty), while the config deserializer enforces a domain semantic (empty from Fleet Control = no intent to update).

**`#[serde(default)]` is required alongside `deserialize_with`:** Normally serde implicitly handles absent `Option<T>` fields as `None`. Adding `deserialize_with` opts out of that implicit behavior, so `default` must be declared explicitly to restore it for the absent-key case.

**`Oci::to_reference` now returns `Result<Reference, InvalidVersion>`:** The `(None, None)` arm — reachable only if a `Version("")` is constructed directly in code, bypassing `from_str` — now returns an explicit error instead of panicking. `tag_and_digest` routes an empty inner string to `(None, None)` to make this path explicit and testable.

**Documentation:** Updated `CONFIG.md`, `docs/ac-remote-update/how-it-works.md` (OnHost now ✅, added on-host section), and `docs/opamp-communication-flows/ac-remote-update.md` (added on-host `version` field examples).

### DIFF
--- a/agent-control/src/agent_control/config.rs
+++ b/agent-control/src/agent_control/config.rs
@@ -277,10 +277,18 @@ pub struct AgentControlDynamicConfig {
     )]
     pub version: Option<Version>,
     /// chart_version represent the AC chart version that needs to be executed.
-    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        default,
+        deserialize_with = "deserialize_chart_version"
+    )]
     pub chart_version: Option<String>,
     /// cd_chart_version represent the agent control cd chart version that needs to be executed.
-    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        default,
+        deserialize_with = "deserialize_chart_version"
+    )]
     pub cd_chart_version: Option<String>,
 }
 
@@ -394,6 +402,17 @@ where
         Some(v) => Version::from_str(v)
             .map(Some)
             .map_err(|e| de::Error::custom(e.to_string())),
+    }
+}
+
+fn deserialize_chart_version<'de, D>(deserializer: D) -> Result<Option<String>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let s: Option<String> = Option::deserialize(deserializer)?;
+    match s.as_deref() {
+        None | Some("") => Ok(None),
+        Some(v) => Ok(Some(v.to_string())),
     }
 }
 
--- a/agent-control/src/agent_control/config.rs
+++ b/agent-control/src/agent_control/config.rs
@@ -22,7 +22,7 @@ use crate::{
 use http::HeaderMap;
 use kube::api::TypeMeta;
 use oci_client::secrets::RegistryAuth;
-use serde::{Deserialize, Deserializer, Serialize};
+use serde::{Deserialize, Deserializer, Serialize, de};
 use std::collections::HashMap;
 use std::fmt::Display;
 use std::str::FromStr;
@@ -270,7 +270,11 @@ const AGENTS_KEY: &str = "agents";
 pub struct AgentControlDynamicConfig {
     pub agents: SubAgentsMap,
     /// version represent the AC version that needs to be executed.
-    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        default,
+        deserialize_with = "deserialize_version"
+    )]
     pub version: Option<Version>,
     /// chart_version represent the AC chart version that needs to be executed.
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -377,6 +381,19 @@ impl TryFrom<YAMLConfig> for AgentControlDynamicConfig {
             .map_err(|e| AgentControlConfigError(format!("deserializing: {e}")))?;
         serde_saphyr::from_str(&value_string)
             .map_err(|e| AgentControlConfigError(format!("serializing: {e}")))
+    }
+}
+
+fn deserialize_version<'de, D>(deserializer: D) -> Result<Option<Version>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let s: Option<String> = Option::deserialize(deserializer)?;
+    match s.as_deref() {
+        None | Some("") => Ok(None),
+        Some(v) => Version::from_str(v)
+            .map(Some)
+            .map_err(|e| de::Error::custom(e.to_string())),
     }
 }
 
@@ -878,8 +895,16 @@ agents: {}
     }
 
     #[test]
+    fn dynamic_config_empty_version_deserializes_as_none() {
+        let config =
+            serde_saphyr::from_str::<AgentControlDynamicConfig>("agents: {}\nversion: \"\"\n")
+                .expect("empty version should deserialize successfully");
+        assert_eq!(config.version, None);
+    }
+
+    #[test]
     fn dynamic_config_invalid_version_fails_to_deserialize() {
-        let yaml = "agents: {}\nversion: invalid-version; rm -rf /\n";
+        let yaml = "agents: {}\nversion: \"invalid-version; rm -rf /\"\n";
         assert!(serde_saphyr::from_str::<AgentControlDynamicConfig>(yaml).is_err());
     }
 
--- a/agent-control/src/agent_type/runtime_config/on_host/package/rendered.rs
+++ b/agent-control/src/agent_type/runtime_config/on_host/package/rendered.rs
@@ -56,6 +56,12 @@ impl Oci {
             (None, Some(digest)) => {
                 Ok(Reference::with_digest(registry_str, repository_str, digest))
             }
+
+            // We should never reach this case.
+            // For sub-agents, the version is always required. If the version is empty, it should be caught during validation.
+            // For Agent Control, the version is optional. If the version isn't configured, or it's an empty string,
+            // we don't update and this function is never called.
+            // This error is here as a defensive programming measure to catch any unexpected cases where the version is empty.
             (None, None) => Err(InvalidVersion("version cannot be empty".to_string())),
         }
     }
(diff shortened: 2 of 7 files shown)


docs/ac-remote-update/how-it-works.md
```diff
@@ -29,6 +29,33 @@ Flux is wrapped in the `agent-control-cd` chart, which is used internally. The c
...
 
+### On-host
+
+Fleet Control pushes a remote config containing a `version` field. Agent Control downloads the specified OCI artifact, verifies its signature, self-replaces its binary, and restarts.
+
+#### `version` field (remote config)
+
+The `version` field controls which version Agent Control should upgrade to.
+
+An absent `version` and an empty `version: ""` are both treated as no-ops — neither will trigger an update. This prevents accidental fleet-wide upgrades caused by pushing an unset version field.
```
paologallinaharbur, line 40: are we logging a warning if this happens or what is happening?

I am seeing that we are launching the error ".map_err(|e| OciClientError::FetchArtifact(format!("building OCI reference: {e}")))?"

agent-control/src/agent_type/runtime_config/on_host/package/rendered.rs
```diff
@@ -40,22 +40,23 @@ pub struct PostDownloadHook {
...
         let registry_str = registry.to_string();
         let repository_str = self.repository.to_string();
 
         match self.version.tag_and_digest() {
-            (Some(tag), Some(digest)) => {
-                Reference::with_tag_and_digest(registry_str, repository_str, tag, digest)
-            }
-            (Some(tag), None) => Reference::with_tag(registry_str, repository_str, tag),
-            (None, Some(digest)) => Reference::with_digest(registry_str, repository_str, digest),
-            (None, None) => {
```
paologallinaharbur, line 56: the previous seemed to point to latest also for missing version not only for empty, no?

agent-control/src/agent_control/config.rs
```diff
@@ -270,7 +270,11 @@ const AGENTS_KEY: &str = "agents";
 pub struct AgentControlDynamicConfig {
     pub agents: SubAgentsMap,
     /// version represent the AC version that needs to be executed.
-    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        default,
```
paologallinaharbur, line 275: why default is needed in this case?

agent-control/src/agent_control/config.rs
```diff
@@ -270,7 +270,11 @@ const AGENTS_KEY: &str = "agents";
...
+        deserialize_with = "deserialize_version"
```
paologallinaharbur, line 276: why` chart_version` and `cd_chart_version` do not need this particular serialiser? Could they be affected by the same bug?

MunhozThiago (commented):
### What's Good

- **Design clarity.** The two-layer separation (type invariant in `from_str` + domain semantic in `deserialize_version`) is well-reasoned, and the PR description explains the `why` exceptionally well.
- **`Oci::to_reference` now returns `Result`.** The `(None, None)` silent `:latest` default was the vulnerability; it's now an explicit `Err`.
- **Tests cover the critical paths.** `dynamic_config_empty_version_deserializes_as_none`, `test_to_reference_empty_version_returns_error`, and the updated invalid-version test all cover the new behavior.
- **Docs are high quality.** `how-it-works.md` clearly documents version field semantics, valid formats, and the happy path.

---

### Issues (not yet raised)

**1. Dead constant `DEFAULT_TAG` — likely compiler warning**

`DEFAULT_TAG: &str = "latest"` in `rendered.rs` is no longer referenced after the `(None, None)` arm was changed to return `Err`. Rust will emit a `dead_code` warning (or error if `deny(warnings)` is set). Should be removed.

---

**2. Missing test for absent `version` key with new deserializer**

The `deserialize_with` + `default` combination is the trickiest part of this change. Worth adding an explicit test to document the absent-key behavior:

```rust
#[test]
fn dynamic_config_absent_version_deserializes_as_none() {
    let config = serde_saphyr::from_str::<AgentControlDynamicConfig>("agents: {}\n")
        .expect("absent version should deserialize successfully");
    assert_eq!(config.version, None);
}
```

---

**3. Nit: duplicate error message across two contexts**

`"version cannot be empty"` appears in both `Version::from_str` and `Oci::to_reference`. The `to_reference` path is reachable only if `Version(String::new())` is constructed directly (bypassing `from_str`), so a distinct message like `"internal error: Version contains empty string"` would better signal it's a programming error vs. input validation.

---

### Verdict

**Approve with minor nits.** The core fix is correct and well-tested. `DEFAULT_TAG` dead constant should be cleaned up before merging.

agent-control/src/agent_type/runtime_config/on_host/package/rendered.rs
```diff
@@ -40,22 +40,23 @@ pub struct PostDownloadHook {
     pub env: Env,
 }
 
-const DEFAULT_TAG: &str = "latest";
```
paologallinaharbur, line 43: this change is affecting also sub-agents, no?

agent-control/src/agent_control/config.rs
DavSanchez: The `deserialize_version` might only be called if there's any data for this field (which would make handling the `None` case somewhat redundant, but Rust's system requires it), the `default` then just makes this value be `None` in case there's no data whatsoever.

docs/ac-remote-update/how-it-works.md
```diff
@@ -6,7 +6,7 @@ Remote update is a capability of Agent Control that enable updating its version
 
 | System     | Supported |
 |------------|-----------|
-| OnHost     | ❌        |
+| OnHost     | ✅        |
```
DavSanchez, line 9: ❤️ 🚀

MunhozThiago (dismissed):

agent-control/src/agent_control/config.rs
paologallinaharbur: Thanks!

docs/ac-remote-update/how-it-works.md
danielorihuela: TL;DR No, we don't log a warning.

The error you mention should never happen. I added it just in case (defensive programming).

Why should never happen? Check the following flow.

1. We get the new `AgentControlDynamicConfig` from remote.
2. `deserialize_version` is called to deserialize the version
3. If the value is empty, we treat it as `Ok(None)`
4. we jump over `update`

```rust
fn update(&self, config: &AgentControlDynamicConfig) -> Result<(), UpdaterError> {
    ...

    let Some(new_version) = &config.version else {
        debug!("Version is not specified in the dynamic config");
        return Ok(());
    };

    ...
}
```

The error is in the `download` function, which is called by `install_archive`, called by `install`, called by `update`.

`download` is also used downloading subagents. There the version is mandatory.

paologallinaharbur: This is what I was thinking, thanks for clarifying!

agent-control/src/agent_type/runtime_config/on_host/package/rendered.rs
paologallinaharbur: I understand that for the same reason there is no error, before we were not reaching this point. I suggest to add a comment saying that we are not expected to reach this point in case of AC.  I would also clarify what happens for subAgents

danielorihuela: No. `to_reference` is implemented over `Oci`, which is

```rust
pub struct Oci {
    pub repository: Repository,
    pub version: Version,
    pub public_key_url: Option<Url>,
}
```

The `Version` has to be there.

For subagents, it's mandatory.
For agent control we had

```rust
#[serde(skip_serializing_if = "Option::is_none")]
pub version: Option<Version>,
```

If it's not configured, then we got a `None` and we jumped over the `update` function.

agent-control/src/agent_type/runtime_config/on_host/package/rendered.rs
paologallinaharbur: I mean.. it was almost impossible to reach such point since version was required, but theoretically if a version variable was not for any reason, we could have reached such point

agent-control/src/agent_control/config.rs
danielorihuela: At first, I thought it wasn't needed and out of scope.
I double checked again because of your comment. I think the current behaviour in k8s is to update the `HelmRelease` with an empty version, which will make Flux fail on the reconcile.

I added the same behaviour in the last commit, LMKWYT https://github.com/newrelic/newrelic-agent-control/pull/2626/changes/e5a505419f487fc850fa2995ec82ddae6476fcfc

agent-control/src/agent_type/runtime_config/on_host/package/rendered.rs
danielorihuela: In theory not, but as you said, yes. If we reach that point, now we will get an error, which might be even better.

agent-control/src/agent_control/config.rs
paologallinaharbur: I believe that is better being explicit anyway rather than rely on flux internal implementation!

paologallinaharbur (dismissed):
Thanks!

paologallinaharbur (approved):

PR #2632, opened 2026-06-25
feat(cli): add uninstall and update subcommands to on-host CLI

MunhozThiago:
## Summary

Adds two new subcommands to `newrelic-agent-control-cli` for lifecycle management of on-host Agent Control installations.

### Why package manager commands are wrong

Agent Control installs and updates managed agents (infra-agent, NRDOT, OTel collectors) via **OCI binary replacement**, completely bypassing `apt`/`yum`/`zypper`. This means:
- `apt remove newrelic-agent-control` only removes the AC package entry — managed agents keep running
- `apt upgrade newrelic-agent-control` risks downgrading AC if Fleet Control already self-updated via OCI

These new CLI commands operate on the actual installation state.

---

### `newrelic-agent-control-cli uninstall [--keep-config] [--dry-run]`

1. Stops the AC daemon (`systemctl stop`)
2. Disables and removes the systemd unit
3. Removes `/var/lib/newrelic-agent-control/` — all OCI-installed managed agent packages live here, invisible to the system package manager
4. Removes `/etc/newrelic-agent-control/` (skip with `--keep-config`)
5. Removes both binaries (`/usr/bin/newrelic-agent-control` and `newrelic-agent-control-cli`)

### `newrelic-agent-control-cli update --version <semver> [--dry-run]`

Downloads the specified AC version from the public OCI registry (`docker.io/newrelic/agent-control-artifacts`) and atomically self-replaces the running binary via `BinarySelfReplacer`. The process exits so `systemd Restart=always` starts the new version.

**Intended for break-glass use only** — Fleet Control release channels are the correct upgrade path for managed installations.

## Test plan

- [ ] `cargo test -p newrelic_agent_control --test integration_tests uninstall` — 3 dry-run tests pass
- [ ] `cargo test -p newrelic_agent_control --test integration_tests update_` — 4 tests pass
- [ ] `newrelic-agent-control-cli uninstall --dry-run` lists all paths that would be removed
- [ ] `newrelic-agent-control-cli update --version 1.17.0 --dry-run` prints version and OCI repo

## Related

Companion to guided install PR: https://source.datanerd.us/nr1-dev-experience/shared-component-framework-configs/pull/1254

🤖 Generated with [Claude Code](https://claude.com/claude-code)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ Remember that the keywords that you can use are the following:
 - On-host self-update: an empty `version: ""` pushed from Fleet Control now behaves the same as an absent `version` field (no-op, no update attempted). Previously it silently triggered a pull of the `:latest` OCI tag.
 
 ### enhancement
+- Added `uninstall` and `update --version <semver>` subcommands to `newrelic-agent-control-cli` for on-host installations not managed by Fleet Control. `uninstall` stops the service, removes all OCI-managed agent packages, and removes binaries. `update` performs an OCI-based self-replace with signature verification enabled by default.
```
paologallinaharbur, line 21: I understood that we should assume that all update should come via FC and that we should not allow to ping a specific version, isn't this the case anymore?

scripts/test-cli-lifecycle.sh
paologallinaharbur, line 1: I would not add a new way to test the cli, e2e and unit test should be enough!

agent-control/src/bin/main_agent_control_onhost_cli.rs
```diff
@@ -24,6 +26,16 @@ enum Commands {
...
+    /// Uninstall Agent Control and all managed agents from this host.
+    ///
+    /// Stops the service, removes all OCI-installed managed agent packages, removes
+    /// binaries, and (unless --keep-config) removes configuration and state directories.
+    Uninstall(UninstallArgs),
+    /// Update Agent Control to the specified version using the OCI registry.
+    ///
+    /// This bypasses Fleet Control release channels and is intended only as a break-glass
+    /// operation for installations not managed by Fleet Control.
+    Update(UpdateArgs),
```
paologallinaharbur, line 38: We currently do not have an "install" command, and I believe that we never interact with systemctl or the package manage directly

paologallinaharbur: is this the right place to have this?

agent-control/src/cli/on_host/uninstall.rs
```diff
@@ -0,0 +1,284 @@
...
+    #[arg(long)]
+    pub keep_config: bool,
+
+    /// Show what would be removed without making any changes.
+    #[arg(long)]
+    pub dry_run: bool,
+
+    /// Skip the confirmation prompt (for automation / CI).
+    #[arg(long)]
+    pub yes: bool,
```
paologallinaharbur: the name of the arg is just "yes"?

agent-control/src/cli/on_host/uninstall.rs
```diff
@@ -0,0 +1,284 @@
...
+        confirm_uninstall(args.keep_config)?;
+    }
+
+    stop_service_and_wait(args.dry_run);
+    disable_service(args.dry_run);
+
+    // Remove state dir — contains all OCI-installed managed agent packages.
+    // These agents are NOT tracked by the system package manager and must be
+    // removed explicitly here.
+    remove_path(
```
paologallinaharbur, line 87: This seems a hacky way to delete everything, why the package manager is not used?

agent-control/src/cli/on_host/update.rs
```diff
@@ -0,0 +1,151 @@
...
+            .map_err(|e| CliError::Command(format!("failed to create async runtime: {e}")))?,
+    );
+
+    let oci_client = Client::try_new(ClientConfig::default(), ProxyConfig::default(), runtime)
+        .map_err(|e| CliError::Command(format!("failed to create OCI client: {e}")))?;
+
+    let downloader = OCIPackageArtifactDownloader::new(
+        oci_client,
+        Registry::default(),
+        None, // no auth — public registry
```
paologallinaharbur, line 119: We should consider as well private registries

agent-control/src/cli/on_host/uninstall.rs
```diff
@@ -0,0 +1,284 @@
...
+use tracing::{info, warn};
+
+#[cfg(target_family = "unix")]
+const AC_BINARY: &str = "/usr/bin/newrelic-agent-control";
+#[cfg(target_family = "unix")]
+const AC_CLI_BINARY: &str = "/usr/bin/newrelic-agent-control-cli";
+#[cfg(target_family = "windows")]
+const AC_BINARY: &str =
+    r"C:\Program Files\New Relic\newrelic-agent-control\newrelic-agent-control.exe";
+#[cfg(target_family = "windows")]
```
paologallinaharbur, line 18: we have already a uninstall.ps1

agent-control/src/cli/on_host/update.rs
```diff
@@ -0,0 +1,151 @@
...
+    info!(
+        "Binary downloaded to {}. Performing self-replace.",
+        new_binary.display()
+    );
+
+    BinarySelfReplacer::self_replace(&new_binary)
+        .map_err(|e| CliError::Command(format!("self-replace failed: {e}")))?;
+
+    // The binary has been replaced. Exit so that systemd (Restart=always) starts
+    // the new version.
```
paologallinaharbur: I think that the cli is hallucinating here! if we exit the CLI it does not mean that the AC will restart

paologallinaharbur:
> Why package manager commands are wrong
Agent Control installs and updates managed agents (infra-agent, NRDOT, OTel collectors) via OCI binary replacement, completely bypassing apt/yum/zypper. This means:
> apt remove newrelic-agent-control only removes the AC package entry — managed agents keep running
apt upgrade newrelic-agent-control risks downgrading AC if Fleet Control already self-updated via OCI
These new CLI commands operate on the actual installation state.

I do not think this is correct!

Moreover:
> apt upgrade newrelic-agent-control risks downgrading AC if Fleet Control already self-updated via OCI
Is still an issue

paologallinaharbur (changes requested):
I do not think that this is the right place to implement the update/uninstallation.

I think that it should be implemented next to the code that is installing it!
Apart from this, the implementation needs to be cleaned 😄

CHANGELOG.md
MunhozThiago: I am working on this to provide a way for customers to get out from stucked situations...

agent-control/src/cli/on_host/uninstall.rs
MunhozThiago: Because if it has never been installed via package manager, I thought we should not rely on it to uninstall as well... maybe we can find a better way to remove it though.

MunhozThiago (commented):
Agree the UX should live next to install, but I would rather prefer to have a thin layer on the nr-cli rather than reimplementing everything there. In my mind, the problem will always be the nr-cli is not aware of the OCI installations. So, do you think something like this could work?

(1) this PR provides the binary implementation that handles OCI-installed managed agents correctly (package manager can't do this);
(2) follow-up PR adds an `uninstall:` section to the open-install-library recipe that calls this binary. The same pattern as `generate-config` today: recipe bootstraps, binary executes.

Or do you think it the only way would be to add `agent-control uninstall` wrapper to nr-cli too?

paologallinaharbur:
We can talk about it async, what I do not agree with is this "The same pattern as generate-config today: recipe bootstraps, binary executes." The cli is not installing, just generating a config. There is a wrapper on top of it doing the installation.

The same wrapper should be the one in charge of doing the uninstallation otherwise there is the risk of leaving behind data (for example the package manager entries or the service in systemctl)

> Or do you think it the only way would be to add agent-control uninstall wrapper to nr-cli too?

I think there are different solutions. We could either leverage the post uninstall script of the package manager, we could rely on the cli as you did, but before implementing we should avaluate first pros-cons. Moreover, I would split the uninstall and the upgrade discussion since they are two different problems!

paologallinaharbur:
Closing this for now, we are going to do a step-back and to consider the two features (uninstall/upgrade) separately.  @danielorihuela will start checking the uninstall use case to understand where we should put the code

PR #2636, opened 2026-06-26
feat: add e2e for post_download_hook

vjripoll:
Adds end-to-end tests for the on-host post_download_hook feature in the e2e-runner, covering the success and failure paths against a real Agent Control installation.

Scenarios:

- post-download-hook-success — package ships a hook that succeeds; verifies (via a filesystem marker written by the hook) that the hook ran, that the agent started afterwards, and that Agent Control reports the sub-agent healthy over OpAMP.
- post-download-hook-failure — package ships a hook that fails (non-zero exit, writes to stderr); verifies the hook ran but the agent did not start, and that Agent Control reports the sub-agent unhealthy with the hook's stderr surfaced in the health error over OpAMP.

 **_Note on agent.sh_**: these scenarios use a trivial script (touch a marker, then sleep) as the sub-agent instead of a real agent, on purpose. The feature under test is the post_download_hook mechanism, which is agent-agnostic  the e2e still exercises the real path (AC install + OCI download + signature verification + extraction + hook execution + systemd). A trivial executable keeps it hermetic and deterministic (exact control over hook success/failure, verified via a filesystem marker + sub-agent health over OpAMP, no backend/NRQL). A real agent would be more contrived (real agent types don't define a hook) and would couple this to backend reporting. Windows is already covered by the integration tests

### DIFF
--- a/test/e2e-runner/src/linux/scenarios/post_download_hook.rs
+++ b/test/e2e-runner/src/linux/scenarios/post_download_hook.rs
@@ -1,0 +1,270 @@
+use crate::common::oci::{OciRegistry, PushedPackage, push_hook_package};
+use crate::common::on_drop::CleanUp;
+use crate::common::runtime::tokio_runtime;
+use crate::common::test::{TestResult, retry_panic};
+use crate::common::{InstallationArgs, RecipeData, config};
+use crate::linux;
+use crate::linux::install::{install_agent_control_from_recipe, tear_down_test};
+use crate::linux::service::{STATUS_RUNNING, restart_service_and_wait};
+use fake_opamp_server::FakeServer;
+use std::fs;
+use std::path::Path;
+use std::time::Duration;
+use tracing::info;
+
+const AGENT_ID: &str = "posthook-agent";
+const AGENT_TYPE: &str = "newrelic/com.newrelic.posthook_e2e:0.1.0";
+const AGENT_TYPE_FILE: &str = "/etc/newrelic-agent-control/dynamic-agent-types/posthook_e2e.yaml";
+
+const SUPERVISOR_KEY_ATTR: &str = "supervisor.key";
+const HOOK_ERROR_MSG: &str = "e2e-posthook-failed-on-purpose";
+
+// Installs Agent Control with a sub-agent whose OCI package defines a successful
+// `post_download_hook`, and verifies (via filesystem markers) that the hook runs and the agent
+// starts afterwards.
+pub fn test_post_download_hook_success(args: InstallationArgs) {
+    info!("Starting post-download-hook success scenario");
+
+    let registry = OciRegistry::start();
+
+    let (hook_marker, agent_marker) = markers();
+
+    // The hook succeeds; the agent then starts and writes its own marker.
+    let hook_script = format!("#!/bin/bash\ntouch {hook_marker}\nexit 0\n");
+    let agent_script = format!("#!/bin/bash\ntouch {agent_marker}\nexec sleep 3600\n");
+    let pushed = push_hook_package(&hook_script, &agent_script);
+
+    let mut opamp_server = FakeServer::start(tokio_runtime().handle());
+    let _clean_up = CleanUp::new(tear_down_test);
+
+    deploy_hook_agent(args, &registry, &pushed, &mut opamp_server);
+
+    info!("Verifying the post-download hook ran (hook marker created)");
+    retry_panic(
+        60,
+        Duration::from_secs(2),
+        "post-download hook marker",
+        || marker_exists(&hook_marker),
+    );
+
+    info!("Verifying the agent started after the hook (agent marker created)");
+    retry_panic(60, Duration::from_secs(2), "agent started marker", || {
+        marker_exists(&agent_marker)
+    });
+
+    info!("Verifying Agent Control reports the sub-agent as healthy via OpAMP");
+    retry_panic(60, Duration::from_secs(2), "sub-agent healthy", || {
+        let instance = opamp_server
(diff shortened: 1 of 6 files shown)


test/e2e-runner/src/linux/scenarios/post_download_hook.rs
```diff
@@ -0,0 +1,270 @@
...
+const AGENT_TYPE: &str = "newrelic/com.newrelic.posthook_e2e:0.1.0";
+const AGENT_TYPE_FILE: &str = "/etc/newrelic-agent-control/dynamic-agent-types/posthook_e2e.yaml";
+
+const SUPERVISOR_KEY_ATTR: &str = "supervisor.key";
+const HOOK_ERROR_MSG: &str = "e2e-posthook-failed-on-purpose";
+
+// Installs Agent Control with a sub-agent whose OCI package defines a successful
+// `post_download_hook`, and verifies (via filesystem markers) that the hook runs and the agent
+// starts afterwards.
+pub fn test_post_download_hook_success(args: InstallationArgs) {
```
danielorihuela, line 25: The difference between this test and the integration test `test_postdownload_hook_successful_execution_with_oci_registry`, is that we check the existence of a marker.

Couldn't we do that in the integration test and get rid of the e2e? Am I missing something?

test/e2e-runner/src/linux/scenarios/post_download_hook.rs
```diff
@@ -0,0 +1,270 @@
...
+        }
+    });
+
+    info!("Post-download-hook success scenario completed successfully");
+}
+
+// Installs Agent Control with a sub-agent whose OCI package defines a failing
+// `post_download_hook`, and verifies that the hook runs but the agent is never started because
+// the hook returns a non-zero exit code.
+pub fn test_post_download_hook_failure(args: InstallationArgs) {
```
danielorihuela, line 78: Same as above but with `test_postdownload_hook_failure_prevents_agent_start_with_oci_registry`.

test/e2e-runner/src/linux/scenarios/post_download_hook.rs
vjripoll: You're right that the marker check itself could live in the integration test, and the hook logic is the same. The only thing the e2e adds is running the hook under the real .deb-installed AC managed by systemd (real paths/permissions/cwd), which the in-process integration test can't exercise. 
If we don't think is worth it, I'm happy to move it to the integration test and drop the e2e

danielorihuela: Mmm. I don't know.

* `deb` installed -> I might be too naive. Does this imply anything for this feature? I mean, we just run the script after install. Nothing about the package manager, not the systemd comes into play.
* real paths -> I think using a temporal directory is mostly the same.
* permissions -> `make_executable_if_exists` is called on `execute`, and this is run on both places, isn't it?
* cwd -> The code is generalized. So it should work both on integration and e2e.

Let's see if someone else wants to chime in. Maybe @paologallinaharbur or @gsanchezgavier.
I'm okay with e2e tests by the way. If these are adding value, I'll approve it right away.

gsanchezgavier: IMO 
- I would only keep the happy path (success test) in the e2e as they cost to much.
- Most of these test try to use "real" agent scenarios , and we tests this kind of features in integration tests, otherwise the current e2e solution will collapse soon. We have some exceptions for things like self-upgrade in onhost , which is a feature that involves systemd etc and is super critical. So I lean towards what @danielorihuela  mentions

gsanchezgavier: BTW if the test were not so expensive i would love to cover more scenarios , so we could always thing to improve our harness

danielorihuela: I don't think that's the real problem. Look into https://github.com/newrelic/newrelic-agent-control/actions/runs/28429660817. The time spent is the same. It can be computed with: build time + slowest test.

All tests are done in parallel, so the time spent by tests is dictated by the slower test. It doesn't matter if we add a 1000 tests more. As long as they are faster than the current slowest test, the time will remain the same.

My point was that these 2 new e2e tests don't add anything of value. At least for me, they seem exactly the same as the integration tests. We just change a temporal folder with a non temporal folder. We don't care about services, nor about getting info from the backend. Besides, we say

> Windows is already covered by the integration tests

So... why are integration tests enough for windows but not linux?

vjripoll: Well, we actually have two options because right now, the E2E doesn't 100% reflect what a real E2E should be. Since we don't have any real agent with the postdownload script in the package right now, if an agent has the script in the future, we can use it for the installation and see if it works, but it comes with its pros and cons. Right now, it brings very little value compared to integration tests.

1. Add these tests now and wait until we have an agent with a real script to change it.

2. Wait for the agent with the script directly."

danielorihuela (approved):

test/e2e-runner/src/linux/scenarios/post_download_hook.rs
danielorihuela: I don't think we need a real agent. The agent you are pushing to the local oci registry is already a real agent. I just don't feel it's adding much more than the integration test.

I also find strange that we only add the e2e for linux but not for windows. Either they are enough on both OSes or not.

I don't want to be a blocker, and the tests are all right in its current shape.

gsanchezgavier: agree if we are covering this with integration tests let's not merge this . Also let's make sure the feature is tested in both OS

PR #2640, opened 2026-06-29
fix: only one thread for oci_registry tests

alvarocabanas:
Fixes intermittent Windows CI failures in test_ac_self_update_with_oci_registry where self-update reported Failed instead of Applied (Access is denied). 
The integration tests run AC in-process, so self-replace was rewriting the shared test-harness binary; the self-replace target is now injectable so each test replaces its own disposable copy. Also surfaces the remote-config error_message on assertion mismatch for easier diagnosis.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ Remember that the keywords that you can use are the following:
 - Added support for remote agent type definition retrieval from OCI registries.
 - Agent type definitions now tolerate unknown fields for forward compatibility.
 - Replace filesystem in on-host agent-type definitions with an explicit, recursive, tagged-kind tree: every entry declares `kind: file | dir | dir_content_from_map`, and `dir` entries nest via `entries:`.
+- Added support to set a different self-update binary target.
 
 ## v1.17.0 - 2026-06-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,6 @@ Remember that the keywords that you can use are the following:
 - Added support for remote agent type definition retrieval from OCI registries.
 - Agent type definitions now tolerate unknown fields for forward compatibility.
 - Replace filesystem in on-host agent-type definitions with an explicit, recursive, tagged-kind tree: every entry declares `kind: file | dir | dir_content_from_map`, and `dir` entries nest via `entries:`.
-- Added support to set a different self-update binary target.
 
 ## v1.17.0 - 2026-06-16
 
--- a/agent-control/tests/on_host/scenarios/ac_self_update.rs
+++ b/agent-control/tests/on_host/scenarios/ac_self_update.rs
@@ -1,4 +1,5 @@
 use crate::common::agent_control::start_agent_control_with_custom_config;
+use crate::common::agent_control::start_agent_control_with_self_replace_target;
 use crate::common::base_paths::TempBasePaths;
 use crate::common::remote_config_status::check_latest_remote_config_status;
 use crate::common::remote_config_status::check_latest_remote_config_status_is_expected;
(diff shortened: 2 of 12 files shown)


agent-control/tests/on_host/scenarios/ac_self_update.rs
```diff
@@ -399,16 +417,17 @@ agents: {{}}
...
+    // self-replace target.
     let installed_packages_dir = dirs
         .remote_dir()
         .join(PACKAGES_FOLDER_NAME)
         .join(AGENT_CONTROL_ID)
         .join("stored_packages")
         .join(AGENT_CONTROL_BIN_PACKAGE_ID);
     retry(60, Duration::from_secs(5), || {
-        if dir_contains_file(&installed_packages_dir, AGENT_CONTROL_BIN) {
+        if dir_contains_subdir(&installed_packages_dir) {
```
sigilioso: Is checking that the directory contains any subdirectory a meaningful assertion? Maybe now we could use the new `self_replace_target` and check that it has changed or something. WDYT?

self-replacer/src/replacer.rs
```diff
@@ -54,17 +54,43 @@ pub enum ReplaceError {
...
 #[derive(Debug)]
-pub struct BinarySelfReplacer;
+pub struct BinarySelfReplacer {
+    target: PathBuf,
+}
+
+impl BinarySelfReplacer {
+    /// Builds a replacer targeting the currently running executable.
+    /// Canonicalize resolves symlinks on both platforms.
+    pub fn new() -> Result<Self, ReplaceError> {
```
sigilioso, line 74: The behavior has slightly changed:
- Before: we got the executable path right before replacing it.
- Now: we get and store the executable path when the _Replacer_ is created.

It is **not** an issue in any case, right? (I doubt that the self-replace mechanism wold work if the binary was moved from its expected location)

self-replacer/src/replacer.rs
```diff
@@ -54,17 +54,43 @@ pub enum ReplaceError {
 }
 
 /// Platform-agnostic self-replacer implementation.
+/// Holds the path of the binary to replace.
 #[derive(Debug)]
-pub struct BinarySelfReplacer;
+pub struct BinarySelfReplacer {
+    target: PathBuf,
+}
```
sigilioso: Really nit: now that we can built it using `BinarySelfReplacer::with_target(...)` I'm not sure if _self_ should still be part of the struct name 😅

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ Remember that the keywords that you can use are the following:
 - Added support for remote agent type retrieval.
 - add post-download script support for OCI packages.
 - Replace filesystem in on-host agent-type definitions with an explicit, recursive, tagged-kind tree: every entry declares `kind: file | dir | dir_content_from_map`, and `dir` entries nest via `entries:`.
+- Added support to set a different self-update binary target.
```
sigilioso: The support is completely internal right? (not exposed in _production_ code), does this belong to the changelog?

sigilioso (commented):
Injecting the path seems to be worth it as it makes tests easier. Thanks for taking care of it!

I've left some comments, the only important doubt is the one regarding the new test assertion.

self-replacer/src/replacer.rs
alvarocabanas: It's not an issue because the path only changes if the running binary is moved or renamed during the process and self-replace would fail.

agent-control/tests/on_host/scenarios/ac_self_update.rs
alvarocabanas: Done in last commit.

self-replacer/src/replacer.rs
alvarocabanas: Changed in last commit.

CHANGELOG.md
alvarocabanas: Removed in last commit.

sigilioso (approved):
LGTM

PR #2642, opened 2026-06-29
feat: allow backward-compatible protocol versions

gsanchezgavier:
Agent Control previously accepted only agent-type files whose `protocol_version` shared the same major as the supported version, rejecting older majors outright. This loosens the check so any version up to and including the supported one is accepted, treating `protocol_version` as a single ordered `MAJOR.MINOR` value.

### DIFF
--- a/agent-control/src/agent_type/protocol_version.rs
+++ b/agent-control/src/agent_type/protocol_version.rs
@@ -6,13 +6,15 @@
 //! validated against the supported version at the registry ingestion boundary before the rest of
 //! the file is parsed.
 //!
-//! Compatibility rules for a file's version against the supported version:
-//! - different `major` (either direction) → rejected: a major bump is a breaking schema change.
-//! - same `major`, higher `minor` → rejected: the file is newer than this Agent Control understands.
-//! - same `major`, equal or lower `minor` → accepted: minor bumps are additive and backward-compatible.
+//! The `protocol_version` is treated as a single ordered `MAJOR.MINOR` value. Compatibility against
+//! the supported version is:
+//! - version newer than supported (higher `major`, or same `major` with higher `minor`) → rejected:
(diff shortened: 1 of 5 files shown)


agent-control/src/agent_type/protocol_version.rs
```diff
@@ -6,13 +6,15 @@
...
-//! Compatibility rules for a file's version against the supported version:
-//! - different `major` (either direction) → rejected: a major bump is a breaking schema change.
-//! - same `major`, higher `minor` → rejected: the file is newer than this Agent Control understands.
-//! - same `major`, equal or lower `minor` → accepted: minor bumps are additive and backward-compatible.
+//! The `protocol_version` is treated as a single ordered `MAJOR.MINOR` value. Compatibility against
+//! the supported version is:
+//! - version newer than supported (higher `major`, or same `major` with higher `minor`) → rejected:
+//!   the file is newer than this Agent Control understands.
+//! - version equal to or older than supported → accepted: Agent Control understands every protocol
+//!   version up to and including the supported one.
```
sigilioso, line 14: This means we don't have a way to fully _deprecate_ Agent Type versions anymore, but I don't think that's an issue. We will cross that bridge when we get there (if we ever get there!)

sigilioso (dismissed):
LGTM

agent-control/src/agent_type/protocol_version.rs
gsanchezgavier: yes is more lax than the before which as a positive side i think it will be a feature more prone to be used. Also we can make it more strict if happens that we want to deprecate older major versions. 
The negative side is ac complexity

danielorihuela (approved):

sigilioso (approved):

PR #2644, opened 2026-06-30
feat(fs): introduce copy operation [NR-579370]

sigilioso:
## Context

This is the first step  support of shared filesystem for OHI agent types. 

To let an OHI agent type place an integration **binary** in a shared location the infra-agent can execute, Agent Control needs to *copy* a file from one path (e.g. an extracted OCI package) to another and guarantee it ends up **executable**.

This PR adds the missing primitive to copy files. It is the first, self-contained step of the shared-filesystem work and is **not wired into any runtime path yet**. Later PRs will consume it.

## Changes

- A new `FileCopier` trait (impl'd on `LocalFile`).
- A redundant comment regarding implementations is removed.

> [!NOTE]
> `std::fs::copy` copies the bytes of the binary, including the bits representing permissions on Linux, however on Windows permissions are handled differently and the corresponding ACL is set (same as write operations).

### DIFF
--- a/fs/src/file/copier.rs
+++ b/fs/src/file/copier.rs
@@ -4,48 +4,32 @@ use std::path::Path;
 use std::{fs, io};
 use tracing::instrument;
 
-/// Copies a file on disk, setting permissions on the destination.
+/// Copies a file on disk, preserving the source's permissions on Unix.
 pub trait FileCopier {
     /// Copies the file at `from` to the file at `to`, creating or truncating `to`.
     ///
-    /// The copy is byte-for-byte, so non-UTF-8 binaries are preserved exactly. On Unix the
-    /// destination's permissions are set to `mode` (for example `0o755` for an executable),
-    /// overriding whatever the source carried. On Windows `mode` is ignored and the destination
-    /// is restricted to administrators, matching [`FileWriter::write`](super::writer::FileWriter).
+    /// The copy is byte-for-byte, so non-UTF-8 binaries are preserved exactly. On Unix it
+    /// **preserves the source file's mode bits**.
     ///
-    /// `mode` is kept in the signature on every platform on purpose: it expresses the caller's
-    /// Unix permission intent (Linux is the primary on-host target), and a single cross-platform
-    /// signature lets callers state that intent uniformly. Windows uses an ACL-based model with no
-    /// Unix mode to honor, so the value is intentionally ignored there rather than dropped from the
-    /// API.
+    /// On Windows, permissions cannot be preserved this way (the copy does not carry the source's
+    /// ACL, and executability is determined by extension, not permissions), so the destination is
+    /// restricted to administrators, matching [super::writer::FileWriter].
     ///
     /// The destination's parent directory must already exist.
-    fn copy(&self, from: &Path, to: &Path, mode: u32) -> io::Result<()>;
+    fn copy(&self, from: &Path, to: &Path) -> io::Result<()>;
 }
 
 impl FileCopier for LocalFile {
-    /// Copies `from` to `to` byte-for-byte and sets the destination's permissions.
-    /// On Unix the destination mode is set to `mode`; on Windows access is restricted to
-    /// administrators and `mode` is ignored.
+    /// Copies `from` to `to` byte-for-byte, preserving the source's mode on Unix and applying the
+    /// administrators-only ACL on Windows.
     #[instrument(skip_all, fields(from = %from.display(), to = %to.display()))]
-    fn copy(&self, from: &Path, to: &Path, mode: u32) -> io::Result<()> {
+    fn copy(&self, from: &Path, to: &Path) -> io::Result<()> {
         validate_path(to)?;
 
         fs::copy(from, to)?;
 
-        #[cfg(target_family = "unix")]
-        {
-            use std::os::unix::fs::PermissionsExt;
-
-            fs::set_permissions(to, fs::Permissions::from_mode(mode))?;
-        }
-
         #[cfg(target_family = "windows")]
         {
-            // Windows has no Unix mode to apply: file access is governed by ACLs, which we set the
-            // same way `FileWriter::write` does. `mode` is accepted for a uniform cross-platform
-            // signature but has no Windows equivalent, so it is intentionally unused here.
-            let _ = mode;
             crate::win_permissions::set_file_permissions_for_administrator(to)
                 .map_err(io::Error::other)?;
         }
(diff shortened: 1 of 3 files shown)


fs/src/file/copier.rs
```diff
@@ -0,0 +1,149 @@
...
+    /// is restricted to administrators, matching [`FileWriter::write`](super::writer::FileWriter).
+    ///
+    /// `mode` is kept in the signature on every platform on purpose: it expresses the caller's
+    /// Unix permission intent (Linux is the primary on-host target), and a single cross-platform
+    /// signature lets callers state that intent uniformly. Windows uses an ACL-based model with no
+    /// Unix mode to honor, so the value is intentionally ignored there rather than dropped from the
+    /// API.
+    ///
+    /// The destination's parent directory must already exist.
+    fn copy(&self, from: &Path, to: &Path, mode: u32) -> io::Result<()>;
```
gsanchezgavier: would make sense to implement something like "preserve mode" ?

fs/src/file/copier.rs
```diff
@@ -0,0 +1,149 @@
...
+    fn copy(&self, from: &Path, to: &Path, mode: u32) -> io::Result<()> {
+        validate_path(to)?;
+
+        fs::copy(from, to)?;
+
+        #[cfg(target_family = "unix")]
+        {
+            use std::os::unix::fs::PermissionsExt;
+
+            fs::set_permissions(to, fs::Permissions::from_mode(mode))?;
```
gsanchezgavier: isn't a way to set permissions at creation time?

gsanchezgavier (commented):
thannks for chopping the prs!!

fs/src/file/copier.rs
sigilioso: Maybe it makes more sense. `std::fs::copy` would preserve the permission bits, so they should be "preserved by default" for Linux. On Windows we would need to keep doing the `crate::win_permissions::set_file_permissions_for_administrator(to)` because permissions are handled differently. Does it make sense to you?

fs/src/file/copier.rs
sigilioso: Yes, but we were copying. If we preserve the permissions as suggested in the previous comment this is not needed anymore.

fs/src/file/copier.rs
gsanchezgavier: sounds good to me

sigilioso: Updated here: [3095630](https://github.com/newrelic/newrelic-agent-control/pull/2644/commits/30956305f4802efeb5dc0a36b7b892ba0fca141e)

fs/src/file/copier.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// ACL, and executability is determined by extension, not permissions), so the destination is
+    /// restricted to administrators, matching [super::writer::FileWriter].
+    ///
+    /// The destination's parent directory must already exist.
+    fn copy(&self, from: &Path, to: &Path) -> io::Result<()>;
+}
+
+impl FileCopier for LocalFile {
+    /// Copies `from` to `to` byte-for-byte, preserving the source's mode on Unix and applying the
+    /// administrators-only ACL on Windows.
```
danielorihuela: I understand this is to make sure that we can execute the binary, right? I think this is what I read in some other comments.

What happens if the OCI package contains the binary with read only permissions, but not execution permissions? Did we take that into account?

fs/src/file/copier.rs
```diff
@@ -0,0 +1,141 @@
...
+            let src = dir.path().join("bin");
+            let dst = dir.path().join("copied-bin");
+
+            fs::write(&src, bytes).unwrap();
+            fs::set_permissions(&src, fs::Permissions::from_mode(mode)).unwrap();
+
+            LocalFile.copy(&src, &dst).unwrap();
+
+            let dst_mode = fs::metadata(&dst).unwrap().permissions().mode() & PERMISSION_BITS;
+            assert_eq!(fs::read(&dst).unwrap(), bytes, "content must be preserved");
```
danielorihuela, line 63: Nit: Leave only the permissions in the for loop.

Copying twice the same thing to the same place is not relevant in the for loop.

fs/src/file/copier.rs
```diff
@@ -0,0 +1,141 @@
...
+            assert_eq!(
+                dst_mode, mode,
+                "destination must preserve the source's mode {mode:#o}"
+            );
+        }
+    }
+
+    /// Copying onto an existing file replaces the existing
```
danielorihuela: Nit: Remove comment

test name already explains the situation

fs/src/file/copier.rs
```diff
@@ -0,0 +1,141 @@
...
+    #[cfg(unix)]
```
danielorihuela: Should we execute this tests in both linux and windows?

fs/src/file/copier.rs
sigilioso: > I understand this is to make sure that we can execute the binary, right? I think this is what I read in some other comments.

Also the files are restricted to admin only (same as other files written by AC).

> What happens if the OCI package contains the binary with read only permissions, but not execution permissions? Did we take that into account?

Permissions are not necessarily preserved when copying the file on Windows. We can highlight the behavior that in the Agent Type's copy documentation (when it is available in next PRs). WDYT?

fs/src/file/copier.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /// The destination path is validated before any I/O: `..` components are rejected.
+    #[test]
+    fn test_copy_rejects_parent_dir_in_destination() {
+        let dir = tempfile::tempdir().unwrap();
+        let src = dir.path().join("src");
+        fs::write(&src, "data").unwrap();
+
+        let result = LocalFile.copy(&src, Path::new("some/path/../../etc/passwd"));
```
danielorihuela, line 103: I just realized that this allows overwriting files like `/etc/passwd`, but... I don't know if we should put some guardrails in place.

This would be just a best effort, and not part of this PR. At the end of the day, this could also be done with a postinstall script.

sigilioso: I think we should, but probably not in this component. The tool implementing `copy_from` for agent type could include some checks in the destination folder (Eg: it needs to be in `shared_filesystem_dir`)

fs/src/file/copier.rs
sigilioso: We discussed it offline and realized that the explicitly set permissions were dropping execution for admin. We finally removed the Windows set-permissions thing here.

fs/src/file/copier.rs
sigilioso: I've just realized that the permissions bits are actually part of the files so those are actually different files to be copied (I need to copy twice to check that the permissions are preserved)

fs/src/file/copier.rs
sigilioso: Done!

fs/src/file/copier.rs
sigilioso: I've just removed the `#[cfg(unix)]` part

fs/src/file/copier.rs
danielorihuela: Oh right, this is another crate. I like your idea way more.

fs/src/file/copier.rs
danielorihuela: interesting

gsanchezgavier (approved):
thanks!

fs/src/file/copier.rs
```diff
@@ -4,48 +4,32 @@ use std::path::Path;
 use std::{fs, io};
 use tracing::instrument;
 
-/// Copies a file on disk, setting permissions on the destination.
+/// Copies a file on disk, preserving the source's permissions on Unix.
```
danielorihuela, line 7: Nit: Remove the unix part from here, it's already explained in the `copy`. Or maybe add windows explanation here. I don't know.

```rust
/// Copies a file on disk.
///
/// On unix, permissions are preserved.
/// On windows, destination is restricted to administrators.
```

fs/src/file/copier.rs
```diff
@@ -4,48 +4,32 @@ use std::path::Path;
...
-    /// `mode` is kept in the signature on every platform on purpose: it expresses the caller's
-    /// Unix permission intent (Linux is the primary on-host target), and a single cross-platform
-    /// signature lets callers state that intent uniformly. Windows uses an ACL-based model with no
-    /// Unix mode to honor, so the value is intentionally ignored there rather than dropped from the
-    /// API.
+    /// On Windows, permissions cannot be preserved this way (the copy does not carry the source's
+    /// ACL, and executability is determined by extension, not permissions), so the destination is
+    /// restricted to administrators, matching [super::writer::FileWriter].
     ///
     /// The destination's parent directory must already exist.
```
danielorihuela, line 18: Those are implementation details. Does it make sense to explain them in the trait? I always don't know what to do in these situations.

fs/src/file/copier.rs
```diff
@@ -0,0 +1,141 @@
...
+
+/// Copies a file on disk, keeping the destination's permissions consistent with the source.
+pub trait FileCopier {
+    /// Copies the file at `from` to the file at `to`, creating or truncating `to`.
+    ///
+    /// The copy is byte-for-byte, so non-UTF-8 binaries are preserved exactly, and the destination
+    /// keeps the permissions a plain copy yields: on Unix the source's mode bits are carried over
+    /// (e.g. the executable bit); on Windows the destination takes the default ACL for its location.
+    ///
+    /// The destination's parent directory must already exist.
```
danielorihuela, line 15: Should these lines be here or on the implementation docstrings. I never know what to do in those situations.

fs/src/file/copier.rs
```diff
@@ -0,0 +1,141 @@
...
+        assert!(
+            result
+                .unwrap_err()
+                .to_string()
+                .contains("dots disallowed in path")
+        );
+    }
+
+    /// A copied executable must stay runnable: copy the running test binary and launch the copy.
+    /// The earlier implementation dropped `FILE_EXECUTE`, which this would have caught.
```
danielorihuela, line 115: Nit: Unnecessary comment.

danielorihuela (approved):

PR #2645, opened 2026-06-30
refactor: k8s scenarios config build

danielorihuela:
Add a k8s config builder for scenario testing. Similar to onhost.

## Context
Each K8s integration test scenario previously required a dedicated `local-data-agent-control.template` file in `tests/k8s/data/<test-name>/`. The template used placeholder tokens (`<ns>`, `<opamp-endpoint>`, etc.) that were substituted at runtime by `create_local_agent_control_config`. This meant adding a new test required creating a new data directory and template file, and the substitution logic was brittle — the test name had to match the folder name exactly.

With the new approach, each test has the information clearly stated in it.

## Technical details
Introduces `K8sAgentControlConfigBuilder` in `tests/k8s/tools/config.rs`, mirroring the existing `AgentControlConfigBuilder` used in on-host tests. The builder composes YAML blocks from typed fields (fleet control, agents, k8s options, secrets providers, etc.) and writes the result to both the ConfigMap in the cluster and the local filesystem via its `write(client, dir)` method.

`start_agent_control_with_testdata_config` is removed; callers now call `K8sAgentControlConfigBuilder::new(namespace)` to build config, then `start_agent_control(...)` to launch the process. Sub-agent ConfigMaps that were previously read from `.yaml` data files are now created inline with `create_config_map`.

The common parts shared between on-host and K8s config (fleet control endpoint, agents block, status server) are extracted into `AgentControlCommonConfigBuilder` in `tests/common/config.rs`, which `K8sAgentControlConfigBuilder` composes.

The rename of `AgentControlConfigBuilder::basic` → `::new` and `AgentControlConfigBuilder` → `OnHostAgentControlConfigBuilder` in on-host tests is included here as it was the same commit scope.


agent-control/tests/k8s/tools/config.rs
```diff
@@ -0,0 +1,198 @@
...
+        self
+    }
+
+    pub fn with_namespace_agents(mut self, namespace_agents: impl Into<String>) -> Self {
+        self.namespace_agents = Some(namespace_agents.into());
+        self
+    }
+
+    // Clippy complains about dead code in windows tests, but this is used in Linux tests.
+    #[allow(dead_code)]
```
sigilioso: Another alternative would be using `#[cfg(unix)]`, right?

agent-control/tests/k8s/tools/config.rs
```diff
@@ -0,0 +1,198 @@
...
+        self.cr_type_meta = Some(cr_type_meta.into());
+        self
+    }
+
+    pub fn with_secrets_providers(mut self, secrets_providers: impl Into<String>) -> Self {
+        self.secrets_providers = Some(secrets_providers.into());
+        self
+    }
+
+    pub fn write(self, client: Client, local_dir: &Path) {
```
sigilioso, line 133: Maybe a little explanation of the reason why we need to write the content both in the ConfigMap and the file will avoid future some guessing/digging.

agent-control/tests/k8s/scenarios/without_flux.rs
```diff
@@ -143,45 +145,24 @@ agents: {}
...
-///
-/// The test:
-/// 1. Starts AC and deploys the config-map-type agent via OpAMP remote config.
-/// 2. Sends a remote config to the sub-agent, which causes AC to store the remote config
-///    and write the `owned-by: agent-control` and `agent-type-id` annotations onto the
-///    fleet-data ConfigMap.
-/// 3. Waits for the annotation to be present, then stops and restarts AC.
-/// 4. On restart, `retain` finds the annotated fleet-data ConfigMap, recognises it as an
-///    Agent Control internal resource, and correctly retains it.
-/// 5. The test passes only if AC starts successfully.
```
sigilioso, line 164: isn't this context relevant anymore?

sigilioso (commented):
Could we double-check if all removed comments are supposed to be removed?

Everything else looks good to me. Great refactor, integration tests are way more maintainable now!

agent-control/tests/k8s/tools/config.rs
danielorihuela: https://github.com/newrelic/newrelic-agent-control/pull/2645/changes/18e7766e1b8f616c0bcd302cbf986ef141a82f33

agent-control/tests/k8s/tools/config.rs
danielorihuela: https://github.com/newrelic/newrelic-agent-control/pull/2645/changes/18e7766e1b8f616c0bcd302cbf986ef141a82f33

agent-control/tests/k8s/scenarios/without_flux.rs
danielorihuela: https://github.com/newrelic/newrelic-agent-control/pull/2645/changes/18e7766e1b8f616c0bcd302cbf986ef141a82f33

sigilioso (approved):
🚀

PR #2646, opened 2026-06-30
feat: expose `${nr-sub:shared_filesystem_dir}` variable [NR-579370]

sigilioso:
## Summary

In order to add support to shared filesystem for OHI agent types. Sub-agents that need to share files (binaries, configs) with another agent require a **common, well-known location** to write them to, instead of each agent's isolated per-agent directory. This PR exposes such variable, but there are no operations to include content into it yet, those will come in the next PRs.

## Changes

- New variable added under `nr-sub` variables namespace.


vjripoll (approved):
LGTM

PR #2647, opened 2026-06-30
fix: remove cleaner on agent id change (NR-582547)

gsanchezgavier:
As part of implementing the packages cleaning hit back this and decided to remove it until we have a clear picture of what needs to be done here (whenever is going to be used).

For k8s we might have a situation where a new agent type version removes some resources so this logic makes sense, but on the other hand it doesn't seem right to remove everything in case a new patch version is updated. So there is probably some design needed here.

### DIFF
--- a/agent-control/src/agent_control.rs
+++ b/agent-control/src/agent_control.rs
@@ -542,14 +542,9 @@ where
                     debug!(%agent_id, "Retaining the existing running SubAgent as its type remains unchanged");
                     Ok(())
                 }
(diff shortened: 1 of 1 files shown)


sigilioso (approved):
LGTM

We need to remember that the new behavior could be _unexpected_: if a new version of an Agent Type defines different resources or the the `agent_id` is reused for a different Agent Type we could leave orphan resources. That said, I agree, the previous behavior (remove everything) wasn't well designed neither.

PR #2648, opened 2026-06-30
chore(health): allow no health

paologallinaharbur:
The health component should be Optional for onHost since not all Ohai will have a way to report their health.
Notice that such field was optional already for K8s agentTypes

### DIFF
--- a/agent-control/src/agent_type/runtime_config/on_host.rs
+++ b/agent-control/src/agent_type/runtime_config/on_host.rs
@@ -28,7 +28,7 @@ pub struct OnHost {
     enable_file_logging: TemplateableValue<bool>,
     /// Enables and define health checks configuration.
     #[serde(default)]
-    health: OnHostHealthConfig,
+    health: Option<OnHostHealthConfig>,
     #[serde(default)]
     filesystem: FileSystem,
     #[serde(default)]
@@ -79,7 +79,10 @@ impl Templateable for OnHost {
                 .map(|e| e.template_with(&extended_vars))
                 .collect::<Result<Vec<_>, _>>()?,
             enable_file_logging: self.enable_file_logging.template_with(&extended_vars)?,
-            health: self.health.template_with(&extended_vars)?,
+            health: self
+                .health
+                .map(|health| health.template_with(&extended_vars))
+                .transpose()?,
             filesystem: self.filesystem.template_with(&extended_vars)?,
             packages: rendered_packages,
         })
@@ -91,7 +94,6 @@ mod tests {
     use super::*;
 
     use crate::agent_type::agent_attributes::AgentAttributes;
-    use crate::agent_type::runtime_config::health_config::HealthCheckTimeout;
     use crate::agent_type::runtime_config::on_host::executable::{Args, Env};
     use crate::agent_type::runtime_config::on_host::package::{Download, Oci};
     use crate::agent_type::runtime_config::restart_policy::{
@@ -100,7 +102,6 @@ mod tests {
     };
     use crate::agent_type::variable::Variable;
     use crate::agent_type::variable::namespace::Namespace;
-    use crate::checkers::health::health_checker::{HealthCheckInterval, InitialDelay};
     use serde_json::Number;
     use std::collections::HashMap;
     use std::path::PathBuf;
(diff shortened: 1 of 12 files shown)


sigilioso:
Did we finally decide to keep using [default capabilities](https://github.com/newrelic/newrelic-agent-control/blob/4a56d3f4fe16759cae738dfc66751145059d6a8c/agent-control/src/agent_control/defaults.rs#L162-L171) for these agents (which include `ReportsHealth`) even if we know they will never report health?

agent-control/src/agent_type/runtime_config/on_host.rs
```diff
@@ -27,7 +27,7 @@ pub struct OnHost {
     enable_file_logging: TemplateableValue<bool>,
     /// Enables and define health checks configuration.
     #[serde(default)]
-    health: OnHostHealthConfig,
+    health: Option<OnHostHealthConfig>,
```
sigilioso, line 31: Because `OnhostHealthConfig` implements default we'll have:

- No health definition --> None
- `health: {}` --> Default
- `health:` --> Default?
- `health: null` None?

Should we document this explicitly in the Agent Type documentation? Would it make sense to have some unit tests to also act as internal docs?

sigilioso (commented):
Leaving the capabilities thing aside, I've only found something related to defaults (just a suggestion to extend docs/tests). Everything else looks good!

agent-control/tests/on_host/tools/custom_agent_type.rs
```diff
@@ -37,7 +37,16 @@ fake_variable:
...
+            health: Some(
+                serde_saphyr::from_str(
+                    r#"
+interval: 60s
+initial_delay: 0s
+timeout: 15s
+"#,
+                )
+                .unwrap(),
+            ),
```
danielorihuela, line 49: A lot of tests were running with none health. Why do we need that change now?

paologallinaharbur: Since health None was actually default health that was reporting true!

agent-control/src/agent_type/runtime_config/on_host.rs
danielorihuela: The first one. Serde just calls the `Default` impl of the type T. In that case we have an `T = Option<OnHostHealthConfig>`. https://doc.rust-lang.org/std/option/enum.Option.html#impl-Default-for-Option%3CT%3E

danielorihuela (dismissed):

agent-control/src/agent_type/runtime_config/on_host.rs
sigilioso: Were you explaining the first the item I mentioned or something else?

I wanted to highlight that the behavior won't be obvious for Agent Type writers, maybe the listing was confusing and the yaml is clearer:

```yaml
# health is not defined at all --> `None` because of what was stated in Daniel's comment
---
health: {} # Empty value (it will use the Default implementation for OnHostHealthConfig)
---
health: # I don't remember if this "empty" is the same as `{}` or not
---
health: null # I think this is equivalent to not defining the field
```

paologallinaharbur: I think we should remove the default!

danielorihuela: Aaaah. I thought the question was which one of the 4 was going to happen. It's clearer now.
Maybe we need docs yes.

sigilioso: > I think we should remove the default!

It probably simplifies things!

paologallinaharbur: I've removed the default and I've noticed that we were not actually checking anywhere in unit test the health threads and I added a test to cover that

sigilioso:
Just to be extra sure. Do we need to update any Agent Type yaml to keep the same behavior after removing introducing `None` and removing the default? I'm worried about accidentally disabling health check for any of the existing agents.

sigilioso (dismissed):
LGTM! 🚀 

As discussed _offline_, let's try to improve the YAML UX of the on-host health-checker in a different PR

sigilioso (approved):

PR #2649, opened 2026-06-30
fix: publish agent.version before package download and reset OCI test registry

alvarocabanas:
Fix two flaky OCI integration tests at their root: publish agent.version from config before the (slow, contended) package download so it no longer races the download window, and wipe the persisted zot registry storage before each run so tests that assume a version is absent start hermetic.                                                                                                                                                                 
Also clears accumulated registry bloat that was slowing concurrent pulls.

### DIFF
--- a/agent-control/Makefile
+++ b/agent-control/Makefile
@@ -30,6 +30,7 @@ test/onhost/root/integration:
 test/onhost/oci-registry/integration:
 	@echo "testing-machine-id" > /tmp/machine-id
 	@../tools/oci-registry.sh install
+	@../tools/oci-registry.sh clean
 	@../tools/oci-registry.sh run > /dev/null 2>&1 &
 	$(TEST_FIXTURES) cargo $(CARGO_CMD) $(PACKAGE_SA) --features=multiple-instances --test integration_tests _with_oci_registry -- --ignored --skip no_root --skip k8s
 	@../tools/oci-registry.sh killall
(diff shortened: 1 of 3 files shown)


vjripoll (dismissed):

agent-control/Makefile
```diff
@@ -38,6 +39,7 @@ test/onhost/oci-registry/integration:
 test/onhost/oci-auth-registry/integration:
```
gsanchezgavier, line 39: ohh these two guys where the ones colliding! now makes sense !

tools/oci-registry.sh
```diff
@@ -166,13 +167,21 @@ function run() {
...
     echo "Invalid option: $OPTION"
     print_usage
     exit 1
 elif [[ $OPTION == "help" || $OPTION == "empty" ]]; then
     print_usage
     exit 0
+elif [[ $OPTION == "clean" ]]; then
+    # Wipe persisted registry storage so each test run starts from an empty registry. zot keeps its
+    # data under STORAGE_PATH across runs; leftover tags break tests that assume a version is absent
+    # (e.g. the self-update recovery test) and bloat the registry, slowing concurrent pulls.
```
gsanchezgavier: i would remove this comment , the code is already clear

gsanchezgavier (dismissed):
thanks!

vjripoll (approved):

PR #2650, opened 2026-07-01
refactor: extract manage-paths reconcile into shared module [NR-579370]

sigilioso:
## Summary

The shared-filesystem will need the same "record what we wrote, delete what we no longer declare" reconcile that the per-agent-filesystem already has (added in #2618). Rather than duplicate that logic, this PR extracts it into a shared module both filesystems can use. There is **no behavior change** the refactor avoids duplication in the upcoming shared-filesystem feature.

### DIFF
--- a/agent-control/src/agent_type/runtime_config/on_host/managed_paths.rs
+++ b/agent-control/src/agent_type/runtime_config/on_host/managed_paths.rs
@@ -1,0 +1,250 @@
+//! Manifest-based reconcile shared by the rendered on-host filesystems.
+//!
+//! Agent Control records the absolute paths it wrote on the previous successful write in a JSON
+//! manifest, so the next write can delete paths that were "previously managed but are no longer
+//! declared". The manifest is read from an agent-writable location, so its entries are vetted
+//! against a base directory before any deletion.
+
+use serde::{Deserialize, Serialize};
+use std::collections::HashSet;
+use std::io;
+use std::path::{Component, Path, PathBuf};
+use tracing::{trace, warn};
+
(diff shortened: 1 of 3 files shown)


agent-control/src/agent_type/runtime_config/on_host/managed_paths.rs
```diff
@@ -0,0 +1,250 @@
...
+            .managed_paths
+            .into_iter()
+            .filter(|p| {
+                // Never let the manifest mark itself for deletion.
+                if p == &self.manifest_path {
+                    return false;
+                }
+                let within = is_within_base(p, &self.base_dir);
+                if !within {
+                    warn!(?p, base_dir = ?self.base_dir, "An agent is not allowed to modify files outside its isolated filesystem. Ignoring path.");
```
gsanchezgavier, line 81: is this message / feature correct? i see the base dir as the boundary not the "its isolated filesystem"

gsanchezgavier: maybe base dir is the agent dir , not sure, if so i think is confusing with the base_dir we used to define

gsanchezgavier (approved):

agent-control/src/agent_type/runtime_config/on_host/managed_paths.rs
sigilioso: I'll rephrase the comment because 'its isolated filesystem' is actually confusing.

`base_dir` is suppose to be:

- The agent-dir for the "regular" filesystem
- The shared filesystem when such feature comes

Should we use name different from `base_dir`?

PR #2654, opened 2026-07-01
feat(host): remove agent packages on clean

gsanchezgavier:
## Summary
On-host resource cleanup now also deletes an agent's installed OCI packages, so removing an agent from the fleet fully reclaims its on-disk footprint. 

## Changes
- Add `AgentPackagesRemover` trait; implement `remove_agent_packages` on `OCIPackageManager` (idempotent).
- `OnHostCleaner` gains a `package_remover` and deletes the agent's `packages/<agent_id>` dir on clean/purge.
- Share a single `Arc<OCIPackageManager>` between the cleaner and the supervisor builder.

### DIFF
--- a/agent-control/src/agent_control/resource_cleaner/on_host.rs
+++ b/agent-control/src/agent_control/resource_cleaner/on_host.rs
@@ -12,42 +12,50 @@ use crate::agent_control::agent_id::AgentID;
 use crate::agent_control::defaults::RESERVED_AGENT_IDS;
 use crate::agent_type::agent_type_id::AgentTypeID;
 use crate::opamp::instance_id::storer::{InstanceIDStorer, StorerError};
+use crate::package::manager::AgentPackagesRemover;
+use crate::package::oci::package_manager::OCIPackageManagerError;
 use crate::values::config_repository::{ConfigRepository, ConfigRepositoryError};
 
 use super::{ResourceCleaner, ResourceCleanerError};
 
 /// On-host implementation of [`ResourceCleaner`] that wipes a sub-agent's fleet data by
 /// delegating to the same storers that wrote it, also recursively deletes the sub-agent's
-/// dedicated filesystem directory, the `persistent` flag is bypassed here because the agent
-/// has been removed from the fleet.
+/// dedicated filesystem directory and its installed packages (via the [`AgentPackagesRemover`],
+/// which owns the on-disk package layout), the `persistent` flag is bypassed here because the
+/// agent has been removed from the fleet.
 /// The same removal logic is reused at startup by [`Self::purge_stale_agents`] to reclaim the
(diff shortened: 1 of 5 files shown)


agent-control/src/agent_control/resource_cleaner/on_host.rs
```diff
@@ -226,6 +268,7 @@ mod tests {
             fs_base(),
             fleet_base(),
             Arc::new(dir_manager),
+            Arc::new(remover_removing(&[id.as_str()])),
```
danielorihuela: Nit: Implement the helper to the Mock object.

Other mocks implement helper methods directly in the mock object. For example, `MockDirectoryManager::should_delete`. Why don't we implement a `should_delete` or `should_remove` function to `MockAgentPackagesRemover` and call it from here.

agent-control/src/package/manager.rs
```diff
@@ -46,6 +46,13 @@ pub trait PackageManager: Send + Sync {
     ) -> Result<(), OCIPackageManagerError>;
 }
 
+/// Removes every package an agent owns on disk.
+pub trait AgentPackagesRemover: Send + Sync {
+    /// Removes every package installed on disk for the given agent. A missing directory is not an
+    /// error.
+    fn remove_agent_packages(&self, agent_id: &AgentID) -> Result<(), OCIPackageManagerError>;
```
danielorihuela: Nit: Simplify method name to `remove` or `delete`.

The trait name already explain what it does, and it seems a one action trait.

agent-control/src/agent_control/resource_cleaner/on_host.rs
gsanchezgavier: yesss i'll do that, is a super generic helper so it makes sense

agent-control/src/package/oci/package_manager.rs
```diff
@@ -904,4 +937,43 @@ mod tests {
...
+        let pm = OCIPackageManager::new(
+            MockOCIDownloader::new(),
+            DirectoryManagerFs,
+            PathBuf::from(root_dir.path()),
+        );
+
+        pm.remove_agent_packages(&agent_id_to_remove).unwrap();
+
+        assert!(!get_agent_packages_dir(root_dir.path(), &agent_id_to_remove).exists());
+        assert!(other_package.exists());
```
danielorihuela, line 970: Nit: Maybe we can assert that the agent id was removed from the cache.

Not really needed, but it might help detecting any kind of issue in the future. I always wonder how useful are those and if we should do them at all.

```
assert_eq!(
    pm.latest_installed_packages
        .lock()
        .unwrap()
        .get(&agent_id_to_remove),
    None
);
```

danielorihuela (dismissed):

danielorihuela (approved):

PR #2659, opened 2026-07-02
feat: Sub-agents could be stopped and restarted unnecessarily on an AC Update

alvarocabanas:
Skip sub-agent reconciliation when an Agent Control self-update is in progress. `VersionUpdater::update` now returns an `UpdateOutcome`, and `RestartPending` (on-host self-replace) makes `handle_remote_config` store the new   
  config and return early instead of recreating sub-agents.                                                                                                                                                                         
                                                                                                                                                                                                                                    
This avoids a redundant double-restart of sub-agents (and the telemetry gap) when a single remote config both bumps the AC version and changes a sub-agent's `agent_type` — the restarted process re-applies the stored config  cleanly. K8s (Flux-driven) and no-op updaters keep their current behavior via `NoUpdate`.                                                                                                                                         
                                                                                                                                                                                                                                    
Adds an integration test asserting that on a pending self-update the changed sub-agent is not recreated, the running one is stopped once by the restart, and the new config is persisted.

## Checklist
 -->


agent-control/src/agent_control/version_updater/k8s.rs
```diff
@@ -71,13 +71,13 @@ impl<C: K8sClient> VersionUpdater for K8sACUpdater<C> {
...
                 );
-                return Ok(());
+                return Ok(UpdateOutcome::NoUpdate);
             };
         } else {
             debug!("Remote updates for Agent Control cd are disabled. Nothing to do.");
         }
 
-        Ok(())
+        Ok(UpdateOutcome::NoUpdate)
```
sigilioso: Even if the behavior is correct, the semantics are misleading, aren't them? Should we re-consider the `UpdateOutcome` enum varians?

agent-control/src/agent_control/version_updater/updater.rs
```diff
@@ -44,9 +56,10 @@ pub trait VersionUpdater {
     /// Verifies if the agent control version should be updated based on the provided configuration and
     /// attempts to update the desired agent control version.
     ///
-    /// Returns `Ok(())` if the desired version has been successfully communicated
-    /// to the external controller, or an `UpdaterError` if the update fails.
-    fn update(&self, config: &AgentControlDynamicConfig) -> Result<(), UpdaterError>;
+    /// Returns [`UpdateOutcome::RestartPending`] if a self-update was initiated and the process is
+    /// about to restart, [`UpdateOutcome::NoUpdate`] if nothing was done in-process, or an
+    /// `UpdaterError` if the update fails.
+    fn update(&self, config: &AgentControlDynamicConfig) -> Result<UpdateOutcome, UpdaterError>;
```
sigilioso, line 64: I wonder if we should flatten the result of this method.

We currently have:

```rust
Result { // Just to show Result<UpdateOucome, UpdateError>
  Ok(
      UpdateOutcome {
           NoUpdate,
            RestartPending,
        }
    ),
    Error(UpdateError)
}
```

We could use:

```rust
UpdateResult {
    NoUpdate,
    RestartPending,
    Error(UpdateError),
}

File: agent-control/src/agent_control.rs [hunk 3]
```diff
@@ -508,6 +508,17 @@ where
             };
             self.sa_dynamic_config_store.store(&config)?;
         }
+
+        // If a restart has been requested, reconciling sub-agents now would recreate any whose
+        // `agent_type` changed in this same config, only for them to be stopped ms later by the
+        // restart that applies it cleanly on startup.
```
sigilioso: Should we also mention that the remote config is already stored and consequently the new list of agents is  used when AC starts over?

agent-control/src/agent_control.rs
```diff
@@ -1633,6 +1651,79 @@ chart_version: 0.0.2 # not actually used, we rely on a mock
         assert_eq!(expected, ev);
     }
 
+    /// A self-update defers sub-agent reconciliation to the restarted process (no double restart)
+    /// when a *single* AC remote config triggers a self-update & changes a sub-agent's agent_type.
+    #[test]
+    fn test_self_update_defers_subagent_reconciliation_to_restart() {
```
sigilioso, line 1658: Have you considered checking the new behavior in integration tests? (where AC is actually restarted)

sigilioso (commented):
Thanks for addressing this, find some comments below

agent-control/src/agent_control/version_updater/updater.rs
DavSanchez: The disadvantage of doing this is that you lose much of the APIs already in place for the `Result` type (e.g. `inspect_err`, unwrapping or propagate with `?` and so on).

sigilioso: Good point! It wasn't the best idea 🙂

agent-control/src/agent_control/version_updater/k8s.rs
alvarocabanas: I modified the values in last commit, WDYT?

agent-control/src/agent_control.rs
alvarocabanas: Modified in last commit.

agent-control/src/agent_control.rs
alvarocabanas: I've added an integration test in last commit.

agent-control/src/agent_control/version_updater/k8s.rs
```diff
@@ -71,13 +71,13 @@ impl<C: K8sClient> VersionUpdater for K8sACUpdater<C> {
                 debug!(
                     "Agent Control CD release name is not set. Skipping Agent Control CD update."
                 );
-                return Ok(());
+                return Ok(UpdateOutcome::NoRestartPending);
```
danielorihuela: Why is this return necessary? Isn't the one on line 80 enough?

alvarocabanas: You are right, it's useless, I removed it in last commit.

agent-control/src/agent_control/version_updater/k8s.rs
```diff
@@ -43,7 +43,7 @@ pub struct K8sACUpdater<C: K8sClient = SyncK8sClient> {
 }
 
 impl<C: K8sClient> VersionUpdater for K8sACUpdater<C> {
-    fn update(&self, config: &AgentControlDynamicConfig) -> Result<(), UpdaterError> {
+    fn update(&self, config: &AgentControlDynamicConfig) -> Result<UpdateOutcome, UpdaterError> {
```
danielorihuela, line 46: For k8s we never return `RestartPending` because it's only an onhost thing?

alvarocabanas: Yes, there is this comment in UpdateOutcome, but nothing is said about RestartPending being only an onhost thing. I can add it.
```
pub enum UpdateOutcome {
    /// No in-process self-update was initiated. Continue applying the remote configuration.
    /// An update may still have been requested out-of-process (k8s Helm release version bump).
    NoRestartPending,
    ...
```

alvarocabanas: Added comments in latest commit.

danielorihuela (dismissed):

danielorihuela (approved):

PR #2660, opened 2026-07-02
chore(health): refactor

paologallinaharbur:
This PRs is changing the Health configuration:
 - to surface the "Process"  health check that was added under the hood
 - to be more similar to the K8s one:
```yaml
  health:
    interval: 30s
    initial_delay: 30s
    checks:
      - namespace: ${nr-ac:namespace}
        name: ${nr-sub:agent_id}
        kind: HelmReleaseWorkload
        target_namespace: ${nr-ac:namespace_agents}
```

therefore the configuration changes as follows

- old
```yaml
  health:
    interval: 30s
    initial_delay: 60s
    timeout: 5s
    http:
      path: "/v1/status/health"
      port: ${nr-var:health_port}
```

- new
```yaml
  health:
    interval: 30s
    initial_delay: 60s
    timeout: 5s
    checks:
      - kind: Process
      - kind: Http
        path: "/v1/status/health"
        port: ${nr-var:health_port}
```

Everything else should stay the same.
I decided to reject two "processes" checks since they are meainingless and we cannot duplicate the channel reader

### DIFF
--- a/agent-control/src/agent_type/runtime_config/health_config/rendered.rs
+++ b/agent-control/src/agent_type/runtime_config/health_config/rendered.rs
@@ -15,14 +15,19 @@ pub struct OnHostHealthConfig {
     pub(crate) initial_delay: InitialDelay,
     /// The maximum duration a health check may run before considered failed.
     pub(crate) timeout: HealthCheckTimeout,
-    /// Details on the type of health check. Defined by the `HealthCheck` enumeration.
-    pub(crate) check: Option<OnHostHealthCheck>,
+    /// The list of health checks to run. Empty means health reporting is disabled.
+    pub(crate) checks: Vec<OnHostHealthCheckDefinition>,
 }
 
+/// A single rendered on-host health check.
 #[derive(Debug, Clone, PartialEq)]
-pub(crate) enum OnHostHealthCheck {
-    HttpHealth(HttpHealth),
-    FileHealth(FileHealth),
+pub(crate) enum OnHostHealthCheckDefinition {
+    /// Process probe.
+    Process,
+    /// HTTP endpoint probe.
+    Http(HttpHealth),
+    /// File-based probe.
+    File(FileHealth),
 }
 
 #[derive(Debug, Clone, PartialEq)]
--- a/agent-control/src/checkers/health/on_host/health_checker.rs
+++ b/agent-control/src/checkers/health/on_host/health_checker.rs
@@ -2,12 +2,13 @@
 use super::exec::ExecHealthChecker;
 use super::file::FileHealthChecker;
 use super::http::HttpHealthChecker;
-use crate::agent_type::runtime_config::health_config::rendered::OnHostHealthCheck;
+use crate::agent_type::runtime_config::health_config::rendered::OnHostHealthCheckDefinition;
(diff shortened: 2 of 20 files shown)


agent-control/src/checkers/health/on_host/health_checker.rs
```diff
@@ -25,34 +26,58 @@ pub struct OnHostHealthCheckers {
 }
 
 impl OnHostHealthCheckers {
+    /// Builds the aggregate checker from the explicit list of check definitions.
+    ///
+    /// Returns `Ok(None)` when the list is empty: health reporting is then disabled for the
+    /// sub-agent. Returns `Err` only when a specific checker (e.g. `Http`) fails to initialize.
+    ///
+    /// The `Process` check is what surfaces the health of supervised executables; if it is not
+    /// declared, `exec_health_consumer` is dropped and per-executable health signals are ignored.
```
sigilioso: Nit: isn't this comment only relevant for the changes? The same happens with the other checkers, if not declared, not automatically set.

sigilioso (dismissed):
Thanks for addressing this!

I only found something related to a comment that can be perfectly ignored

agent-control/src/agent_type/runtime_config/health_config/rendered.rs
```diff
@@ -15,14 +15,19 @@ pub struct OnHostHealthConfig {
...
+    pub(crate) checks: Vec<OnHostHealthCheckDefinition>,
 }
 
+/// A single rendered on-host health check.
 #[derive(Debug, Clone, PartialEq)]
-pub(crate) enum OnHostHealthCheck {
-    HttpHealth(HttpHealth),
-    FileHealth(FileHealth),
+pub(crate) enum OnHostHealthCheckDefinition {
+    /// Health of the supervised executables.
```
danielorihuela: Nit: Change docstring with `Process probe`

Just so it's consistent with other variants.

agent-control/src/checkers/health/on_host/health_checker.rs
```diff
@@ -25,34 +26,58 @@ pub struct OnHostHealthCheckers {
...
-                    start_time,
-                )?));
-            }
-            Some(OnHostHealthCheck::FileHealth(file_config)) => {
-                health_checkers.push(OnHostHealthChecker::File(FileHealthChecker::new(
-                    PathBuf::from(file_config.path),
-                )));
+    ) -> Result<Option<Self>, HealthCheckerError> {
+        if checks.is_empty() {
+            debug!("No health checks configured: no health-checker is built");
```
danielorihuela, line 40: I wonder if this should be a `warning` instead, given that ideally all agents should have a way to measure it's health.

paologallinaharbur: I do not think so, it depends on the agentType and the user has no control over it.
For example if there is an ohai you cannot switch off such warning in any way

agent-control/src/checkers/health/on_host/health_checker.rs
paologallinaharbur: exec_health_consumer is dropped only if the process is not set, but you are right, it is kind of obvious

danielorihuela (dismissed):

sigilioso (dismissed):

sigilioso (approved):
LGTM

PR #2662, opened 2026-07-03
feat: ensure filesystem validation

alvarocabanas:
Adds parse-time validation of on-host agent-type filesystems that rejects any entry marked persistent: true whose parent directory chain isn't also persistent. This prevents a silent footgun: a non-persistent parent is deleted recursively on sub-agent stop, wiping persistent children regardless of their own flag, so the child's persistence never takes effect. Templated persistent values (resolved after parsing) are skipped to avoid false  positives, and the error names the offending path and explains the parent-wins behaviour.

### DIFF
--- a/agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
+++ b/agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
@@ -540,7 +540,7 @@ nri-redis:
                 copy_from_file: Some(TemplateableValue::from_template(
                     source.to_string_lossy().to_string(),
                 )),
-                persistent: TemplateableValue::default(),
+                persistent: false,
             },
         )]));
 
@@ -583,7 +583,7 @@ nri-redis:
                     copy_from_file: Some(TemplateableValue::from_template(
                         source.to_string_lossy().to_string(),
                     )),
-                    persistent: TemplateableValue::default(),
+                    persistent: false,
                 },
             )]));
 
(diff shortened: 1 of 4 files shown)


DavSanchez (approved):

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -290,6 +304,85 @@ fn check_basedir_escape_safety(path: &Path) -> Result<(), String> {
...
+        }
+        // `dir_content_from_map` is always ephemeral and cannot declare persistent children, so it
+        // can neither be a persistent entry nor a parent of one.
+        FilesystemEntry::DirContentFromMap { .. } => {}
+    }
+}
+
+/// Whether the flag is literally `persistent: true`.
+fn is_declared_persistent(persistent: &TemplateableValue<bool>) -> bool {
+    persistent.template == "true"
```
DavSanchez: Sorry approved by mistake earlier. How do we behave if this is templated and resolves to an invalid configuration? I'd say we need the variables for the validation to truly happen no?

Meaning this might be a render time check instead of a parse time check unfortunately :(

DavSanchez (dismissed):

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
alvarocabanas: I am going to make persistent a plain boolean, I did it as a templeatable value but it doesn't make any sense and complicates the solution.

alvarocabanas: Modified in last commit to non templateable

danielorihuela (approved):

PR #2663, opened 2026-07-03
feat: add copy_from_file support for per-agent filesystem [NR-579370]

sigilioso:
## Summary

Adds a `copy_from_file` content source to `kind: file` entries in an agent type's `filesystem:`
block. Until now a file entry could only carry inline (`text`) content; with `copy_from_file` a
file entry's bytes are copied byte-for-byte from an existing on-disk path (typically a binary
extracted from a downloaded OCI package).

## Scope

**This PR extends the per-agent `filesystem:` capability only.** A `copy_from_file` entry is
written into the sub-agent's own dedicated directory (`${nr-sub:filesystem_agent_dir}`), exactly
like every other filesystem entry today. **It does NOT add the shared filesystem yet** (it will be introduced in follow-up PRs).

## Example

```yaml
# In an agent type's on-host deployment `filesystem:` block.
deployment:
  on_host:
    packages:
      nri-redis:
        # ... OCI download coordinates; extracted under ${nr-sub:packages.nri-redis.dir}
    filesystem:
      # Inline content (unchanged, existing behaviour).
      newrelic-infra.yaml:
        kind: file
        text: ${nr-var:config_agent}

      # NEW: copy a binary from the downloaded package into the sub-agent's own dir.
      # `copy_from_file` is the source path; the entry key is the destination filename.
      # Note: integrations will actually use the 'to be introduced' shared-filesystem instead.
      integrations:
        kind: dir
        entries:
          nri-redis:
            kind: file
            copy_from_file: ${nr-sub:packages.nri-redis.dir}/nri-redis
```

## Security

After https://github.com/newrelic/newrelic-agent-control/pull/2663#discussion_r3521232877, the PR was updated to **restrict* the files that can be copied to `${nr-sub:remote_dir}` sub-folders. Any origin outside that path is rejected as invalid configuration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ Remember that the keywords that you can use are the following:
  - bugfix      => Patch
 
 ## Unreleased
+
+### enhacement
 - Add support for `copy_from_file` for on-host "in-agent" filesystem.
 
 ## v1.18.0 - 2026-07-06
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,11 @@ Remember that the keywords that you can use are the following:
 - On-host filesystem entries now accept a `persistent` flag (default `false`): ephemeral entries are deleted on sub-agent stop, persistent entries survive until the agent is removed from the fleet. Reconciliation across writes is driven by a `.ac-managed-paths.json` manifest (reserved filename — agent types must not declare it) so paths Agent Control no longer owns are deleted while sub-agent-created files are preserved.
 - Include new 'shared-filesystem-dir` variable for Agent Types.
 - Extend internal `fs` crate with a copy operation.
+<<<<<<< HEAD
 - Removing an agent from the fleet now also deletes its installed packages from disk.
+=======
+- Add support for `copy_file_from` for on-host "in-agent filesystem"
+>>>>>>> 9f6892161 (chore: update changelog)
 
 ### 🐞 Bug fixes
 - On-host self-update: an empty `version: ""` pushed from Fleet Control now behaves the same as an absent `version` field (no-op, no update attempted). Previously it silently triggered a pull of the `:latest` OCI tag.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Remember that the keywords that you can use are the following:
  - bugfix      => Patch
 
 ## Unreleased
+- Add support for `copy_from_file` for on-host "in-agent" filesystem.
 
 ## v1.18.0 - 2026-07-06
 
@@ -28,7 +29,6 @@ Remember that the keywords that you can use are the following:
 - On-host filesystem entries now accept a `persistent` flag (default `false`): ephemeral entries are deleted on sub-agent stop, persistent entries survive until the agent is removed from the fleet. Reconciliation across writes is driven by a `.ac-managed-paths.json` manifest (reserved filename — agent types must not declare it) so paths Agent Control no longer owns are deleted while sub-agent-created files are preserved.
 - Include new 'shared-filesystem-dir` variable for Agent Types.
 - Extend internal `fs` crate with a copy operation.
-- Add support for `copy_from_file` for on-host "in-agent" filesystem.
 
 ### 🐞 Bug fixes
 - On-host self-update: an empty `version: ""` pushed from Fleet Control now behaves the same as an absent `version` field (no-op, no update attempted). Previously it silently triggered a pull of the `:latest` OCI tag.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,11 +28,7 @@ Remember that the keywords that you can use are the following:
 - On-host filesystem entries now accept a `persistent` flag (default `false`): ephemeral entries are deleted on sub-agent stop, persistent entries survive until the agent is removed from the fleet. Reconciliation across writes is driven by a `.ac-managed-paths.json` manifest (reserved filename — agent types must not declare it) so paths Agent Control no longer owns are deleted while sub-agent-created files are preserved.
 - Include new 'shared-filesystem-dir` variable for Agent Types.
 - Extend internal `fs` crate with a copy operation.
-<<<<<<< HEAD
-- Removing an agent from the fleet now also deletes its installed packages from disk.
-=======
-- Add support for `copy_file_from` for on-host "in-agent filesystem"
->>>>>>> 9f6892161 (chore: update changelog)
+- Add support for `copy_from_file` for on-host "in-agent" filesystem.
 
 ### 🐞 Bug fixes
 - On-host self-update: an empty `version: ""` pushed from Fleet Control now behaves the same as an absent `version` field (no-op, no update attempted). Previously it silently triggered a pull of the `:latest` OCI tag.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ Remember that the keywords that you can use are the following:
 
 ## Unreleased
 
-### enhacement
+### enhancement
 - Add support for `copy_from_file` for on-host "in-agent" filesystem.
 
 ## v1.18.0 - 2026-07-06
--- a/agent-control/src/agent_type/error.rs
+++ b/agent-control/src/agent_type/error.rs
@@ -28,7 +28,7 @@ pub enum AgentTypeError {
     /// A templated string could not be parsed into the target value type.
     #[error("template value not parseable from the string {0}")]
     ValueNotParseableFromString(String),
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ Remember that the keywords that you can use are the following:
 - On-host filesystem entries now accept a `persistent` flag (default `false`): ephemeral entries are deleted on sub-agent stop, persistent entries survive until the agent is removed from the fleet. Reconciliation across writes is driven by a `.ac-managed-paths.json` manifest (reserved filename — agent types must not declare it) so paths Agent Control no longer owns are deleted while sub-agent-created files are preserved.
 - Include new 'shared-filesystem-dir` variable for Agent Types.
 - Extend internal `fs` crate with a copy operation.
+- Add support for `copy_file_from` for on-host "in-agent filesystem"
```
alvarocabanas: Shouldn't it be copy_from_file ?

alvarocabanas (commented):
LGTM, just left a couple of documentation issues, one is the comment, and the other is that INTEGRATING_AGENTS.md should be updated with the new type.

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -25,13 +28,22 @@ pub struct FileSystem {
     pub(super) entries: HashMap<PathBuf, RenderedEntry>,
 }
 
+/// The source of a rendered file's bytes.
+#[derive(Debug, Clone, PartialEq)]
+pub enum FileContent {
+    /// Literal (rendered) content written verbatim.
+    Inline(String),
```
DavSanchez: NIt that maybe we can call this variant `Text` to match the non-rendered one.


```suggestion
    Text(String),
```

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -52,10 +52,16 @@ pub struct FileSystem(HashMap<SafePath, FilesystemEntry>);
...
     File {
-        /// The file's (possibly templated) content.
-        text: TemplateableValue<String>,
+        /// The file's (possibly templated) content. Mutually exclusive with `copy_from_file`.
+        #[serde(default)]
+        text: Option<TemplateableValue<String>>,
+        /// A (possibly templated) source path whose bytes are copied into this entry. Mutually
+        /// exclusive with `text`.
+        #[serde(default)]
+        copy_from_file: Option<TemplateableValue<String>>,
```
DavSanchez, line 65: Should we ensure `SafePath` here? I mean the template could resolve to whatever (relative paths, even different filesystems etc) in the presence of variables.

I see `SafePath` its used on the tests but don't see it used on the templating result.

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -139,10 +145,34 @@ impl Templateable for FilesystemEntry {
...
+                        return Err(AgentTypeError::InvalidFileEntry(
+                            "cannot declare both `text` and `copy_from_file`".to_string(),
+                        ));
+                    }
+                    (None, None) => {
+                        return Err(AgentTypeError::InvalidFileEntry(
+                            "must declare either `text` or `copy_from_file`".to_string(),
+                        ));
+                    }
+                };
```
DavSanchez, line 182: I'm the sum types guy and I recommend a sum type here to ensure this just cannot happen by construction, so there are no error situations to handle after a successful input parse:

```rust
enum FileContent {
    Text(TemplateableValue<String>),
    CopyFromFile(TemplateableValue<String>)
}

impl Templateable for TemplateableValue<FileContent> { /* ... */ }

enum FileSystemEntry {
    File {
        content: TemplateableValue<FileContent>,
        persistent: TemplateableValue<bool>
    },
    // ...
}
```

and then this reduces to:

```suggestion
                // From earlier `match` on `FilesystemEntry::File`...
                let content = match file {
                    Text(text) => {
                        rendered::FileContent::Inline(text.template_with(variables)?)
                    }
                    CopyFromFile(copy_from_file) => {
                        rendered::FileContent::Copy(copy_from_file.template_with(variables)?.into())
                    }
                };
```

Maybe I'm missing something that makes this impossible, haven't checked out the code itself to try, I'm reviewing here directly.

agent-control/src/agent_type/error.rs
```diff
@@ -28,6 +28,9 @@ pub enum AgentTypeError {
     /// A templated string could not be parsed into the target value type.
     #[error("template value not parseable from the string {0}")]
     ValueNotParseableFromString(String),
+    /// A `kind: file` entry did not declare exactly one of `text` or `copy_from_file`.
+    #[error("invalid filesystem file entry: {0}")]
+    InvalidFileEntry(String),
```
DavSanchez: This goes away with the enum I propose above.

DavSanchez (commented):
I have a type of comments!

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
sigilioso: I actually considered it but opted for the `Option<...>` representation because it made the error more readable without custom deserialization (we have the explicit error message instead of 'invalid variant ...')

CHANGELOG.md
sigilioso: Good catch! Thanks!

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
sigilioso: We usually limit the destination file (this would be the origin), we can also limit the paths were a file can be copied from, although `SafePath` wouldn't fit because the path can include variables and is not known until rendered.

agent-control/src/agent_type/error.rs
sigilioso: Let's keep the discussion in the thread above 🙂

CHANGELOG.md
sigilioso: Updated here [9a872bc](https://github.com/newrelic/newrelic-agent-control/pull/2663/commits/9a872bcc7d02bed69408731ff0cbd62c4154764d)

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
sigilioso: Thanks, changed here: [95b7c9f](https://github.com/newrelic/newrelic-agent-control/pull/2663/commits/95b7c9f59689cd45fffda0ff4a29791816c3fc3f)

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
sigilioso: I've restricted the origin here: [1f06b28](https://github.com/newrelic/newrelic-agent-control/pull/2663/commits/1f06b28001f1f67b3b6bbe17d102cfd9d66ca124), and updated the PR description.

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -52,10 +53,16 @@ pub struct FileSystem(HashMap<SafePath, FilesystemEntry>);
...
     File {
-        /// The file's (possibly templated) content.
-        text: TemplateableValue<String>,
+        /// The file's (possibly templated) content. Mutually exclusive with `copy_from_file`.
+        #[serde(default)]
+        text: Option<TemplateableValue<String>>,
+        /// A (possibly templated) source path whose bytes are copied into this entry. Mutually
+        /// exclusive with `text`.
+        #[serde(default)]
+        copy_from_file: Option<TemplateableValue<String>>,
```
gsanchezgavier, line 65: would make sense, to enforce this exclusiveness through a enum ?

gsanchezgavier (approved):
LGTM ! just left a nit

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
sigilioso: I think @DavSanchez actually pointed out the same https://github.com/newrelic/newrelic-agent-control/pull/2663#discussion_r3521258703.

I opted for the two `Option<...>` to have an easy way (without custom deserialization) to have a useful error message: we have `invalid filesystem file entry: cannot declare both 'text' and 'copy_from_file'` (or the corresponding message if there is none of them) instead of `invalid variant ...` when deserializing the enum.

I can change it to an enum but I think it makes the implementation a bit more complex in this case. Note that this is only the yaml-representation, the _rendered_ one does have the enum.

PR #2664, opened 2026-07-03
feat: remove filesystem prune (manifest reconcile) mechanism

gsanchezgavier:
## Summary
Drops the `.ac-managed-paths.json` manifest and its reconcile logic. `write` now only materializes the currently-declared tree — it no longer deletes paths a previous config declared but the current one dropped. This removes an agent-writable manifest that drove deletions, simplifying the on-host filesystem lifecycle to just ephemeral-wipe + write.

## Changes
- Delete the `managed_paths` module (manifest load/save, `prune_stale`, `is_within_base`) and its tests.
- Drop `base_dir` from `rendered::FileSystem`; move `delete_path` into `rendered.rs`.
- Remove the reserved-filename (`check_not_reserved`) validation and its test.
- Update `INTEGRATING_AGENTS.md`, `AC_repositories_and_files.md`, and supervisor comment to reflect no-prune behavior.

## Notes for reviewers
- Behavioral change: a **persistent** entry dropped from an agent type now lingers on disk until the whole agent dir is reclaimed (ResourceCleaner on fleet removal). Ephemeral entries (incl. `dir_content_from_map`) are still cleaned via `delete_ephemeral` + re-render.


docs/INTEGRATING_AGENTS.md
```diff
@@ -538,61 +538,11 @@ agent/integrations.d/
...
-- **The manifest** drives reconciliation on every write event. Anything Agent Control wrote on the previous successful write is recorded in the manifest. On the next write, Agent Control diffs the manifest against the new declared set: paths it owned previously and no longer owns are deleted; paths it never owned are left alone.
-
-The flag does **not** shield the entry from intentional removal: if you delete an entry from the agent type (or remove a key from a `dir_content_from_map` source map), the manifest diff catches it and the on-disk path is deleted on the next write event.
+Every `file` and `dir` entry accepts a boolean `persistent:` (default `false`). The `persistent` flag controls whether the entry's on-disk path is wiped when the tree is cleaned: on sub-agent stop, and just before every (re)write of the tree (start, restart, and config apply). Ephemeral entries are wiped at those points; persistent entries are kept. Wiping before each write means leftover ephemeral content never carries across — even after an ungraceful shutdown (crash/SIGKILL) that skipped the stop-time cleanup.
```
alvarocabanas, line 542: Nit: May we add also an entry explaining the edge case, something like:

**An entry removed from the agent type while the sub-agent is stopped is not reclaimed.** Cleanup only ever walks the *currently declared* tree, so a path that is no longer declared will be kept in disc if its parent is persistent, even if it used to be ephemeral.

alvarocabanas (dismissed):
LGTM, it removes the manifest part and the rest is left as it was.

I left a NIT about adding a comment for the edge case, but I don't know if we want to tackle it in a future PR.

docs/INTEGRATING_AGENTS.md
gsanchezgavier: yes makes sense, I'll add the comment. I think there are several options to cover a usecase like that currently , and one of this is to keep the old defined file and marked as non persistance, it would be a workaround for the cases where a file needs to be removed

gsanchezgavier: done!

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -342,6 +316,15 @@ fn check_basedir_escape_safety(path: &Path) -> Result<(), String> {
     })
 }
 
+/// Returns `true` only when `path` is genuinely contained in `base_dir`: an absolute path, with no
+/// `..` traversal, that lies under `base_dir`. Used to vet manifest entries before acting on them,
+/// since the manifest is read from an agent-writable location and must not be trusted to point
+/// outside the agent's own directory.
```
sigilioso: > [...] Used to vet manifest entries [...]

Nit: the comment is outdated, isn't it?

sigilioso (approved):
LGTM

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -249,6 +192,15 @@ fn copy_file(
         .copy(source, path)
         .map_err(|err| FileSystemEntriesError(format!("copying {source:?} to {path:?}: {err}")))
 }
+/// Recursively deletes the file or directory at `path`.
+fn delete_path(path: &Path) -> io::Result<()> {
+    trace!("Deleting path {}", path.display());
+    if path.is_dir() {
+        std::fs::remove_dir_all(path)
```
danielorihuela, line 199: Should we use here our abstractions for directories and files? `LocalFile` and `DirectoryManagerFs`?

danielorihuela:
If I understood correctly we now:
* remove all ephemeral files on stop and start
* overwrite persistent files

and the following is out of scope
* modifications while Agent Control is stopped
* removing "persistent" files that no longer appear on the agent type

So, the only issues are:
* files might linger there indifinetely
* if an agent stores files on an ephemeral folder, they are removed even if they are not strictly under AC control (it's fine for me)

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
gsanchezgavier: yes i commented the same in @alvarocabanas  PR , i think he is tracking this

sigilioso: @alvarocabanas Did we finally tackled it? I thought we were addressing it in #2662 but I also see it merged 🤔

PR #2666, opened 2026-07-06
feat: shared filesystem support [NR-579370] [NR-579374]

sigilioso:
## Summary

Adds a `shared_filesystem:` block to on-host agent types. Its entries are materialized under the
base shared across sub-agents (`${nr-sub:shared_filesystem_dir}` →
`/var/lib/newrelic-agent-control/shared-filesystem`) instead of the per-agent dir. This is how an
OHI agent type makes its config and binaries available to the infra-agent, which reads them from a
shared location.

## Out of scope

- **Cleanup on uninstall** Removing shared entries on uninstall/reconfigure is a separate, deferred task. As a result, the `persistent` field for the shared filesystem is ignored.
- **Handle concurrent writes**: two agents writing to the same file at the same time could cause race-conditions. This should be tackled when implementing the clean-up mechanism.
- Extended integration tests (with cleanup) and e2e.

## Changes

- Added shared-filesystem support
- Added integration tests for Agent Types using shared-filesystem
- Extended the CustomAgentType tooling to support multiple agent-types and the shared-filesystem field
- Updated Agent Type docs (including a callout for the missing cleanup support)

## YAML example

```yaml
deployment:
  on_host:
    packages:
      nri-redis:
        # ... OCI download coordinates; extracted under ${nr-sub:packages.nri-redis.dir}
    shared_filesystem:
      # Config the infra-agent reads, placed in the shared location.
      infra-agent-ohi-configs:
        kind: dir
        entries:
          nri-redis.yaml:
            kind: file
            text: ${nr-var:config_integration}
      # The OHI binary, copied into a sibling shared folder.
      infra-agent-ohi-binaries:
        kind: dir
        entries:
          nri-redis:
            kind: file
            copy_from_file: ${nr-sub:packages.nri-redis.dir}/nri-redis
```

### DIFF
--- a/agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
+++ b/agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
@@ -127,24 +127,55 @@ impl From<SafePath> for PathBuf {
 #[derive(Debug, Default, PartialEq, Clone)]
 pub struct DirEntriesMap(HashMap<SafePath, String>);
 
+impl FileSystem {
+    /// Whether no entries are declared.
+    fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    /// Templates each entry and roots it under `base_dir`, prepending `base_dir` to each relative
+    /// top-level key (the only place a final on-disk path is constructed).
+    fn render_entries(
+        self,
+        base_dir: &Path,
+        variables: &Variables,
+    ) -> Result<HashMap<PathBuf, rendered::RenderedEntry>, AgentTypeError> {
+        self.0
+            .into_iter()
+            .map(|(key, entry)| {
+                let path = base_dir.join(&key);
+                Ok((path, entry.template_with(variables)?))
+            })
+            .collect()
+    }
+}
+
 impl Templateable for FileSystem {
     type Output = rendered::FileSystem;
 
     fn template_with(self, variables: &Variables) -> Result<Self::Output, AgentTypeError> {
         let base_dir = PathBuf::from(filesystem_agent_dir(variables)?);
-
-        let entries = self
-            .0
-            .into_iter()
-            .map(|(key, entry)| {
-                // The only place we construct a final-on-disk path: prepend the sub-agent's
-                // dedicated filesystem dir to the user-provided relative top-level key.
-                let path = base_dir.join(&key);
-                Ok((path, entry.template_with(variables)?))
-            })
-            .collect::<Result<HashMap<_, _>, AgentTypeError>>()?;
-
+        let entries = self.render_entries(&base_dir, variables)?;
         Ok(rendered::FileSystem::new(entries))
+    }
+}
+
+/// Files and directories shared across sub-agents, rooted at `${nr-sub:shared_filesystem_dir}`.
+#[derive(Debug, Default, Deserialize, Clone, PartialEq)]
+pub struct SharedFileSystem(FileSystem);
+
+impl Templateable for SharedFileSystem {
+    type Output = rendered::SharedFileSystem;
+
+    fn template_with(self, variables: &Variables) -> Result<Self::Output, AgentTypeError> {
+        // Skip resolving the shared base when nothing is declared: an unused shared filesystem
+        // must not require `${nr-sub:shared_filesystem_dir}` to be present.
+        if self.0.is_empty() {
+            return Ok(rendered::SharedFileSystem::new(HashMap::new()));
+        }
+        let base_dir = PathBuf::from(shared_filesystem_dir(variables)?);
+        let entries = self.0.render_entries(&base_dir, variables)?;
+        Ok(rendered::SharedFileSystem::new(entries))
     }
 }
 
@@ -223,6 +254,15 @@ impl Templateable for FilesystemEntry {
 
 fn filesystem_agent_dir(variables: &Variables) -> Result<String, AgentTypeError> {
     let key = Namespace::SubAgent.namespaced_name(AgentAttributes::VARIABLE_FILESYSTEM_AGENT_DIR);
+    match variables.get(&key).and_then(Variable::get_final_value) {
+        Some(TrivialValue::String(s)) => Ok(s.clone()),
+        _ => Err(AgentTypeError::MissingValue(key)),
+    }
+}
+
+/// Resolves `${nr-sub:shared_filesystem_dir}`, the base for the shared filesystem tree.
+fn shared_filesystem_dir(variables: &Variables) -> Result<String, AgentTypeError> {
+    let key = Namespace::SubAgent.namespaced_name(AgentAttributes::VARIABLE_SHARED_FILESYSTEM_DIR);
     match variables.get(&key).and_then(Variable::get_final_value) {
         Some(TrivialValue::String(s)) => Ok(s.clone()),
         _ => Err(AgentTypeError::MissingValue(key)),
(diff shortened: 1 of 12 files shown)


docs/AC_repositories_and_files.md
```diff
@@ -79,13 +81,18 @@ The following shows the directory structure used by Agent Control, assuming an e
...
-    │       └── filesystem
-    │            └── nr-infra
-    │                ├── .ac-managed-paths.json
-    │                ├── integrations.d
-    │                │   └── nri-redis.yaml
-    │                └── config
-    │                      └── newrelic-infra.yml
+    │       ├── filesystem
+    │       │    └── nr-infra
+    │       │        ├── .ac-managed-paths.json
```
danielorihuela: Remove managed path

docs/AC_repositories_and_files.md
```diff
@@ -79,13 +81,18 @@ The following shows the directory structure used by Agent Control, assuming an e
...
+    │       │        ├── integrations.d
+    │       │        │   └── nri-redis.yaml
+    │       │        └── config
+    │       │              └── newrelic-infra.yml
```
danielorihuela: These shouldn't be there from that point on, but maybe it's better to document that on the infra-agent type change task

agent-control/src/agent_type/runtime_config/on_host/rendered.rs
```diff
@@ -18,6 +21,8 @@ pub struct OnHost {
     pub health: Option<OnHostHealthConfig>,
     /// Files and directories to materialize on disk.
     pub filesystem: FileSystem,
+    /// Files and directories to materialize in the base shared across sub-agents (write-only).
```
danielorihuela: `write-only`? Can't we read the files?

sigilioso: I'll update the comment it was trying to highlight that it doesn't delete anything but it is actually misleading.

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -1446,4 +1486,110 @@ new.txt:
...
+            let yaml = format!("{name}:\n  kind: file\n  text: {content}");
+            serde_saphyr::from_str::<SharedFileSystem>(&yaml)
+                .unwrap()
+                .template_with(&variables)
+                .unwrap()
+                .write(&LocalFile, &DirectoryManagerFs)
+                .unwrap();
+        }
+
+        assert!(shared.join("agent-a.yaml").exists(),);
```
danielorihuela: nit: remove comma

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -1446,4 +1486,110 @@ new.txt:
...
+            .unwrap();
+
+        assert_eq!(
+            std::fs::read_to_string(shared.join("infra-agent-ohi-configs/nri-redis.yaml")).unwrap(),
+            "integration: redis"
+        );
+    }
+
+    #[test]
+    fn shared_filesystem_write_does_not_prune_other_entries() {
```
danielorihuela: I don't get this test. The name says "... does not prune other entries", but we do a single write. There's no delete at any point. Am I right?

sigilioso: The intention was to assure that one agent's filesystem is not removing other agent's ones. But considering the implementation, it is probably misleading. I'll review and update or remove the tests. Thanks!

docs/AC_repositories_and_files.md
sigilioso: I'm actually rolling back these changes and better document on the infra-agent Agent Type change as suggested here https://github.com/newrelic/newrelic-agent-control/pull/2666#discussion_r3535646567

docs/AC_repositories_and_files.md
sigilioso: Rolled it back here: [2291ba8](https://github.com/newrelic/newrelic-agent-control/pull/2666/commits/2291ba8e03a16f39785a2e6f9feaeb6c01ee4095)

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
sigilioso: Done! [0d25573](https://github.com/newrelic/newrelic-agent-control/pull/2666/commits/0d25573a53d03c9940e6cda637a1aece75443140)

agent-control/src/agent_type/runtime_config/on_host/rendered.rs
sigilioso: Comment updated: [0d25573](https://github.com/newrelic/newrelic-agent-control/pull/2666/commits/0d25573a53d03c9940e6cda637a1aece75443140)

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
sigilioso: I think the test made sense but it needed better naming: [0d25573](https://github.com/newrelic/newrelic-agent-control/pull/2666/commits/0d25573a53d03c9940e6cda637a1aece75443140)

danielorihuela:
@claude detected a potential vulnerability. We can copy files from outside the remote folder.

Basically, it says that we can create a "source" file which is a symlink. That means that we can copy any file of the system into the remote folder. We should canonicalize the path to avoid that.

It's not a big deal, at the end, the post-install script allows that and many more nasty things. But I don't know, if we say that we can only copy from and to the remote folder, we should make sure the behaviour is like that.

Here is the test that I got from claude (and I modified).

```rust
    /// Demonstrates that `copy_from_file` confinement can be bypassed via a symlink planted inside
    /// `remote_dir` that points to a file outside it. `is_within_base` passes because the check is
    /// lexical (the symlink path itself is within the base), but `std::fs::copy` follows the link
    /// and reads the target — which is outside the allowed base.
    ///
    /// This test is intentionally written to **pass** (i.e. the copy succeeds and reads the secret)
    /// to document the existing vulnerability. Once the fix is applied (canonicalize the source
    /// before the confinement check, or open with O_NOFOLLOW), this test should be updated to
    /// assert an `InvalidFileEntry` error instead.
    #[cfg(unix)]
    #[test]
    fn copy_from_file_symlink_bypass_demonstrates_vulnerability() {
        use std::os::unix::fs::symlink;

        let tmp_dir = TempDir::new().unwrap();

        // Secret file lives *outside* remote_dir — confinement should prevent reading it.
        let secret_file = tmp_dir.path().join("outside-secret.txt");
        std::fs::write(&secret_file, "secret-content").unwrap();

        let remote = tmp_dir.path().join("remote");
        let agent_dir = remote.join("filesystem").join("agent");
        std::fs::create_dir_all(&agent_dir).unwrap();

        // Plant a symlink *inside* remote_dir pointing at the secret file.
        // The symlink path is lexically within `remote_dir`, so `is_within_base` passes.
        let symlink_inside_remote = remote.join("innocent-looking-file.txt");
        symlink(&secret_file, &symlink_inside_remote).unwrap();

        let fs_input = FileSystem(HashMap::from([(
            PathBuf::from("output").try_into().unwrap(),
            FilesystemEntry::File {
                text: None,
                copy_from_file: Some(TemplateableValue::from_template(
                    symlink_inside_remote.to_string_lossy().to_string(),
                )),
                persistent: false,
            },
        )]));

        // `template_with` does NOT reject the symlink path — it is lexically within remote_dir.
        // This is the bug: the check should canonicalize first.
        let rendered = fs_input
            .template_with(&fs_variables(&agent_dir, &remote))
            .expect("confinement check passes because symlink path is lexically within base");

        // Writing materializes the copy — std::fs::copy follows the symlink and reads the secret.
        rendered.write(&LocalFile, &DirectoryManagerFs).unwrap();

        let output_path = agent_dir.join("output");
        let copied_content = std::fs::read_to_string(&output_path).unwrap();

        // The secret content is now accessible outside its intended boundary.
        // FIX NEEDED: this assertion should instead be an `InvalidFileEntry` error from
        // `template_with`, rejected before any disk I/O occurs.
        assert_eq!(
            copied_content, "secret-content",
            "vulnerability confirmed: symlink bypass copied content from outside remote_dir"
        );
    }
```

danielorihuela (dismissed):
Left a doubt regarding the behaviour of the copy.
It can be solved on a different PR.

Other than that, great work!!!

sigilioso:
🤔 To do that we would need not only to target the file in the agent-type but also create the symlink in the remote-dir folder...

In order to do that remotely you would need an Agent Type whose install script creates the symlink (and, as you pointed out, you can do anything in the install-script!).

That said, since we added limits to the `copy_from_file`, I guess it also makes sense to canonicalize as well.

However, I'd rather take care of that in a different PR (after all the `copy_from_file` feature) was already in main and was introduced in #2663

alvarocabanas (dismissed):
LGTM, I just left a super small NIT

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -127,27 +127,58 @@ impl From<SafePath> for PathBuf {
...
+/// Files and directories shared across sub-agents, rooted at `${nr-sub:shared_filesystem_dir}`.
+#[derive(Debug, Default, Deserialize, Clone, PartialEq)]
+pub struct SharedFileSystem(FileSystem);
+
+impl Templateable for SharedFileSystem {
+    type Output = rendered::SharedFileSystem;
+
+    fn template_with(self, variables: &Variables) -> Result<Self::Output, AgentTypeError> {
+        // Skip resolving the shared base when nothing is declared: an unused shared filesystem
+        // must not require `${nr-sub:shared_filesystem_dir}` to be present.
```
alvarocabanas: Nit: the comment was hard to understand, what about something like:

// Return early if no shared entries are declared. This way agent types that don't use the
// shared filesystem aren't forced to provide the `${nr-sub:shared_filesystem_dir}` variable.

sigilioso: Updated here: [5fd183e](https://github.com/newrelic/newrelic-agent-control/pull/2666/commits/5fd183ea1a85187794034856128655883b35ca36)

danielorihuela (approved):

PR #2668, opened 2026-07-06
fix: preserve local config on rpm reinstall

danielorihuela:
Mark local_config.yaml as config|noreplace in the RPM package to prevent it from being overwritten on reinstall or renamed to .rpmsave on uninstall.

## Context
On RedHat and SUSE systems, uninstalling the package renamed the user's `local_config.yaml` to `local_config.yaml.rpmsave` instead of leaving it in place. On reinstall, `local_config.yaml` is created with the package template, discarding user configuration (license key, fleet ID, etc.).

This did not affect Debian/Ubuntu because dpkg uses noreplace semantics by default.

## Technical details
nfpm's `type: config` maps to RPM's `%config`, which overwrites modified files on upgrade and renames them to `.rpmsave` on uninstall. Adding `|noreplace` maps to `%config(noreplace)`, which leaves the existing file untouched on upgrade (saving the new template as `.rpmnew`) and leaves it in place on uninstall.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -157,7 +157,7 @@ nfpms: # deb and rpm are only built for linux targets
           mode: 0700
       - src: build/package/config.yaml
         dst: /etc/newrelic-agent-control/local-data/agent-control/local_config.yaml
-        type: config
+        type: config|noreplace
         file_info:
           mode: 0600
       - src: build/package/newrelic-agent-control.service
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ Remember that the keywords that you can use are the following:
 
 ## Unreleased
 
+### bugfix
+- RPM packages now restore `local_config.yaml` from the `.rpmsave` backup left by a prior uninstall.
+
 ### enhancement
 - Add support for `copy_from_file` for on-host "in-agent" filesystem.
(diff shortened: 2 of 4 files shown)


danielorihuela:
I tested a reinstall (install with old package, uninstall, and reinstall with new package) and it's working.
The only problem is that `rpmsave` is not automatically restored. We have to restore it for the install to pick it up again. At least it is not overwritten anymore.

```bash
[ec2-user@ip-10-10-35-155 dist]$ sudo ls /etc/newrelic-agent-control/local-data/agent-control
local_config.yaml  local_config.yaml.rpmnew  local_config.yaml.rpmsave
[ec2-user@ip-10-10-35-155 dist]$ sudo cat /etc/newrelic-agent-control/local-data/agent-control/local_config.yaml
fleet_control:
  endpoint: xxx
  signature_validation:
    public_key_server_url: xxx
  fleet_id: xxx
  auth_config:
    token_url: xxx
    client_id: xxx
    provider: xxx
    private_key_path: xxx
server:
  enabled: true
agents: {}
[ec2-user@ip-10-10-35-155 dist]$ sudo cat /etc/newrelic-agent-control/local-data/agent-control/local_config.yaml.rpmnew
# fleet_control:
#   endpoint: https://opamp.service.newrelic.com/v1/opamp
#   headers:
#     api-key: API_KEY_HERE
#   auth_config:
#     token_url: PLACEHOLDER
#     client_id: PLACEHOLDER
#     provider: PLACEHOLDER
#     private_key_path: PLACEHOLDER

agents: {}
```

Now the template goes into `rpmnew` and the original is untouched.

danielorihuela:
I also tested an install, uninstall, reinstall (with the new package) and now the `rpmsave` file is used on reinstall.

MunhozThiago (approved):
LGTM

paologallinaharbur (approved):

PR #2670, opened 2026-07-07
fix: add package delete after post download hook fails

vjripoll:
When an on-host agent's `post_download_hook` fails, Agent Control now **removes the extracted OCI package** before returning the error.

The package is extracted to its final directory *before* the post-download hook runs. On a hook failure we only cleaned up the temporary download dir, leaving the extracted package on disk. Because `install()` treats an already-present package directory as "already installed" and skips download/extraction/hook, a later Agent Control restart would **skip the hook and start a half-installed agent** whose setup never completed.

Deleting the package on hook failure means the next start re-attempts a clean install (download → extract → hook) instead of running a broken agent.

In `install_archive`, if `PostDownloadHookExecutor::execute` fails, delete the extracted `install_path` before propagating the error (mirrors the existing cleanup done for the temp dir and for extraction failures).

### DIFF
--- a/agent-control/src/package/oci/package_manager.rs
+++ b/agent-control/src/package/oci/package_manager.rs
@@ -121,7 +121,7 @@ where
     /// The package is first downloaded to `temp_package_path` and then extracted to `package_path`.
     fn install_archive(
         &self,
-        package_data: PackageData,
+        package_data: &PackageData,
         tmp_download_path: &Path,
         install_path: &Path,
     ) -> Result<InstalledPackageData, OCIPackageManagerError> {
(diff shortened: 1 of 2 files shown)


sigilioso:
As discussed _offline_, we could actually have the same issue when rolling back to a version that was successfully installed and whose files are still on the disk. We could cover both cases by always executing the script (regardless of the files being present or not).

paologallinaharbur:
> We could cover both cases by always executing the script (regardless of the files being present or not).

This might be not super easy, we might need to have some sort of "state" or require the script to be idempotent and clarify it in the docs (this would be the best)

PR #2671, opened 2026-07-08
feat: Multipackage agent version

alvarocabanas:
On-host agent-type parsing now validates `version_package`: it must reference a declared package, and is required when more than one package is defined. Invalid configurations are rejected at parse time with a descriptive error. 
This `version_package` will be used to know which version to report as `agent.version`.

### DIFF
--- a/agent-control/src/agent_type/runtime_config/on_host.rs
+++ b/agent-control/src/agent_type/runtime_config/on_host.rs
@@ -20,12 +20,27 @@ pub mod rendered;
 ///
 /// It contains the instructions of what are the agent binaries, command-line arguments, the environment variables passed to it and the restart policy of the supervisor.
 #[derive(Debug, Deserialize, Default, Clone, PartialEq)]
+#[serde(try_from = "OnHostRaw")]
 pub struct OnHost {
+    executables: Vec<Executable>,
+    enable_file_logging: TemplateableValue<bool>,
+    /// Enables and defines health checks configuration.
+    health: Option<OnHostHealthConfig>,
+    filesystem: FileSystem,
+    packages: Packages,
+    shared_filesystem: SharedFileSystem,
+    /// Package whose OCI version is reported as the `agent.version` identifying attribute.
+    version_package: Option<PackageID>,
+}
+
+type Packages = HashMap<PackageID, Package>;
+
+#[derive(Debug, Deserialize)]
+struct OnHostRaw {
     #[serde(deserialize_with = "deserialize_executables", default)]
     executables: Vec<Executable>,
     #[serde(default)]
     enable_file_logging: TemplateableValue<bool>,
-    /// Enables and define health checks configuration.
     #[serde(default)]
     health: Option<OnHostHealthConfig>,
     #[serde(default)]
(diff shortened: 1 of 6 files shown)


agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -331,6 +328,39 @@ where
         );
     }
 
+    fn version_package_to_report(&self) -> Option<(&PackageID, &RenderedPackage)> {
```
danielorihuela: At this point, data was already validated. Either we have the id of a package, or we don't because there are no packages. I understand this warning are to be defensive, but are not needed, right?

agent-control/src/agent_type/runtime_config/on_host.rs
```diff
@@ -34,9 +49,61 @@ pub struct OnHost {
...
+        0 => Ok(None),
+        1 => Ok(packages.keys().next().cloned()),
+        _ => Err(format!(
+            "`version_package` is required when more than one package is defined; declared packages: [{}]",
+            sorted_package_ids(packages)
+        )),
+    }
+}
+
+fn sorted_package_ids(packages: &Packages) -> String {
```
paologallinaharbur: why do we see do sort package ids?

agent-control/src/agent_type/runtime_config/on_host.rs
```diff
@@ -20,12 +20,27 @@ pub mod rendered;
 ///
 /// It contains the instructions of what are the agent binaries, command-line arguments, the environment variables passed to it and the restart policy of the supervisor.
 #[derive(Debug, Deserialize, Default, Clone, PartialEq)]
+#[serde(try_from = "OnHostRaw")]
```
paologallinaharbur: I do not like that much that we are duplicating the struct, I think that usually we are adding an anonymous struct inside the function
Usually we reimplement `impl<'de> Deserialize<'de> for AgentTypeMetadata {`
Adding the custom raw struct inside

agent-control/src/agent_type/runtime_config/on_host.rs
```diff
@@ -34,9 +49,61 @@ pub struct OnHost {
     shared_filesystem: SharedFileSystem,
     #[serde(default)]
     packages: Packages,
+    #[serde(default)]
+    version_package: Option<PackageID>,
```
paologallinaharbur: I do not like that much this name

version_from_package_id?
version_from_package?

paologallinaharbur: to be honest I do not like even mines 😓

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -239,11 +242,20 @@ where
...
+            version_package: None,
             filesystem,
             shared_filesystem,
         }
     }
 
+    /// Sets the package whose OCI version is reported as `agent.version`. When left unset (`None`),
+    /// [`check_subagent_version`](Self::check_subagent_version) falls back to the sole configured
+    /// package.
+    pub fn with_version_package(mut self, version_package: Option<PackageID>) -> Self {
```
paologallinaharbur: since in prod code we are always calling  "with_version_package", would make sense to just add it as an extra parameter of the new?

agent-control/src/sub_agent/on_host/supervisor.rs
paologallinaharbur: yes I agree that feels too much! I would trust the previous validation

paologallinaharbur: If it is None we could just report nothing

agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -694,6 +725,90 @@ pub mod tests {
...
+        // No version_package + multiple packages is ambiguous: nothing is reported.
+        let packages = RenderedPackages::from([
+            (
+                "infra".to_string(),
+                rendered_package("newrelic-infra", "1.2.3"),
+            ),
+            ("flex".to_string(), rendered_package("nri-flex", "4.5.6")),
+        ]);
+        let supervisor = supervisor_with_packages(packages, None);
+        assert!(supervisor.version_package_to_report().is_none());
```
paologallinaharbur: Could we have table test for this?

paologallinaharbur: Moreover I feel like there are some reduntant use cases if we remove version_package_to_report

agent-control/src/sub_agent/on_host/supervisor.rs
alvarocabanas: I know, i added it as an extra validation, it shouldn't arrive here unvalidated, but yes I can remove it if it feels too much.

agent-control/src/agent_type/runtime_config/on_host.rs
alvarocabanas: Really we don't need to, it's just a matter of treating them to be displayed, I removed the sorting, but showing the packages separated by comma is valuable in the error when parsing.

agent-control/src/sub_agent/on_host/supervisor.rs
danielorihuela: From my side, it's okay. I just wanted to make sure if this was called elsewhere or something.

danielorihuela: But if we want to remove it, I also think it's okay. It's what I would have done myself.

danielorihuela: @alvarocabanas I see you removed one validation, but what I was proposing really is to go all the way down to

```rust
self.version_package.map(|id| self.packages_config.get_key_value(id))
```

No function would be needed at all.

agent-control/src/agent_type/runtime_config/on_host.rs
alvarocabanas: Changed in last commit.

agent-control/src/agent_type/runtime_config/on_host.rs
alvarocabanas: I change it to reported_version_package, WDYT?

agent-control/src/sub_agent/on_host/supervisor.rs
alvarocabanas: I did remove it in last commit.

danielorihuela: Are we talking about the same? I still see `fn version_package_to_report`

agent-control/src/sub_agent/on_host/supervisor.rs
alvarocabanas: I put it because I thought it's an optional parameter, but it's true that we always call it, I was actually not adding it to the apply and wouldn't report the version after a reconcile.
Fixed in last commit.

agent-control/src/sub_agent/on_host/supervisor.rs
alvarocabanas: Modified in last commit.

agent-control/src/sub_agent/on_host/supervisor.rs
alvarocabanas: You are right, sorry, I changed it now.

danielorihuela (dismissed):
Okey from my side. Let's see what @paologallinaharbur says.

paologallinaharbur (approved):

PR #2672, opened 2026-07-08
fix: add trigger always post download hook

vjripoll:
The post_download_hook was only executed as part of the download+extraction path inside install_archive. When a package was already present on disk, install() return early without running the hook. This left two gaps:

- Agent Control restart: the package is persisted under stored_packages, so on restart install() saw the files already present and skipped the hook entirely the agent could start without its host-side prerequisites being (re)validated.
- Rollback to a previously-installed version: same situation the older package is still on disk, so switching back to it never re-ran the hook.

Decouple the hook from the download/extraction step and run it on every install, whether the package was just downloaded or was already present on disk. A non-zero hook exit fails the install so the agent is not started. The package folder is not deleted on failure the safety guarantee comes from install() returning Err, which prevents the supervisor from starting the executables (verified in both SupervisorOnHost::start() and apply()).

- install_archive now only downloads and extracts (takes &PackageData); the hook block was moved out of it.
- New helper run_post_download_hook(&PackageData, &Path): runs the configured hook (if any) with its working directory set to the package dir; no-op when no hook is configured.
- install() restructured so the "package already present" and "freshly downloaded" branches both produce the InstalledPackageData, and the hook runs unconditionally afterwards, before retain_packages.

### DIFF
--- a/agent-control/src/package/oci/package_manager.rs
+++ b/agent-control/src/package/oci/package_manager.rs
@@ -121,7 +121,7 @@ where
     /// The package is first downloaded to `temp_package_path` and then extracted to `package_path`.
     fn install_archive(
         &self,
-        package_data: PackageData,
+        package_data: &PackageData,
         tmp_download_path: &Path,
         install_path: &Path,
     ) -> Result<InstalledPackageData, OCIPackageManagerError> {
@@ -131,27 +131,40 @@ where
 
         let downloaded_package = self
             .downloader
-            .download(&package_data, tmp_download_path)
+            .download(package_data, tmp_download_path)
             .map_err(OCIPackageManagerError::Download)?;
 
         self.extract_package(&downloaded_package, install_path)
             .inspect_err(|e| warn!("OCI package installation failed: {}", e))?;
-
-        // Execute post-download hook if configured
-        if let Some(ref hook) = package_data.post_download_hook {
-            debug!(
-                "Executing post-download hook for package {}",
-                package_data.id
-            );
-            let executor = PostDownloadHookExecutor::new(install_path.to_path_buf());
-            executor.execute(hook)?;
-        }
 
         debug!("OCI package installed at {}", install_path.display());
         Ok(InstalledPackageData {
             id: package_data.id.clone(),
             installation_path: install_path.to_path_buf(),
         })
+    }
+
+    // The hook runs on **every** install, including when the package files were already present on
+    // disk (an Agent Control restart, or a rollback to a previously-installed version). This is
+    // intentional: the hook performs host-side setup/validation that can go stale (kernel changes,
+    // missing headers, removed directories), so skipping it when the files are present could start
+    // an agent whose prerequisites are no longer met. A non-zero exit fails the install so the
+    // agent is not started.
+    fn run_post_download_hook(
+        &self,
+        package_data: &PackageData,
+        install_path: &Path,
+    ) -> Result<(), OCIPackageManagerError> {
+        let Some(hook) = package_data.post_download_hook.as_ref() else {
+            return Ok(());
+        };
+        debug!(
+            "Executing post-download hook for package {}",
+            package_data.id
+        );
+        let executor = PostDownloadHookExecutor::new(install_path.to_path_buf());
+        executor.execute(hook)?;
+        Ok(())
     }
 
     /// Extract the downloaded package file from `download_filepath` to `extract_dir`.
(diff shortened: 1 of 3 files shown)


agent-control/src/package/oci/package_manager.rs
```diff
@@ -372,33 +385,32 @@ where
...
-            return Ok(installed_package);
-        }
-
-        let temp_package_path = get_temp_package_path(&self.remote_dir, agent_id, &package_data)?;
+            self.directory_manager
+                .delete(&temp_package_path)
+                .map_err(OCIPackageManagerError::Install)?;
 
-        // If we face an error during installation, we must ensure the temporary directory is deleted.
-        // We hide the error of the folder if something else went wrong.
```
sigilioso, line 394: Did we leave out this comment on purpose?

agent-control/src/package/oci/package_manager.rs
```diff
@@ -938,6 +950,112 @@ mod tests {
         assert_eq!(installed.id, TEST_PACKAGE_ID);
     }
 
+    #[test]
+    fn test_install_runs_hook_even_when_package_already_present() {
+        use crate::agent_type::runtime_config::on_host::executable::rendered::{Args, Env};
+        use crate::agent_type::runtime_config::on_host::package::rendered::PostDownloadHook;
+        use std::collections::HashMap;
```
sigilioso: Is there any specific reason to add the imports on each test?

agent-control/src/package/oci/package_manager.rs
```diff
@@ -938,6 +950,112 @@ mod tests {
...
+            downloader,
+            DirectoryManagerFs,
+            PathBuf::from(remote_dir.path()),
+        );
+
+        let result = pm.install(&agent_id, package_data);
+        assert!(matches!(
+            result,
+            Err(OCIPackageManagerError::PostDownloadHook(_))
+        ));
```
sigilioso, line 989: When reading the test, the fact of failing to check that the post-download was called surprised me. Maybe we could add a comment explaining why we check it like that.

agent-control/src/package/oci/package_manager.rs
```diff
@@ -938,6 +950,112 @@ mod tests {
...
+    fn test_install_runs_successful_hook_when_package_already_present() {
+        use crate::agent_type::runtime_config::on_host::executable::rendered::{Args, Env};
+        use crate::agent_type::runtime_config::on_host::package::rendered::PostDownloadHook;
+        use std::collections::HashMap;
+
+        let mut downloader = MockOCIDownloader::new();
+        let agent_id = AgentID::try_from("agent-id").unwrap();
+        let mut package_data = test_package_data();
+        package_data.post_download_hook = Some(PostDownloadHook {
+            path: "/usr/bin/true".to_string(),
```
sigilioso: We could use the current binary (the compiled test executable) and run it with `--list` to make the test work on Windows as well. We did something similar on the [copier tests](https://github.com/newrelic/newrelic-agent-control/blob/28d5f15158e19c603e1cae572b030d0c25cf139f/fs/src/file/copier.rs#L118).

agent-control/tests/on_host/oci_package_manager.rs
```diff
@@ -139,3 +139,92 @@ fn test_install_skips_download_if_exists_with_oci_registry() {
...
     );
 }
+
+// Verifies that the post-download hook runs on *every* install, including when the package files
+// are already present on disk (an Agent Control restart or a rollback to a previously-installed
+// version). The hook appends a line to a counter file each time it runs; after two installs of the
+// same package the file must contain two lines even though the second install skips download and
+// extraction.
+#[test]
+#[cfg(unix)]
```
sigilioso: Would making the integration test work on Windows be too complex? It would help improving coverage.

sigilioso (commented):
Thanks for addressing this! I've left some question, mostly _nits_

agent-control/src/package/oci/package_manager.rs
vjripoll: deleted by error

agent-control/src/package/oci/package_manager.rs
vjripoll: yes moved to the top

agent-control/src/package/oci/package_manager.rs
vjripoll: comments added

agent-control/src/package/oci/package_manager.rs
vjripoll: changed

sigilioso (approved):
LGTM

PR #2673, opened 2026-07-08
feat: modify infra agent agent types

danielorihuela:
Update infra-agent agent type to use the shared filesystem.

* `integrations.d` is removed in favour of the new shared folder for configs.
* `integrations` is removed in favour of the new shared folder for binaries.
* all embedded binaries are copied to the shared folder for binaries. fluent bit stays the same.
* `testing_agent_attributes` was modified to include as `remote_dir` the "root" path. Otherwise, `all_agent_type_definitions_are_resilient_linux` and it's windows counterpart would fail.
* modified infra agent e2e tests to verify that shared folders and files are created.


sigilioso (approved):
LGTM

PR #2674, opened 2026-07-08
feat: expose shared-filesystem declared paths [NR-579376]

sigilioso:
This PR adds a method to `OnHost` runtime to expose the shared-filesystem declared paths.

## Scope

This will used in next iterations to both:
- Perform AC remote config validation and detect two agents declaring conflicting shared paths
- Implement the shared-fs cleanup logic when an agent is removed from the lists (and when AC starts).

There is no behavior change for now, this PR just exposes the data for next iterations.

### DIFF
--- a/agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
+++ b/agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
@@ -10,7 +10,7 @@
 //! (`${nr-sub:filesystem_agent_dir}`).
 
 use std::{
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     io::{Error as IOError, ErrorKind},
     path::{Component, Path, PathBuf},
 };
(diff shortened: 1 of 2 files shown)


agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -179,6 +179,51 @@ impl Templateable for SharedFileSystem {
...
+/// The shared-filesystem paths an Agent Type declares ownership of, rooted under a base directory
+/// and split by ownership granularity.
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct DeclaredSharedPaths {
+    /// Files owned individually. Several agents may drop sibling files into the same
+    /// co-owned directory, but no two may own the exact same file path.
+    pub files: HashSet<PathBuf>,
+    /// Directories owned as a whole: the entire subtree is managed as a unit, and no
+    /// other agent may own anything at or under the path.
+    pub managed_dirs: HashSet<PathBuf>,
```
danielorihuela, line 191: How do you know that this directores are shared? Even if you use the `dir_content_from_map`, you could be using it on a shared folder. Nothing makes that impossible.

Let's say this assumption is accurate. We are eventually going to remove the `config_integrations`, and this is the only use case for `dir_content_from_map`. Maybe we can get rid of it sooner than later, and avoid adding extra code here?

sigilioso: We use this field to precisely represent `dir_content_from_map`, this is the only use case where the directory content depends on variables. If an agent uses `dir_content_from_map` on dir `a` we won't allow any other agent to write to directory 'a' and we will delete the whole directory 'a' when such agent is removed.

danielorihuela: But I mean, we could theoretically have a shared folder that is declared as a `dir` in one agent type and as a `dir_content_from_map` in another, right? It's a minor thing.

Actually, the "important" thing is if we will need it at all after a couple of days, but I don't want to block it.

danielorihuela (approved):

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
sigilioso: That's actually a good point because it could cause issues depending on how the cleanup is implemented. We can apply these rules when cleaning up:

- `file`: remove it
- `dir`: remove it **if it is empty**
- `dir_content_from_map`: remove the files and remove the directory **if it is empty** after removing the files

> Actually, the "important" thing is if we will need it at all after a couple of days, 

I Agree! Maybe we can remove it next week! 😄

PR #2675, opened 2026-07-08
feat(version): report all package versions in attributes

paologallinaharbur:
This PRs adds the version of the packages as identifying attributes. 
We could also report them as non-identifying, but I did not want to diverge with the existing attribute

### DIFF
--- a/agent-control/src/sub_agent/on_host/supervisor.rs
+++ b/agent-control/src/sub_agent/on_host/supervisor.rs
@@ -2,7 +2,9 @@
 //! and version checks for a single sub-agent.
 
 use crate::agent_control::agent_id::AgentID;
-use crate::agent_control::defaults::OPAMP_AGENT_VERSION_ATTRIBUTE_KEY;
+use crate::agent_control::defaults::{
+    OPAMP_AGENT_VERSION_ATTRIBUTE_KEY, OPAMP_PACKAGE_VERSION_ATTRIBUTE_KEY_PREFIX,
+};
 use crate::agent_type::runtime_config::health_config::rendered::OnHostHealthConfig;
 use crate::agent_type::runtime_config::on_host::filesystem::rendered::{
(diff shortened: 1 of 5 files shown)


agent-control/src/sub_agent/on_host/supervisor.rs
```diff
@@ -318,14 +320,25 @@ where
             version=%version,
             "Agent version determined from OCI package; publishing AgentAttributesUpdated event"
         );
+        let mut identifying_attributes = HashMap::from([(
+            OPAMP_AGENT_VERSION_ATTRIBUTE_KEY.to_string(),
+            version.into(),
+        )]);
+
+        // Per-package versions as identifying attributes: `package.version.<id>`.
+        self.packages_config
```
danielorihuela, line 329: The version for the "main" package is duplicated. I remember discussing that, but not the outcome. Was it okay?

danielorihuela:
> We could also report them as non-identifying, but I did not want to diverge with the existing attribute

Is this really a good approach? I don't want to block it, but I want to understand why is that okay?
I was wondering what would happen in the following situations.

We have an agent type that has two OCI packages.
We have to update the "non-main" package to fix a vulnerability issue, but the main package versions stays the same.
If the "non-main" package is sent as an identifying attribute, that means that we can check telemetry apart. Do we want that?

I'm always super confused with all this jargon.

danielorihuela (approved):

agent-control/src/sub_agent/on_host/supervisor.rs
paologallinaharbur: Yes I though about that, IMO it is ok
The idea is to deprecate the other if in the future we have often more packages and the UI start supporting it.

As I see they are two different things, in one case it is the whole agent version in the other the single package version 🤔

alvarocabanas (approved):
LGTM, and I also think it makes sense in identifying attributes, at the end is an identifier for the packages that are composing the agent.

PR #2676, opened 2026-07-08
fix: unsigned package test and other failure tests

vjripoll:
ISSUE
---
Each of these tests waits for the remote-config status to be Failed with a specific message (signature verification, requested version does not exist, pre-flight check failed). That message only shows up when an upgrade attempt actually runs and fails.

The problem: after the first failure, the self-update backoff gate goes into a cooldown. From then on, every poll just reports a generic, stable message: upgrade to X suppressed: retrying after previous failure. The fake OpAMP server only keeps the last status, so the real error message gets overwritten by the "suppressed" one almost right away.

With the default backoff (starts at 30s and keeps growing), that cooldown lasts way longer than the test's polling window, so the test almost always reads "suppressed" and eventually times out. It only passed when a poll happened to land on the short moment where the real error was still the latest status. That timing is why it was flaky, and why it kept failing on the slower Windows CI.

FIX
---
Add create_self_update_no_cooldown_config: a self-update config with the backoff set to zero (upgrade_backoff.base_delay: 0s, max_delay: 0s), with signature_verification_enabled as a parameter, and make the three failing tests use it.

With a zero delay the gate never sees now < next_attempt_at, so it never goes into cooldown and never suppresses. Every poll runs the attempt again and reports the real error, so the real error is now the status that stays, and the test reliably sees it.

### DIFF
--- a/agent-control/tests/on_host/scenarios/ac_self_update.rs
+++ b/agent-control/tests/on_host/scenarios/ac_self_update.rs
@@ -212,7 +212,7 @@ fn test_ac_self_update_fails_for_unsigned_package_with_oci_registry() {
 
     let dirs = TempBasePaths::default();
 
-    create_self_update_local_config(&opamp_server, &signer, &dirs.local_dir(), true);
+    create_self_update_no_cooldown_config(&opamp_server, &signer, &dirs.local_dir(), true);
 
     let mut agent_control = start_agent_control_with_custom_config(
         dirs.base_paths().clone(),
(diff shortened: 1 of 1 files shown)


agent-control/tests/on_host/scenarios/ac_self_update.rs
```diff
@@ -552,6 +552,45 @@ fn create_self_update_local_config(
...
+    public_key_server_url: {}
+agents: {{}}
+oci:
+  registry: {OCI_TEST_REGISTRY_URL}
+self_update:
+  enabled: true
+  signature_verification_enabled: {signature_verification_enabled}
+  upgrade_backoff:
+    base_delay: 0s
+    max_delay: 0s
```
sigilioso, line 580: Do we still have tests covering the use case when these are set to something different from `0s`? If not, shall we add a single test with short delays checking that the OpAMP message is updated as expected?

sigilioso (commented):
Thanks for addressing this! I left a single question, everything else looks good!

agent-control/tests/on_host/scenarios/ac_self_update.rs
vjripoll: The non-zero backoff behavior is already covered, but the exact OpAMP message isn't with the backoff diferent to 0

sigilioso: 👍 We weren't check that message before, we could add a new test to start checking it now or not, your call! 😄

sigilioso (approved):

PR #2678, opened 2026-07-09
chore(custom-agents): improve docs

paologallinaharbur:
This PR aims to document how to leverage custom agent types while developing. 

I am changing the location of the readme of the agentTypes so that all docs are in a unique place

Related to https://github.com/newrelic/helm-charts/pull/2333

### DIFF
--- a/docs/agent_type.md
+++ b/docs/agent_type.md
@@ -409,7 +409,17 @@ deployment:
     initial_delay: 10s # Defaults to 30s.
 ```
 
-## Development
+## Development Custom agent types
+
+Agent Control ships a set of embedded agent types (see
+[`agent-control/agent-type-registry/newrelic/`](../agent-control/agent-type-registry/newrelic)). You can add your own —
+or override an embedded one — by placing the definition YAML in the *dynamic agent types* directory. Custom definitions
+take precedence over embedded ones with the same id.
+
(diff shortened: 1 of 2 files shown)


docs/agent_type.md
```diff
@@ -387,7 +387,17 @@ deployment:
     initial_delay: 10s # Defaults to 30s.
 ```
 
-## Development
+## Development Custom agent types
+
+Agent Control ships a set of embedded agent types (see
+[`agent-control/agent-type-registry/newrelic/`](../agent-control/agent-type-registry/newrelic)). You can add your own —
+or override an embedded one — by placing the definition YAML in the *dynamic agent types* directory. Custom definitions
+take precedence over embedded ones with the same id.
```
sigilioso:
```suggestion
Agent Control ships a set of embedded agent types (see
[`agent-control/agent-type-registry/newrelic/`](../agent-control/agent-type-registry/newrelic)). You can add your own —
or override an embedded one — by placing the definition YAML in the *dynamic agent types* directory. Custom definitions
take precedence over any other definition with the same id (namespace + name + version).
```

Should we already consider that this would also override the remote ones?

paologallinaharbur: I think it is nice to have this kind of "open description" that covers all use cases

sigilioso (approved):

PR #2679, opened 2026-07-09
feat(on-host): add validation for conflicting files in shared-fs [NR-579376]

sigilioso:
## Summary

When multiple sub-agents share the on-host shared filesystem (/var/lib/newrelic-agent-control/shared-filesystem), two agents declaring the same file would silently fight over it (last write wins), leaving the shared folder with un-predictable content. This PR adds a single-owner rule: a remote config that would make two agents claim the same shared path is rejected up front and reported as Failed.

## Details

- Added a new validator for on-host implementation: `SharedFileSystemPathValidator`. It detects if there is any conflicting file in the filesystem for the agents in the list.
- Since the validator needs to load the corresponding Agent Types to check conflicts the previous check regarding Agent Type definition  is also covered.

## Notes

- ⚠️ The validation runs on remote-config only. This means that a startup local configuration is not validated (the previous validation for agent-type existence wasn't performed). We will address this separately, since is a broader consideration that applies on both environments on-host and k8s. 
- `dir_content_from_map` in the filesystem is supported as a whole owned directory since its content cannot be statically known.


agent-control/src/agent_control/config_validator/on_host.rs
```diff
@@ -0,0 +1,340 @@
...
+        Self {
+            agent_id,
+            agent_type,
+            claimed,
+        }
+    }
+
+    /// Returns the first pair of claimed paths (one from each agent) that overlap, if any.
+    /// Check [SharedFilesystemPathValidator] docs for details.
+    fn overlapping_path<'a>(&'a self, other: &'a Self) -> Option<(&'a Path, &'a Path)> {
```
danielorihuela:
```suggestion
    fn overlapping_path<'a>(&self, other: &'a Self) -> Option<(&Path, &'a Path)> {
```

I noticed this is not needed. Not sure it makes it easier to be honest. Just sharing it.

agent-control/src/agent_control/config_validator/on_host.rs
```diff
@@ -0,0 +1,340 @@
...
+        self.claimed.iter().find_map(|own| {
+            other
+                .claimed
+                .iter()
+                .find(|other| own.starts_with(other) || other.starts_with(own))
```
danielorihuela, line 55: Should we use `start_with`? What if we have `/some_file_config.yaml` and `/some_file_config.yaml.yaml`. These should be different. The name is strange, but allowed.

agent-control/src/agent_control/config_validator/on_host.rs
```diff
@@ -0,0 +1,340 @@
...
+            agent_id,
+            agent_type,
+            claimed,
+        }
+    }
+
+    /// Returns the first pair of claimed paths (one from each agent) that overlap, if any.
+    /// Check [SharedFilesystemPathValidator] docs for details.
+    fn overlapping_path<'a>(&'a self, other: &'a Self) -> Option<(&'a Path, &'a Path)> {
+        self.claimed.iter().find_map(|own| {
```
danielorihuela, line 51: I wonder if this is more readable:

```rust
let mut uniques = HashSet::new();

let mut all_paths = self.claimed.iter().chain(other.claimed.iter());
let first_duplicate = all_paths.find(|&path| !uniques.insert(path));
```

Then, we could return just one path. At the end, they are duplicated, why do we need to pass both?

agent-control/src/agent_control/config_validator/on_host.rs
sigilioso: You are absolutely right, let's make it slightly smarter 👍

agent-control/src/agent_control/config_validator/on_host.rs
sigilioso: We need both because they are not necessarily duplicated, they could be overlapping. Example:

- Agent A:  `dir_content_from_map` in folder `some_folder/`
- Agent B: `file` in `some_folder/some-file.txt`

If we finally get rid of `dir_content_from_map` we can simplify.

agent-control/src/agent_control/config_validator/on_host.rs
sigilioso: We already had it covered. [Path::starts_with](https://doc.rust-lang.org/beta/std/path/struct.Path.html#method.starts_with) maches whole path components instead of string prefixes. I've added two new unit tests in [fc3f720](https://github.com/newrelic/newrelic-agent-control/pull/2679/commits/fc3f72057de55ea36f2c37c06f8ad416d935785f) that also document this behavior.

agent-control/src/agent_control/config_validator/on_host.rs
```diff
@@ -0,0 +1,424 @@
...
+
+/// Enforces the single-owner rule for the on-host shared filesystem. For files declared by two different agents:
+/// * Two files with the same path overlap
+/// * A declared directory and a child file overlap
+/// * Two siblings files in the same undeclared directory do not overlap
+///
+/// Shared paths are static (the Agent Type entry keys), so this reads them straight from the
+/// registry without rendering. Resolving each agent type also surfaces unknown-type errors, therefore
+/// the [RegistryDynamicConfigValidator](super::RegistryDynamicConfigValidator) check is addressed by this validator.
+pub struct SharedFilesystemPathValidator<R> {
```
danielorihuela: We should add the bound type for `R`. `R: AgentTypeRegistry`

agent-control/src/agent_control/config_validator/on_host.rs
```diff
@@ -0,0 +1,424 @@
...
+    pub fn new(registry: Arc<R>) -> Self {
+        Self { registry }
+    }
+}
+
+impl<R> DynamicConfigValidator for SharedFilesystemPathValidator<R>
+where
+    R: AgentTypeRegistry,
+{
+    fn validate(
```
alvarocabanas, line 84: Should this validator also be called on first start up? 
This is only triggered when a remote_config arrives with conflicting agent types, should we also fail the whole agent control if they set conflicting paths on custom local agent_types?

Also, can we document somewhere the behavior of this validation (when remote config arrives and current config already has a file it discards remote and keeps local and so on), it makes sense both for the user and for us. Not sure if I would add it to the docs if we want to document it for the user, but I would also put it as a comment in this Validator or in the caller.

danielorihuela:
> The validation runs on remote-config only. This means that a startup local configuration is not validated (the previous validation for agent-type existence wasn't performed). We will address this separately, since is a broader consideration that applies on both environments on-host and k8s.

Have you considered how will this be done? I was thinking that maybe we can do it as part of the `build.rs` or a CI pipelines check? We could have an agent type crate, and start putting there all the code we are using. Test helpers to create agent types, agent type validations, etc. Agent control and other teams could use that crate.

agent-control/src/agent_control/config_validator/on_host.rs
sigilioso: > Should this validator also be called on first start up?
> This is only triggered when a remote_config arrives with conflicting agent types, should we also fail the whole > agent control if they set conflicting paths on custom local agent_types?

I think that not only this validator but any validator that runs for remote. But, as stated in the description, I left this out of scope and opened a ticket to tackle it.

> Also, can we document somewhere the behavior of this validation (when remote config arrives and current config already has a file it discards remote and keeps local and so on), it makes sense both for the user and for us. Not sure if I would add it to the docs if we want to document it for the user, but I would also put it as a comment in this Validator or in the caller.

I'll check if we can improve the documentation, I hoped the unit tests were documenting the behavior clearly. After all, the validator behavior hasn't changed: if the list of agent is invalid the remote configuration is rejected (and not applied). We simply have added an additional condition to the list of agents.

sigilioso:
@danielorihuela 

> Have you considered how will this be done? I was thinking that maybe we can do it as part of the build.rs or a CI pipelines check? We could have an agent type crate, and start putting there all the code we are using. Test helpers to create agent types, agent type validations, etc. Agent control and other teams could use that crate.

But it is a runtime validation, isn't it? Two agents can define the same path in the same folder under the shared folder and that's perfectly fine unless they are used at the same time. 

I opened a ticket to run this validator when AC starts (and define the behavior if validation failed). We could try to restrict the usage of conflicting files when the Agent Type but I'm not sure if it is feasible considering that Agent Types can be defined anywhere including custom Agent Types 🤔

danielorihuela:
@sigilioso 

> But it is a runtime validation, isn't it? Two agents can define the same path in the same folder under the shared folder and that's perfectly fine unless they are used at the same time.


For remote configs runtime makes sense. For embedded configs, not necessarily. I would argue that this is an issue on our side. If they are embedded, why don't we make sure that they work with each other? We should embed a minimal number of agent types at the end, and they should be "official". I would expect them to work.

I haven't thought about custom types though. It's just an idea I wanted to threw in case it was simpler to implement. It probably isn't. It might be worth having the logic separated anyway, so that it can be used by other teams, but it's probably to soon to know that.

sigilioso:
> We should embed a minimal number of agent types at the end, and they should be "official". I would expect them to work.

I'm not so sure about that. We will keep a small set of Agent Types embedded for backwards compatibility reasons but, once non-embedded are available, embedded will likely be there just for that reason. I expect the infra-agent's Agent Type `0.2.0` (or `1.0.0` or whatever) to be available through the OCI registry.

If we supported embedded Agent Types only I would definitely agree. But I think it won't be the case. For that reason I don't think we can implement these kinds of validations there.

> I haven't thought about custom types though. It's just an idea I wanted to threw in case it was simpler to implement. It probably isn't. It might be worth having the logic separated anyway, so that it can be used by other teams, but it's probably to soon to know that.

It is actually a good point, and we definitely need to provide some tools to validate Agent Types before publishing them but I'm not sure if we can avoid conflicting files completely there.

alvarocabanas (dismissed):
LGTM

agent-control/src/agent_control/config_validator/on_host.rs
sigilioso: Added a comment in the caller (to clarify that _invalid_ configuration makes the apply fail): [d12dde6](https://github.com/newrelic/newrelic-agent-control/pull/2679/commits/d12dde60016f8d74464427bb03690f002dcd684b)

agent-control/src/agent_control/config_validator/on_host.rs
sigilioso: Bound added here: [d12dde6](https://github.com/newrelic/newrelic-agent-control/pull/2679/commits/d12dde60016f8d74464427bb03690f002dcd684b). Thanks!

danielorihuela (approved):

PR #2681, opened 2026-07-09
fix: onhost supervisor reloading test

vjripoll:
***ISSUE***
---
File logging reload: run_file_logging_scenario dropped the StartedAgentControl handle on return, so Agent Control stopped before the test's final retry. That retry could never wait for the post-reload log to appear (AC was already gone), and the only window for the second-run log was a fixed 5s sleep, which slow CI exceeded.

Self-replacer: the workspace tests run in parallel. While one test writes a binary (fs::copy/fs::write holding a writable fd), another test fork (to spawn its process) inherits that fd, so executing the just-written binary intermittently fails with ETXTBSY ("Text file busy").

***FIX***
---
File logging: the scenario now returns the StartedAgentControl handle and the test keeps it alive, so AC stays running while the test waits. The final retry now waits for the full expected log state (both messages present/absent as expected, up to 60s) instead of checking once after AC had stopped.

Self-replacer: serialize the write+exec tests behind a shared lock so only one write+exec sequence runs at a time, and build the example binary at most once (Once) to remove the concurrent-build.

### DIFF
--- a/agent-control/tests/on_host/scenarios/file_logging.rs
+++ b/agent-control/tests/on_host/scenarios/file_logging.rs
@@ -2,8 +2,10 @@ use std::{fs, io, path::Path, time::Duration};
 
 use crate::{
     common::{
-        agent_control::start_agent_control_with_custom_config, base_paths::TempBasePaths,
-        retry::retry, runtime::tokio_runtime,
+        agent_control::{StartedAgentControl, start_agent_control_with_custom_config},
+        base_paths::TempBasePaths,
+        retry::retry,
+        runtime::tokio_runtime,
     },
     on_host::tools::{
         config::{OnHostAgentControlConfigBuilder, create_file, create_local_config},
@@ -123,7 +125,7 @@ fn run_file_logging_scenario(
     initial_message: &str,
     reload_file_logging: bool,
     reload_message: &str,
-) -> TempBasePaths {
+) -> (TempBasePaths, StartedAgentControl) {
     let mut opamp_server = FakeServer::start(tokio_runtime().handle());
 
     let dirs = TempBasePaths::default();
@@ -154,13 +156,13 @@ fn run_file_logging_scenario(
         dirs.local_dir(),
     );
 
-    let _agent_control =
+    let agent_control =
         start_agent_control_with_custom_config(dirs.base_paths(), AGENT_CONTROL_MODE_ON_HOST);
 
     let sub_agent_instance_id =
         get_instance_id(&AgentID::try_from(agent_id).unwrap(), dirs.base_paths());
 
-    // Give some time for the echo output to be captured in the log files
+    // Let the first run write its log before the reload can change/disable logging.
     std::thread::sleep(Duration::from_secs(5));
 
     // Trigger a reload by sending a new remote config via OpAMP with updated values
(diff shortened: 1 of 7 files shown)


self-replacer/tests/test_helpers.rs
```diff
@@ -2,6 +2,21 @@
 
 use std::fs;
 use std::path::{Path, PathBuf};
+use std::sync::{Mutex, Once};
+
+/// Serializes write+exec tests to avoid ETXTBSY (a concurrent fork inheriting the writable fd of a
+/// binary being written makes its exec fail with "Text file busy").
+static EXEC_SERIAL_LOCK: Mutex<()> = Mutex::new(());
```
sigilioso: I think we use [serial_test](https://docs.rs/serial_test/latest/serial_test/) in other tests that can be problematic when executed in parallel. Does the `Mutex` approach cover something that using the `#[serial]` macro doesn't?

agent-control/tests/on_host/scenarios/file_logging.rs
```diff
@@ -172,7 +174,7 @@ fn run_file_logging_scenario(
     // Give some time for the echo output to be captured in the log files
     std::thread::sleep(Duration::from_secs(5));
 
-    dirs
```
sigilioso, line 175: I'd like to confirm that I'm understanding the issue and the fix. 

Before, we had a 5s. sleep keep the AC proccess running to write the logs and, sometimes, the expected logs weren't written on time and that made the test fail. Now, we keep the AC process running for as long as the test runs and we rely on the check's retries to give some time to the process to write the files.

If that's right, would it make sense to remove this 5s. sleep and just adjust the retries in the corresponding test(s)?

sigilioso (commented):
Two minor questions. Thanks!

self-replacer/tests/test_helpers.rs
vjripoll: You're right, I didn't even remember the serial_test. I'll change it and add the dependency. Thank you!

agent-control/tests/on_host/scenarios/file_logging.rs
vjripoll: Right on both counts. Removed the post-reload sleep and let the retry (up to 60s, with AC now kept alive) wait for the logs. Kept the pre-reload sleep, since it ensures the first run's log lands before the reload can disable logging.

danielorihuela:
👍

danielorihuela (dismissed):

self-replacer/Cargo.toml
```diff
@@ -17,6 +17,7 @@ tempfile = { workspace = true }
 assert_matches = { workspace = true }
 assert_cmd = "2.2.2"
 predicates = "3.1.4"
+serial_test = "3.5.0"
```
sigilioso: Nit: could we add `serial_test` to the workspace dependencies? After all, the agent-control crate also uses it.

self-replacer/tests/integration_tests.rs
```diff
@@ -19,6 +20,7 @@ use self_replacer::BACKUP_SUFFIX;
 // ============================================================================
 
 #[test]
+#[serial]
```
sigilioso, line 29: Nit: should we add a short comment explaining why these need to be serial. We found it out the hard way and it can be easy to forget.

sigilioso (dismissed):
I've left two _nits_, feel free to ignore them

sigilioso (approved):
LGTM

PR #2682, opened 2026-07-09
fix: net fails on dependencies and chartmuseum

vjripoll:
Makes the CI setup resilient to transient network failures when downloading dependencies and artifacts, without adding any retry around the tests themselves. Covers the download points that were failing intermittently in the k8s / checks jobs.

***CHANGES***
---

Cargo dependencies
- .cargo/config.toml: [net] retry = 10 (up from the default 3). Applies to every cargo network op across all jobs; only affects downloads.
- k8s job: a dedicated cargo fetch --locked step with retry, and the test step runs with CARGO_NET_OFFLINE=true, so once deps are fetched a network hiccup can't fail the tests (and the download retry can't mask a flaky test).

chartmuseum container image
- Preload step in the workflow: docker pull (with retry) + minikube image load of ghcr.io/helm/chartmuseum:v0.16.3 (ghcr login to avoid anonymous rate limits).
- Tilt file pins the image (image.tag=v0.16.3, pullPolicy: IfNotPresent) so the pod uses the side-loaded image instead of pulling at runtime.

chartmuseum helm chart
- Tilt file pre-pulls the chart with retry into a local dir and installs from that local copy, so helm upgrade --install no longer downloads the chart tgz from GitHub release assets at runtime (that download was timing out).

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,3 +1,6 @@
+[net]
+retry = 10
+
 [target.x86_64-unknown-linux-musl]
 rustflags = ["-C", "target-feature=+crt-static"]
 
--- a/.github/workflows/push_pr_checks_tests.yml
+++ b/.github/workflows/push_pr_checks_tests.yml
@@ -212,7 +212,7 @@ jobs:
           identifier: 'test-dev'
           restore-strategy: 'nearest'
           save-cache: false
-      
+
       - if: ${{ matrix.os == 'windows-latest' }}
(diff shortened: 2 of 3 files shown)


danielorihuela (approved):
Tilt is not my thing. It would be great another pair of eyes.

From my side, the changes look good!!! I like what you propose 💯

alvarocabanas (approved):
LGTM

PR #2683, opened 2026-07-10
fix: add test to cover opamp message in backoff

vjripoll:
Adds a unit test for the message rendered by UpdaterError::UpdateInCooldown.


sigilioso (approved):
LGTM

PR #2684, opened 2026-07-10
feat: remove config integrations variable

danielorihuela:
`config_integrations` variable removed.

* tests updated
* documentation updated

### DIFF
--- a/agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
+++ b/agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
@@ -800,9 +800,6 @@ agent:
   entries:
     data:
       kind: dir
-    integrations.d:
-      kind: dir_content_from_map
-      source: ${nr-var:config_integrations}
     newrelic-infra.yaml:
       kind: file
       text: ${nr-var:config_agent}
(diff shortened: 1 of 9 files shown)


agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -800,9 +800,6 @@ agent:
   entries:
     data:
       kind: dir
-    integrations.d:
-      kind: dir_content_from_map
-      source: ${nr-var:config_integrations}
```
alvarocabanas, line 805: I would not remove this test, Maybe it could be modified with something generic, but we have not removed the dir_content_from_map feature, removing this will remove the coverage for it.

alvarocabanas (commented):
I left a comment about the tests

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
danielorihuela: `dir_content_from_map` is used on line 795. I think we are okay.

alvarocabanas: True, thanks for pointing it out!

alvarocabanas (approved):
LGTM

PR #2686, opened 2026-07-10
Feat: dynamic canaries

alvarocabanas:
- Changes fleet_deployment.sh to use the Fleet API in newrelic-cli instead of using Graphql directly.
- Adds a scheduled GitHub Action (every 6h) that rotates the canary fleets through an ordered set of agent config versions, advancing one step per run and wrapping around. Each run fans out over a matrix of environment (staging/production) and platform (linux/windows), deploying via the newrelic-CLI-based fleet_deployment.sh.

### DIFF
--- a/.github/workflows/scheduled_onhost_canary_config_rotation.yml
+++ b/.github/workflows/scheduled_onhost_canary_config_rotation.yml
@@ -36,14 +36,24 @@ env:
   # We don't have 'latest' OCI packages, so I've set them to a fixed version.
   #   First  -> NRInfra 1.77.1  (canaries-infra-config)
   #   Second -> NRDOT   1.19.0  (canaries-nrdot)
-  AGENT_SPECS_LINUX: >-
+  AGENT_SPECS_LINUX_PRODUCTION: >-
     NRInfra:1.77.1:NjQyNTg2NXxOR0VQfEFHRU5UX0NPTkZJR1VSQVRJT05fVkVSU0lPTnwwMTlmNGM3Yi1lYjAyLTc1MmQtOGViOC0zNDYyMWM2ZDRkMzQ
     NRDOT:1.19.0:NjQyNTg2NXxOR0VQfEFHRU5UX0NPTkZJR1VSQVRJT05fVkVSU0lPTnwwMTlmNGM3Yy05M2Y3LTcyMjctODJmZS01OGYwMjcyMDViMDI
   #   First  -> NRInfra 1.77.1  (canaries-infra-config-windows)
   #   Second -> NRDOT   1.19.0  (canaries-nrdot-windows)
-  AGENT_SPECS_WINDOWS: >-
+  AGENT_SPECS_WINDOWS_PRODUCTION: >-
     NRInfra:1.77.1:NjQyNTg2NXxOR0VQfEFHRU5UX0NPTkZJR1VSQVRJT05fVkVSU0lPTnwwMTlmNGNjNC0zNmU1LTdmMmQtYjIyMS1mM2ZjY2QzOTAwOTg
     NRDOT:1.19.0:NjQyNTg2NXxOR0VQfEFHRU5UX0NPTkZJR1VSQVRJT05fVkVSU0lPTnwwMTlmNGNjNC1kMTE0LTdmMzktYWRmZS02ZTY4ZWFiYWZkY2Y
+  #   First  -> NRInfra 1.77.1  (canaries-infra-config)
+  #   Second -> NRDOT   1.19.0  (canaries-nrdot)
+  AGENT_SPECS_LINUX_STAGING: >-
+    NRInfra:1.77.1:MTIyMTMwNjh8TkdFUHxBR0VOVF9DT05GSUdVUkFUSU9OX1ZFUlNJT058MDE5ZjViZTItYjdkYy03YmM5LTg0OTUtYTIyMmI5NTYyZDAy
+    NRDOT:1.19.0:MTIyMTMwNjh8TkdFUHxBR0VOVF9DT05GSUdVUkFUSU9OX1ZFUlNJT058MDE5ZjViZTMtNmU3YS03NzA0LWI0NjgtOTAwMDY1NjgyMWJj
+  #   First  -> NRInfra 1.77.1  (canaries-infra-config-windows)
+  #   Second -> NRDOT   1.19.0  (canaries-nrdot-windows)
+  AGENT_SPECS_WINDOWS_STAGING: >-
+    NRInfra:1.77.1:MTIyMTMwNjh8TkdFUHxBR0VOVF9DT05GSUdVUkFUSU9OX1ZFUlNJT058MDE5ZjViZTMtNmU3YS03NzA0LWI0NjgtOTAwMDY1NjgyMWJj
+    NRDOT:1.19.0:MTIyMTMwNjh8TkdFUHxBR0VOVF9DT05GSUdVUkFUSU9OX1ZFUlNJT058MDE5ZjViZTQtOWFkZS03MWVjLTg0NTMtYjlkOTExMGViODc3
 
 jobs:
   rotate-config-deployment:
@@ -65,20 +75,22 @@ jobs:
           PLATFORM: ${{ matrix.platform }}
           ENVIRONMENT: ${{ matrix.environment }}
         run: |
-          # Pick this platform's fleet id (per environment) and its ordered spec list.
+          # Pick this platform+environment's fleet id and its ordered spec list.
           if [ "${PLATFORM}" = "windows" ]; then
-            SPECS_STR="${AGENT_SPECS_WINDOWS}"
             if [ "${ENVIRONMENT}" = "production" ]; then
               FLEET_ID="${WINDOWS_FLEET_ID_PRODUCTION}"
+              SPECS_STR="${AGENT_SPECS_WINDOWS_PRODUCTION}"
             else
               FLEET_ID="${WINDOWS_FLEET_ID_STAGING}"
+              SPECS_STR="${AGENT_SPECS_WINDOWS_STAGING}"
             fi
           else
-            SPECS_STR="${AGENT_SPECS_LINUX}"
             if [ "${ENVIRONMENT}" = "production" ]; then
               FLEET_ID="${FLEET_ID_PRODUCTION}"
+              SPECS_STR="${AGENT_SPECS_LINUX_PRODUCTION}"
             else
               FLEET_ID="${FLEET_ID_STAGING}"
+              SPECS_STR="${AGENT_SPECS_LINUX_STAGING}"
             fi
           fi
 
--- a/.github/workflows/scheduled_onhost_canary_config_rotation.yml
+++ b/.github/workflows/scheduled_onhost_canary_config_rotation.yml
@@ -1,0 +1,110 @@
+name: 🔁🐤 Scheduled onhost canary config rotation
+
+# Every 6 hours, deploy the *next* config version in an ordered set to the
+# canary fleet using test/onhost-canaries/scripts/fleet_deployment.sh.
+# Each run advances one step through the set; after the last entry it wraps
+# back to the first. This is a rotation (one deployment per run), not a matrix
+# fan-out.
+
+on:
+  schedule:
+    # Every 6 hours (UTC). Minute offset from :00 to dodge the top-of-hour
+    # scheduler backlog that delays cron workflows on GitHub-hosted runners.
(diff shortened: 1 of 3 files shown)


test/onhost-canaries/scripts/fleet_deployment.sh
```diff
@@ -61,17 +65,22 @@ fi
...
-# When a new fleet is created both rings are always created, all hosts added to a fleet are added to default unless
-# they are moved to canary; to warrant all hosts are always deployed, we always deploy both rings, even if in the vast
-# majority of cases only default exists.
-RINGS_TO_DEPLOY='["canary", "default"]'
-
+# When a new fleet is created both rings are always created; all hosts added to
+# a fleet land in `default` unless moved to `canary`. To guarantee every host is
+# always deployed we always deploy both rings, even though in the vast majority
+# of cases only `default` exists. The CLI expects a comma-separated list.
+RINGS_TO_DEPLOY="canary,default"
```
danielorihuela, line 77: I lack contenxt. Shouldn't we deploy only to the canary ring?

.github/workflows/scheduled_onhost_canary_config_rotation.yml
```diff
@@ -0,0 +1,112 @@
...
+  FLEET_ID_PRODUCTION: "NjQyNTg2NXxOR0VQfEZMRUVUfDAxOWFlM2EyLTA3OTctNzczYS05Y2JjLWMzNzZkMjAwMWFkZg"
+  WINDOWS_FLEET_ID_PRODUCTION: "NjQyNTg2NXxOR0VQfEZMRUVUfDAxOWMyMjljLWVhYTQtN2JmNi04YWIyLTU2ZWI0YjE2ZWZjZQ"
+
+  # Ordered set of agent specs to rotate through, one per line:
+  #   <agentType>:<version>:<configVersionId>
+  # Agent type, version and config version are paired here so they always rotate
+  # together and stay matched. One deployment per run advances to the next entry,
+  # wrapping to the first after the last.
+  #   First  -> NRInfra 1.77.1  (canaries-infra-config)
+  #   Second -> NRDOT   1.19.0  (canaries-nrdot)
```
danielorihuela: The config names are only for production linux. We could remove the list here, and just comment next to each config under `AGENT_SPECS_LINUX` and `AGENT_SPECS_WINDOWS` the name.

.github/workflows/scheduled_onhost_canary_config_rotation.yml
```diff
@@ -0,0 +1,112 @@
...
+          fi
+
+          # Split the whitespace-separated specs. Each entry is a full
+          # <agentType>:<version>:<configVersionId> spec.
+          read -ra SPECS <<< "${SPECS_STR}"
+          COUNT=${#SPECS[@]}
+          if [ "${COUNT}" -eq 0 ]; then
+            echo "ERROR: no agent specs configured for platform '${PLATFORM}'." >&2
+            exit 1
+          fi
```
danielorihuela: This can't happen. If platform is other than windows, we assume is linux. In that case we fill `SPECS_STR`. The count can never be 0.

test/onhost-canaries/scripts/fleet_deployment.sh
alvarocabanas: I've replicated what is done in the UI when deploying, it's done in batches, first it's deployed in canary and if the ring works it's deployed to default. I think you could be right we could deploy them directly in canary, but just in case I put both because at the moment I am not aware of us having control of which hosts go in each ring.

.github/workflows/scheduled_onhost_canary_config_rotation.yml
alvarocabanas: Removed in last commit.

.github/workflows/scheduled_onhost_canary_config_rotation.yml
alvarocabanas: I added a comment to the second block, I can't inline them because it's an env-var string and doesn't support comments.

danielorihuela: Okey. I have another doubt. This time regardind staging. I don't see any configuration named `canaries-infra-config`. Is it missing? Should we create it? What am I missing?

alvarocabanas: Oh, thanks for realizing it, it will definitely fail for staging without the configs, I need to add them.

alvarocabanas: Done in last commit.

.github/workflows/scheduled_onhost_canary_config_rotation.yml
```diff
@@ -0,0 +1,122 @@
...
+  #   First  -> NRInfra 1.77.1  (canaries-infra-config-windows)
+  #   Second -> NRDOT   1.19.0  (canaries-nrdot-windows)
+  AGENT_SPECS_WINDOWS_PRODUCTION: >-
+    NRInfra:1.77.1:NjQyNTg2NXxOR0VQfEFHRU5UX0NPTkZJR1VSQVRJT05fVkVSU0lPTnwwMTlmNGNjNC0zNmU1LTdmMmQtYjIyMS1mM2ZjY2QzOTAwOTg
+    NRDOT:1.19.0:NjQyNTg2NXxOR0VQfEFHRU5UX0NPTkZJR1VSQVRJT05fVkVSU0lPTnwwMTlmNGNjNC1kMTE0LTdmMzktYWRmZS02ZTY4ZWFiYWZkY2Y
+  #   First  -> NRInfra 1.77.1  (canaries-infra-config)
+  #   Second -> NRDOT   1.19.0  (canaries-nrdot)
+  AGENT_SPECS_LINUX_STAGING: >-
+    NRInfra:1.77.1:MTIyMTMwNjh8TkdFUHxBR0VOVF9DT05GSUdVUkFUSU9OX1ZFUlNJT058MDE5ZjViZTItYjdkYy03YmM5LTg0OTUtYTIyMmI5NTYyZDAy
+    NRDOT:1.19.0:MTIyMTMwNjh8TkdFUHxBR0VOVF9DT05GSUdVUkFUSU9OX1ZFUlNJT058MDE5ZjViZTMtNmU3YS03NzA0LWI0NjgtOTAwMDY1NjgyMWJj
```
danielorihuela: It should be `MTIyMTMwNjh8TkdFUHxBR0VOVF9DT05GSUdVUkFUSU9OX1ZFUlNJT058MDE5ZjViZTQtMWFmNC03YmM4LTlmZDgtZmQ3YzIzMWNlNzMx` I think.

The one here points to `canaries-infra-config-windows`. It's a copy paste from the one below.

.github/workflows/scheduled_onhost_canary_config_rotation.yml
```diff
@@ -0,0 +1,122 @@
...
+  # together and stay matched. One deployment per run advances to the next entry,
+  # wrapping to the first after the last.
+  # We don't have 'latest' OCI packages, so I've set them to a fixed version.
+  #   First  -> NRInfra 1.77.1  (canaries-infra-config)
+  #   Second -> NRDOT   1.19.0  (canaries-nrdot)
+  AGENT_SPECS_LINUX_PRODUCTION: >-
+    NRInfra:1.77.1:NjQyNTg2NXxOR0VQfEFHRU5UX0NPTkZJR1VSQVRJT05fVkVSU0lPTnwwMTlmNGM3Yi1lYjAyLTc1MmQtOGViOC0zNDYyMWM2ZDRkMzQ
+    NRDOT:1.19.0:NjQyNTg2NXxOR0VQfEFHRU5UX0NPTkZJR1VSQVRJT05fVkVSU0lPTnwwMTlmNGM3Yy05M2Y3LTcyMjctODJmZS01OGYwMjcyMDViMDI
+  #   First  -> NRInfra 1.77.1  (canaries-infra-config-windows)
+  #   Second -> NRDOT   1.19.0  (canaries-nrdot-windows)
```
sigilioso: Are any of the existing alerts affected by switching between the infrastructure agent and nr-dot?

alvarocabanas: Oh, good point, I didn't control it, and they will for sure fail, we are working on the samples, so the only way to make this work is making the alerts be compatible with both infra-agent and NRDOT.

gsanchezgavier:
cleaning up a bit the board, it can be re-opened later

.github/workflows/scheduled_onhost_canary_config_rotation.yml
```diff
@@ -0,0 +1,136 @@
...
+
+env:
+  # AC Staging Account `host-canaries-staging` fleet
+  FLEET_ID_STAGING: "MTIyMTMwNjh8TkdFUHxGTEVFVHwwMTlhZTNiNS01Yjg5LTdkNjYtYWU0MC1lNmZkOTY2ZDFhMDA"
+  # AC Staging Account `agent_control_canaries` fleet
+  WINDOWS_FLEET_ID_STAGING: "MTIyMTMwNjh8TkdFUHxGTEVFVHwwMTljMjNjYy0yM2I2LTc1OTYtODBkNy0yMzAzYWIyYzcwMTA"
+  # AC US Production Account `host-canaries-production` fleet
+  FLEET_ID_PRODUCTION: "NjQyNTg2NXxOR0VQfEZMRUVUfDAxOWFlM2EyLTA3OTctNzczYS05Y2JjLWMzNzZkMjAwMWFkZg"
+  # AC US Production Account `agent-control-canary` fleet
+  WINDOWS_FLEET_ID_PRODUCTION: "NjQyNTg2NXxOR0VQfEZMRUVUfDAxOWMyMjljLWVhYTQtN2JmNi04YWIyLTU2ZWI0YjE2ZWZjZQ"
```
DavSanchez: I had included these 4 as repository variables earlier. Check https://github.com/newrelic/newrelic-agent-control/settings/variables/actions we might leverage these to have a single source of truth and even build the specs below from them.

alvarocabanas: Changed in last commit.

.github/workflows/scheduled_onhost_canary_config_rotation.yml
```diff
@@ -0,0 +1,132 @@
...
+  #   Even runs (run_number % 2 == 0) -> NRInfra only        (remove)
+  # ---------------------------------------------------------------------------
+  # AC US Production Account `canaries-linux-flex-config` config
+  FLEX_SPEC_LINUX_PRODUCTION: "nri-flex:$CANARY_NRI_FLEX_VERSION:NjQyNTg2NXxOR0VQfEFHRU5UX0NPTkZJR1VSQVRJT05fVkVSU0lPTnwwMWEwMTU3ZC0yOGYwLTc5ZDgtODkzZi1mNWUwMzY0ZWFkNzA"
+  # AC US Production Account `canaries-windows-flex-config` config
+  FLEX_SPEC_WINDOWS_PRODUCTION: "nri-flex:$CANARY_NRI_FLEX_VERSION:NjQyNTg2NXxOR0VQfEFHRU5UX0NPTkZJR1VSQVRJT05fVkVSU0lPTnwwMWEwMTU3YS1iZWFlLTdiZWMtYTI1NC0wNDQyNjdlZmNlN2Q"
+  # AC Staging Account `canaries-linux-flex-config` config
+  FLEX_SPEC_LINUX_STAGING: "nri-flex:$CANARY_NRI_FLEX_VERSION:MTIyMTMwNjh8TkdFUHxBR0VOVF9DT05GSUdVUkFUSU9OX1ZFUlNJT058MDE5ZmViZjItNjJjNi03MDU4LTk2OTAtZWFlYWYxMTcyYzI2"
+  # AC Staging Account `canaries-windows-flex-config` config
+  FLEX_SPEC_WINDOWS_STAGING: "nri-flex:$CANARY_NRI_FLEX_VERSION:MTIyMTMwNjh8TkdFUHxBR0VOVF9DT05GSUdVUkFUSU9OX1ZFUlNJT058MDE5ZmVjMDAtYmU3Ni03NWEyLWExODMtM2EwNGJiYTRjOTgx"
```
DavSanchez, line 56: These ones can interpolate the `vars` for the Fleet ID segment as well so we don't forget to change them no?

alvarocabanas: No, those are config vsrsions, I changed the comments in las commit.

DavSanchez: Ahh sorry for the confusion!

DavSanchez (approved):

PR #2689, opened 2026-07-13
feat: add nri-docker default config

alvarocabanas:
Add the default linux config for nri-docker.


danielorihuela (approved):
Looks good to me!!!

PR #2691, opened 2026-07-13
feat: cleanup shared-fs on agent removal [NR-579376]

sigilioso:
## Summary

This PR makes the on-host resource cleaner reclaim an agent's shared-filesystem entries when it is removed.

Cleanup lives in `OnHostCleaner` (the on-host resource cleaner), fired at the same points as the existing per-agent cleanup. Shared paths are **static** (derived from the Agent Type definition, not rendered) so they are resolved straight from the registry, no assembly needed.

> [!NOTE]
> #2679 Adds the shared-fs validation, the clean-up assumes validated (collision-free) files.

## Changes

- **Agent _uninstall_**:  `ResourceCleaner::clean(agent_id, agent_type)` per removed agent was already called. On host, `clean` now also deletes that agent's declared shared paths (individually-owned files and whole
  `dir_content_from_map` directories). Co-owned directories are never declared, so they are left in place. This is safe because `clean` runs only on the validated remote-config apply path, where single ownership of shared paths is guaranteed.
- **Stale agent removal on startup**: missing agents' type is no longer known, so it can't be targeted individually. At startup a whole-set reconcile walks the shared base and deletes every file not owned by a currently-configured agent.

## Out of scope

- Migration from `std::fs` to namespaced `fs` abstraction: the `OnHostResourceCleaner` was already using (and still uses) some function outside the `fs` abstraction. We can tackle this as a different refactor that should update both the resource cleaner and `filesystem/rendered.rs`. We would need to extend the current abstractions and update both codebases.


agent-control/src/agent_control/resource_cleaner/on_host.rs
```diff
@@ -137,25 +157,131 @@ where
...
+            if owned_managed_dirs.contains(&path) {
+                continue;
+            }
+            reconcile_shared_dir(&path, expected_files, owned_managed_dirs)?;
+        } else if !expected_files.contains(&path) {
+            remove_path(&path)?;
+        }
+    }
+    Ok(())
+}
```
alvarocabanas, line 264: If I have a shared folder declared as:

```
  shared_filesystem:
    # OHIs config folders.
    # infra-agent is configured through NRIA_PLUGIN_DIR, to read configs from this folder.
    infra-agent-ohi-configs:
      kind: dir
      persistent: true
      entries:
        my-custom:
          kind: dir
          entries:
            my-file:
              kind:file
              text: "my-text"
```

And we don't have the `my-custom` dir in any other agent-type, and we removed this agent, the folder my-custom will keep existing, isn't it, and only the file `my-file` will be deleted, not? We always keep the directories, even if they are subdirectories not owned by anybody else. Wouldn't it be possible to delete these sub-directories?

sigilioso: It should be possible, but it is not easy. 

We could delete directories y they are empty but this could lead to inconsistencies if some agents declare empty directories.

Another option would be to check if the directory is still declared in any of the remaining agents still points to the directory. However, `cleaner.clean(agentID)` doesn't receive the new list of agents. Maybe we need to modify that trait and the code calling it 🤔

sigilioso: As discussed _offline_, it makes sense to modify the `ResourceCleaner` trait to provide both, the removed agents and the agents that keep running.

Since the consequent refactor is quite a bit wide and keeping the directories for now seems harmless. I'll take it in a follow-up PR to avoid making this change harder to review.

agent-control/src/agent_control/resource_cleaner/on_host.rs
```diff
@@ -137,25 +157,131 @@ where
...
+/// Deletes a file or a whole directory subtree at `path`. A missing path is not an error.
+fn remove_path(path: &Path) -> io::Result<()> {
+    if !path.exists() {
+        return Ok(());
+    }
+    if path.is_dir() {
+        std::fs::remove_dir_all(path)
+    } else {
+        std::fs::remove_file(path)
+    }
```
danielorihuela, line 236: There's something wrong in the logic. This method removes directories, but it's only called from one place, right? `reconcile_shared_dir`. Inside this method, we only call remove paths for files, never for directories.

Do we have a test with that situation? I'm thinking that an OHI could create a `dir` in a shared folder, and it would be left behind. That `dir` should be removed. I know there's not a use case for that, but the behaviour is strange (if what I'm mentioning is correct).

danielorihuela: I mean


```yaml
shared-filesystem:
  ohis-configs: # shared
    kind: dir
    persistent: true
    entries:
      private-folder: # not shared
        kind: dir
        persistent: false
        entries:
          ...
```

danielorihuela: Maybe this is removed as part of the `ephemeral` thing? I'm confused right now.

agent-control/src/agent_control/resource_cleaner/on_host.rs
```diff
@@ -512,4 +722,96 @@ mod tests {
...
+        let shared = tmp.path().to_path_buf();
+        std::fs::create_dir_all(shared.join("ohi-configs")).unwrap();
+        std::fs::write(shared.join("ohi-configs").join("nri-redis.yaml"), "redis").unwrap();
+        std::fs::write(shared.join("ohi-configs").join("nri-other.yaml"), "other").unwrap();
+        std::fs::create_dir_all(shared.join("additional-files")).unwrap();
+        std::fs::write(shared.join("additional-files").join("app.log"), "log").unwrap();
+
+        let definition = host_type_with_shared(serde_json::json!({
+            "ohi-configs": {
+                "kind": "dir",
```
danielorihuela, line 738: Shouldn't we use `persistent: true`? Isn't persistency respected yet in the shared file system?

agent-control/src/agent_control/resource_cleaner/on_host.rs
sigilioso: I think it is the same issue that @alvarocabanas pointed out here: https://github.com/newrelic/newrelic-agent-control/pull/2691#discussion_r3572229441. Directories are kept because they can be shared and we don't have enough information to know if they are referenced from other agent when we perform `clean`. 

I makes sense to control it and avoid leaving empty directories behind. But, as stated in that conversation, it involves an additional refactor and I'm covering it in a follow-up PR.

danielorihuela: I don't think it's the same.

He is complaining about the fact that we are not cleaning shared folders (something we left out of scope, but we can solve it if it's doable with the current implementation) if everything is removed inside it.

What I'm complaining about is the fact that... wait, yeah. It's not the same, but it's caused because of the same decision.

agent-control/src/agent_control/resource_cleaner/on_host.rs
sigilioso: This is not handled in the ResourceCleaner (the resource cleaner deletes regardless of the `persistent` value for both the 'shared-fs' and the 'per-agent-fs') but it is actually very good point. We need to support that for the shared-fs as well.

danielorihuela: This is kind of related with the other issue I guess. I would add `persistent: true` here just in case, but maybe it's not needed.

danielorihuela: I'm not sure if we should really support it. Like,

* config files must be removed regardless of the persistency mode, I guess
* directories also don't care about persistency, we should remove them as soon as their are empty and not declared by any running agent (or leave them behind)

I'm not sure `persistent` makes sense in the shared file system now that I think of it.

danielorihuela (approved):

alvarocabanas (approved):
LGTM

PR #2692, opened 2026-07-14
feat: oci-utils for testing

achow-nr:
Add oci-utils CLI for pushing agent packages and agent type yaml
[NR-578592](https://new-relic.atlassian.net/browse/NR-578592)

[NR-578592]: https://new-relic.atlassian.net/browse/NR-578592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/crates/oci-test-utils/README.md
+++ b/test/crates/oci-test-utils/README.md
@@ -1,0 +1,133 @@
+# oci-test-utils
+
+In-tree test utilities for pushing **agent packages** and **agent type definition** artifacts to an OCI registry — local (e.g. a `zot`/`registry:2` on `localhost:5001`) or remote (Docker Hub, GHCR).
+
+It is **not** production tooling. Distribution is source-only (the crate has `publish = false`); there are no published binaries. Run via `cargo run` or vendor the library into integration tests.
+
+## Two ways to use it
+
+### As a library (integration tests)
+
+`PackagePublisher` is the workhorse. Tests construct one and push artifacts as part of fixture setup:
+
+```rust
+use oci_test_utils::{LOCAL_HTTP_REGISTRY_URL, PackageMediaType, PackagePublisher};
+
+let publisher = PackagePublisher::new(runtime.handle().clone(), LOCAL_HTTP_REGISTRY_URL);
+let reference = publisher.push_with_tag(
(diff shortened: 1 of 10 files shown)


test/crates/oci-test-utils/README.md
```diff
@@ -0,0 +1,133 @@
...
+);
+```
+
+The library exposes:
+
+- `PackagePublisher` — publishes both agent packages and agent type artifacts. Supports an optional repository name (default `"test"`) and either basic or bearer-token auth.
+- `AgentTypeDefinitionMeta` — minimal agent type metadata + the `<environment-prefix>-<name>-<version>` tag composition rule mirrored from `agent-control/src/agent_type/oci.rs`.
+- `OCISigner` — generates ephemeral signing keys and produces cosign-style signatures (used by tests exercising AC's signature-verification path).
+
+### As a standalone binary (`oci-utils`)
```
gsanchezgavier, line 30: cool , simple ux !

test/crates/oci-test-utils/README.md
```diff
@@ -0,0 +1,133 @@
...
+echo "$GHCR_TOKEN" | cargo run -p oci-test-utils --bin oci-utils -- \
+    --registry ghcr.io --repository myorg/nr-agent \
+    --token-stdin \
+    push-agent-type ./agent-type.yaml
+```
+
+## Notes for maintainers
+
+- `agent_type_meta.rs` is a deliberate, hand-maintained mirror of `agent-control/src/agent_type/oci.rs::AgentTypeTag` and its environment-prefix helper. The duplication is intentional: this crate cannot depend on `newrelic-agent-control`. Keep the two in sync by hand when either side changes.
+- `flate2` and `tar` are pinned to the same versions agent-control uses so the workspace resolves to one copy each.
```
gsanchezgavier: nit: we could move these out of AC to the workspace level and update both AC and this to use them i think

gsanchezgavier (dismissed):
Awesome Thanks! I just left a nit

gsanchezgavier (approved):
Thanks for taking care of this!

PR #2693, opened 2026-07-14
refactor: k8s scenarios tests custom agent type

danielorihuela:
Remove custom agent types data files from k8s scenarios tests by using a builder, like for onhost tests.
Don't review as a whole, **go commit by commit**.

### Context

I noticed that onhost scenarios tests had a builder for custom agent types, but k8s didn't have one. So I added it. Not sure if it's better, give me feedback. At the very least, we now are using the same mechanism that for onhost.

These are the changes:

* custom agent types data files are removed
* added `K8sCustomAgentTypeBuilder`
* added `CommonCustomAgentTypeBuilder`
* renamed onhost builder to `OnHostCustomAgentTypeBuilder`
* renamed `build` to `write`

There's probably more room for improvement, but this is enough for this PR.


agent-control/tests/common/custom_agent_type.rs
```diff
@@ -0,0 +1,55 @@
...
+            ..self
+        }
+    }
+
+    pub fn with_agent_type_id(self, agent_type_id: &str) -> Self {
+        Self {
+            agent_type_id: AgentTypeID::try_from(agent_type_id).unwrap(),
+            ..self
+        }
+    }
```
sigilioso, line 33: Nit: why do we have `new` receiving an AgentTypeID and also this `with_agent_type_id` method? Isn't the `new` enough?

agent-control/tests/k8s/tools/custom_agent_type.rs
```diff
@@ -0,0 +1,233 @@
...
+        content.insert("variables".into(), variables);
+        content.insert("deployment".into(), deployment.into());
+        let content = serde_json::Value::from(content);
+
+        write!(f, "{}", serde_saphyr::to_string(&content).unwrap())
+    }
+}
+
+impl K8sCustomAgentTypeBuilder {
+    pub fn new() -> Self {
```
sigilioso: Nit: Considering that `new()` has only a _default_  `AgentTypeID`, whereas `default` has that `AgentTypeID` and default health/objects. Would it make sense to rename `new` to `empty`?

sigilioso (dismissed):
I like the approach, now k8s and on-host integration tests are aligned. On top of that, the builder approach is used for both the config and the custom agent types.

I've left two cosmetic comments. Feel free to ignore them!

agent-control/tests/k8s/tools/custom_agent_type.rs
danielorihuela: https://github.com/newrelic/newrelic-agent-control/pull/2693/changes/bb71910b19fb60cffb8229e8c5526c0a1a2bab26

agent-control/tests/common/custom_agent_type.rs
danielorihuela: I'll take a look on a different PR. That's one of the things I left out of scope of this PR cause it was already huge.

sigilioso (approved):

PR #2694, opened 2026-07-14
feat: handle on_agent_type_changed for a subagent

alvarocabanas:
Adds `on_agent_type_changed` to the `ResourceCleaner` trait, called when an agent's type is bumped during a live  
recreate_sub_agent cycle. 

**On Kubernetes**, it calls `collect(agent_id, old_type)` to garbage-collect k8s objects annotated with the old type — only those objects, since collect in Collect mode requires an exact annotation match, leaving new-type resources untouched. 
**On-host,** it reconciles both the per-agent and shared filesystems:
- Paths declared by the old type that no longer appear in any active agent's declarations are deleted, while    
  the union of all active agents' shared paths acts as a "keep" set to protect co-owned entries. 
- The existing  `reconcile_shared_dir` startup sweep is kept distinct, as it walks the actual filesystem to catch undeclared    
  runtime files — a broader operation only safe when all agents are quiesced. 

An integration test in `shared_filesystem.rs` covers the full end-to-end scenario: a type bump deletes stale per-agent and shared  entries, writes new ones, and preserves a shared path co-owned by a second active agent.


agent-control/src/agent_control/resource_cleaner/on_host.rs
```diff
@@ -112,6 +112,18 @@ where
...
+    /// This method only reclaims resources for agents that are **absent** from `configured`. If
+    /// an agent is still present but its type was changed in the local config while Agent Control
+    /// was stopped, the old type's per-agent filesystem entries are **not** cleaned up here:
+    /// the previous agent type is not persisted anywhere, so there is no way to know which paths
+    /// to remove. [`on_agent_type_changed`](super::ResourceCleaner::on_agent_type_changed) handles
+    /// type-change reconciliation for live remote config updates (where both old and new types are
+    /// known), but it is never called for local config changes made offline. Old per-agent entries
+    /// from such type bumps are an accepted side-effect; they are eventually reclaimed if the agent
+    /// is removed from the fleet entirely.
     pub fn cleanup_stale_agents(&self, configured: &SubAgentsMap) {
```
sigilioso: Is the comment in the right? The method do reclaim resources **absent** from configured. Therefore, in the described situation (agent-type is changed in local configuration) it will delete any file that was present in the old agent-type but is absent from the new agent type as it compares what was written in the filesystem with what is supposed to be written according to the new set of agents.

agent-control/src/agent_control/resource_cleaner/on_host.rs
```diff
@@ -174,10 +186,67 @@ where
...
+        let new_agent = self.declared_agent_filesystem_paths(agent_id, new_type)?;
+        delete_stale_paths(&old_agent, &new_agent)
+            .map_err(|(path, source)| OnHostCleanerError::AgentFilesystem { path, source })?;
+
+        let old_shared = self.declared_shared_paths(old_type)?;
+        let active_shared = self.declared_shared_paths_from_all_agents(active_agents)?;
+        delete_stale_paths(&old_shared, &active_shared)
+            .map_err(|(path, source)| OnHostCleanerError::SharedFilesystem { path, source })?;
+
+        Ok(())
```
sigilioso: Nit: would it make sense that the helpers `declared_agent_filesystem_paths` and `declared_shared_pahts` received the actual `AgentTypeDefinition` instead of the agent_type `id`? We would avoid looking up the same Agent Type in the registry twice.

agent-control/src/agent_control/resource_cleaner/on_host.rs
```diff
@@ -814,4 +963,238 @@ mod tests {
...
+
+        assert!(
+            !shared.join("persistent.yaml").exists(),
+            "persistent file absent in new type must be deleted"
+        );
+        assert!(
+            !shared.join("ephemeral.yaml").exists(),
+            "ephemeral file absent in new type must be deleted"
+        );
+    }
```
sigilioso: Nit: I'd consider joining these two tests together (with entries with `persistent: true` and `persistent: false`) both show that the method works regardless of the _persistence_ status of the entry.

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -204,11 +213,11 @@ impl SharedFileSystem {
...
 /// Recursively gathers the paths declared by `entry` (rooted at `path`) into `declared`.
-fn collect_declared_paths(
-    path: &Path,
-    entry: &FilesystemEntry,
-    declared: &mut DeclaredSharedPaths,
-) {
+///
+/// `Dir` entries are treated as co-owned drop zones: the directory path itself is not tracked,
+/// only the declared files inside are. This allows multiple agents to write into the same
+/// directory without one agent's cleanup removing another agent's files or the directory itself.
```
sigilioso, line 219: So we leave directories out of scope for now and we address all directories removal (including 'on agent type change') on the PR I'm working on, is that right?

sigilioso (commented):
Thanks for addressing this! I've left some comments/questions but I'd say it is almost ready.

agent-control/src/agent_control/resource_cleaner/on_host.rs
alvarocabanas: I've modified this in the last commits.

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -135,12 +144,44 @@ impl FileSystem {
...
+
+    /// Deletes the on-disk path of every entry in the filesystem that is not declared and is not a
+    /// child of a persistent declared directory.
+    ///
+    /// # Known limitation — previously-declared entries inside persistent directories
+    ///
+    /// Persistent directories are skipped entirely so that runtime files the agent wrote (and
+    /// never declared) are preserved. The unavoidable side-effect is that a file which *was*
+    /// declared inside a persistent directory in a previous type version but is no longer declared
+    /// in the current one also survives. We would need to safe the state to handle this case
```
sigilioso: I think it is clearer now! 👍 
```suggestion

    /// Deletes the on-disk path of every entry in the filesystem that is not declared and is not a
    /// child of a persistent declared directory.
    ///
    /// # Known limitation — previously-declared entries inside persistent directories
    ///
    /// Persistent directories are skipped entirely so that runtime files the agent wrote (and
    /// never declared) are preserved. The unavoidable side-effect is that a file which *was*
    /// declared inside a persistent directory in a previous type version but is no longer declared
    /// in the current one also survives. We would need to save the state to handle this case
```

agent-control/src/agent_control/resource_cleaner/on_host.rs
alvarocabanas: I've refactored in the last commit,  still the definitions need to be computed before, but the code has less repetition with this change and a helper.

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
alvarocabanas: Yes, the change here is, a part from the comment, the fact that now there is the generic DeclaredPaths instead of DeclaredSharedPaths since there are helpers working for both types of Filesystems, the logic here is unaltered.

agent-control/src/agent_control/resource_cleaner/on_host.rs
alvarocabanas: Done in last commits

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -167,6 +208,67 @@ impl SharedFileSystem {
...
+                    persistent: true, ..
+                } => {}
+                RenderedEntry::Dir { children, .. } => {
+                    let child_refs: HashMap<PathBuf, &RenderedEntry> =
+                        children.iter().map(|(k, v)| (k.clone(), v)).collect();
+                    prune_undeclared(&abs, &child_refs, file_ops, dir_manager)?;
+                }
+                RenderedEntry::DirContentFromMap { files } => {
+                    prune_undeclared_in_map_dir(&abs, files, file_ops, dir_manager)?;
+                }
```
sigilioso: The content of the directory declared as `dir_content_from_map` is completely built from variable values. If we _prune_ the directory on spin-up only we can easily leave behind old content. Example:

```yaml
filesystem:
  logging.d:
    kind: dir_content_from_map
    source: ${nr-var:logging_config}
```

When the agent spins-up this _prune_ method is called, but it is not called when a remote configuration is received. Therefore if we received:

```yaml
config_logging:
  file-1.conf: "some config"
```

And later

```yaml
config_logging:
  file-2.conf: "some other config"
```

We'll keep both `file-1.conf` and `file-2.conf` because the first is not pruned.

In order to fix this, I think we **shouldn't** prune `dir_content_from_map` and, instead compute the directory completely Every time we write it in the corresponding [write](https://github.com/newrelic/newrelic-agent-control/blob/999212acc5d2c2b087a1fa9715bd9dd322c990fb/agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs#L85-L91).

```rust
            // ...
            Self::DirContentFromMap { files, .. } => {
                ensure_dir(dir_manager, path)?;
                // We should also remove any previous content on that dir before writing anything
                for (file_name, content) in files {
                    write_file(file_ops, dir_manager, &path.join(file_name), content)?;
                }
                Ok(())
            }

```

alvarocabanas: I've modified it in the last commit.

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
sigilioso: The comment is outdated now, isn't it?

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -240,35 +245,14 @@ fn prune_undeclared(
                         children.iter().map(|(k, v)| (k.clone(), v)).collect();
                     prune_undeclared(&abs, &child_refs, file_ops, dir_manager)?;
                 }
-                RenderedEntry::DirContentFromMap { files } => {
-                    prune_undeclared_in_map_dir(&abs, files, file_ops, dir_manager)?;
-                }
+                // Skip. write() deletes and recreates this directory from the current map.
+                RenderedEntry::DirContentFromMap { .. } => {}
```
sigilioso: very nit: I'd move this to right below the `RenderedEntry::File` because we do the very same for `File` and `DirContentFromMap`

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
alvarocabanas: Modified in last commit

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
alvarocabanas: Modified in last commit

sigilioso (approved):
LGTM

PR #2696, opened 2026-07-15
feat: shared-fs entries are always persistent [NR-579376]

sigilioso:
This PR adds validation to assure that `persistent: false` cannot be set to `shared-fs` entries.

By default unknown fields are ignored and, considering the per-agent filesystem implementation where the _persistence_ can be configured, it can be misleading. Therefore validation has been added by implementing custom deserialization for SharedFilesystem in order to assure that the `persistent` field cannot be set.

The corresponding Agent Types have been updated as well.


agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -73,18 +73,22 @@ pub enum FilesystemEntry {
...
         /// The directory's child entries.
         #[serde(default)]
         entries: HashMap<SafePath, FilesystemEntry>,
-        /// The persistency attribute marking its lifecycle.
+        /// The persistency attribute marking its lifecycle. `None` (omitted) resolves to the
+        /// default (ephemeral) for the per-agent filesystem; the shared filesystem is always
+        /// persistent and rejects an explicit `persistent: false`.
         #[serde(default)]
-        persistent: bool,
+        persistent: Option<bool>,
```
danielorihuela, line 91: I'm not sure I like this change. Isn't it hard to reason about? And also hard to maintain?

### Reasoning

Basically, `filesystem` has a persistence option that is honored. `shared_filesystem` works differently, which is okay, but then why can we put `persistence: true`, but not `persistence: false`? Ah wait, because if we don't add it, then it's ignored? Does this work the same way on the `filesystem`? Not, because there we transform the empty case to the default case.

filesystem + empty -> false
filesystem + persistence false -> false
filesystem + persistence true -> true

shared + empty -> empty
shared + persistence false -> error
shared + persistence true -> true

Wouldn't it be easier just to not allow `persistence` on the shared filesystem. To make it clear that it doesn't make any sense the concept there. Files are just removed when agent is removed.

Not sure how this is pieced together with changes. Into the next section!

### Maintain

With the current model, what happens if we want to start managing `persistence: false`?

It's backwards compatible, but the behaviour changes. Suddenly, the same agent type works different with a newer AC version. Now, files that don't declare `persistence: true` will go from be handle as persistent files to be handle as ephemeral. I personally don't like that, but to be honest, I don't think users will care. At least for OHIs, we don't need that feature.

If we didn't allow `persistent` at all, we would have the same problem. In that case, it will be even worse because we would force all agent types to create a new version declaring folders as `persistent: true` to avoid removing them. Mmm, this also happens with the current implementation. We have to remember to put `persistence: true` to avoid the issue.

### Conclusion

I'm not sure what's best haha. Maybe we can keep your approach but force all shared folders to have `persistence: true` in case we want to change the behaviour in the future? I think that would be the best, even though it's strange from the agent type author point of view.

Any feedback? I'm just throwing out thoughts.

sigilioso: > shared + empty -> empty
> shared + persistence false -> error
> shared + persistence true -> true

Actually, I'd say that empty means true for shared-fs. Therefore:

```
filesystem + empty -> false
filesystem + persistence false -> false
filesystem + persistence true -> true

shared + empty -> true
shared + persistence false -> error
shared + persistence true -> true
```

> Wouldn't it be easier just to not allow persistence on the shared filesystem. To make it clear that it doesn't make any sense the concept there. Files are just removed when agent is removed.

In fact, that's a really good point, it would be easier to reason about it. We can also roll back the changes (keeping the docs) and just ignore the field (as we ignore any other _unknown_ field).

> [...] It's backwards compatible, but the behaviour changes.  [...]

If the behavior changes, it is not backwards compatible 🙃. I'd say that would be wrong.

If we needed to start handling `persistent: false` in the future we can do it (and keep it backwards compatible), but we **cannot change the default behavior** because that would be a breaking change. For `shared-fs` the default would be `persistent: true`.

danielorihuela: > Actually, I'd say that empty means true for shared-fs.

True. I got confused.

> If the behavior changes, it is not backwards compatible 🙃. I'd say that would be wrong.

True. I was thinking that the system will keep running, but with unintended consequences.

> In fact, that's a really good point, it would be easier to reason about it. We can also roll back the changes (keeping the docs) and just ignore the field (as we ignore any other unknown field).

I'm not sure I like that either :D
It's easy for the agent type author to miss that detail.

> but we cannot change the default behavior because that would be a breaking change

I see. You are right. I think this is what bothers me. I would like not to have that difference. We could avoid that forcing the agent type author to put `persistent: true` everywhere, but I don't like that either, because the feature might never come. In fact, it could be solved with alternative ways. For example, with a backup mechanism instead of declaring it `persistent`.

Anyway, I think I'm more inclined now to just disallow `persistent` field at all in `shared_filesystem`. It doesn't make any sense there. Everything has to be "persistent". We can't allow it to be false, and adding it's true counterpart feels useless.

Ignoring seems a good approach. It can be a bit confusing though. An author can think that both filesystems have the `persistent` and assume they work the same, to later realize they don't work the same.

sigilioso: Ignoring can be confusing because almost everything else behaves the same in the per-agent filesystem and shared filesystem. However, we have already taken that approach for one _kind_ in the filesystem. `dir_content_from_map` doesn't declare the `persistent` field

https://github.com/newrelic/newrelic-agent-control/blob/27b0319ee6cfd45894688f90dfa4c809ed26b488/agent-control/src/agent_type/runtime_config/on_host/filesystem.rs#L93-L98

And therefore, setting it up is just ignored. We _cover_ that with documentation. 

https://github.com/newrelic/newrelic-agent-control/blob/27b0319ee6cfd45894688f90dfa4c809ed26b488/docs/INTEGRATING_AGENTS.md#L523

Maybe we should do the same here 🤔

danielorihuela: I still hope we can remove `dir_content_from_map` some day 🙏.

I think it's a different case. For `dir_content_from_map`, we ignore it so that the behaviour is the same within the `filesystem`. In other words, why would a `dir` allow `persistent` and `dir_content_from_map` throw an error? I guess we could do that too, but the behaviour is not uniform. Not that ignoring it makes it uniform all of a suden.

For `shared_filesystem` we have the opportunity to always error if `persistent` is added, regardless of the entry type.

I don't know. As long as we avoid asking ourselves why the shared filesystem only allows `persistent: true` but fails on `persistent: false` (something that seems not useful), I'm okay. I guess ignore + docs + removing `persistent` from all OHIs and infra should be enough to avoid confusion. The validation might not add anything at the end, other than more code to maintain. We can always implement it later if we see that's needed.

sigilioso: Fair enough! 👍  I'm updating the PR (and the corresponding usages in the existing Agent Types) to disallow the `persistent` field. I'm not found of the `Option<bool>` I've introduced... but it definitely represents the required absence of the `persistent` field.

danielorihuela: If no one else says anything, maybe we can "quickly" ask in the refinement if anyone has any concern.

sigilioso: I've just updated the approach to disallow the `persistent` field for now. If we change our minds after tomorrow's refinement, we'll revisit it again.

sigilioso:
I'm closing this PR as we will finally remove the `persistent` field completely.

PR #2697, opened 2026-07-15
feat: add binary size comparasion to the pipelines

vjripoll:
Adds a binary size regression check to the release pipelines, so an unexpected size increase in the release binaries.

- `measure-release-binaries.sh` extracts binary sizes from the published archives (version-agnostic keys, so the same binary matches across releases).
- `check-binary-size.sh` compares current vs. baseline, prints a table to the job summary, and fails if any binary grows past the threshold. New binaries are reported, not failed.
- Baselines are stored via `actions/cache`, scoped to the branch that wrote them (plus its base/default branch) that's why the PR baseline is only published from `main`, so every branch shares the same reference.

**PR**
---
image
image
image

**NIGHTLY**
---
image
image
image

**RELEASE**
---
image

### DIFF
--- a/.github/actions/binary-size-baseline/action.yml
+++ b/.github/actions/binary-size-baseline/action.yml
@@ -1,0 +1,58 @@
+name: 'Binary size baseline check'
+description: 'Restores a cached binary-size baseline, checks against it, and optionally republishes it'
+inputs:
+  current-file:
+    description: 'Path to the current sizes file (name size-in-bytes per line)'
+    required: false
+    default: 'binary-size/current.txt'
+  baseline-file:
+    description: 'Path to write/read the baseline sizes file'
+    required: false
+    default: 'binary-size/baseline.txt'
+  cache-key:
+    description: 'Exact cache key to save the baseline under'
+    required: true
+  cache-restore-prefix:
+    description: 'Cache key prefix used to restore the latest baseline (restore-keys)'
+    required: true
+  threshold:
+    description: 'Max allowed percent growth over baseline before failing'
+    required: false
+    default: '10'
+  enforce:
+    description: "If 'false', never fail the job on a regression (still produces the summary table)"
+    required: false
+    default: 'true'
+  publish-baseline:
+    description: "If 'true', republish current sizes as the new baseline (only if the check passed, when enforced)"
+    required: false
+    default: 'false'
+runs:
+  using: 'composite'
+  steps:
+    - name: Restore binary size baseline
+      uses: actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5
+      with:
+        path: ${{ inputs.baseline-file }}
+        key: ${{ inputs.cache-key }}
+        restore-keys: |
+          ${{ inputs.cache-restore-prefix }}
+
+    - name: Check binary size against baseline
+      id: check
+      shell: bash
+      continue-on-error: ${{ inputs.enforce == 'false' }}
+      run: bash .github/scripts/binary-size.sh check "${{ inputs.current-file }}" "${{ inputs.baseline-file }}" "${{ inputs.threshold }}"
+
+    # Always publish when not enforcing (e.g. main); otherwise only if the check passed.
+    - name: Publish new baseline
+      if: ${{ inputs.publish-baseline == 'true' && (inputs.enforce == 'false' || steps.check.outcome == 'success') }}
+      shell: bash
+      run: cp "${{ inputs.current-file }}" "${{ inputs.baseline-file }}"
+
+    - name: Save binary size baseline
+      if: ${{ inputs.publish-baseline == 'true' && (inputs.enforce == 'false' || steps.check.outcome == 'success') }}
+      uses: actions/cache/save@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5
+      with:
+        path: ${{ inputs.baseline-file }}
+        key: ${{ inputs.cache-key }}
--- a/.github/scripts/binary-size.sh
+++ b/.github/scripts/binary-size.sh
@@ -1,0 +1,173 @@
+#!/usr/bin/env bash
+# Binary size tooling for CI. Subcommands:
+#   measure        <archives-dir> <output-file>  # extract sizes from packaged tar.gz/zip (nightly/prerelease)
+#   measure-local  <output-file> <name=path>...  # stat already-built local binaries (PR check)
+#   check          <current-file> <baseline-file> [threshold-percent, default 10]
+set -euo pipefail
+
+# Binaries packaged by .goreleaser.yml; keep in sync by hand with Cargo.toml's [[bin]] entries.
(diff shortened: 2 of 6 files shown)


.github/workflows/push_pr_checks_tests.yml
```diff
@@ -183,6 +183,22 @@ jobs:
...
+      # Enforced on PRs/merge queue; on main it only republishes the baseline for future PRs.
+      - name: Binary size baseline check
+        uses: ./.github/actions/binary-size-baseline
+        with:
+          cache-key: binary-size-baseline-${{ github.sha }}
+          cache-restore-prefix: binary-size-baseline-
+          enforce: ${{ github.ref != 'refs/heads/main' }}
+          publish-baseline: ${{ github.ref == 'refs/heads/main' }}
+
   unit-docs-onhost-integration-tests:
```
sigilioso, line 202: If we are measuring the release binary size on PRs, would it make sense to also check it?

If possible, I'd try to avoid adding new checks that are only detected on nightly and pre-release pipelines because they can cause noise and make the root cause of the binary-size increase harder to diagnose.

If we fail only on nightly and release, on busy days with many changes merged, maybe including some dependency bumps, we need to find out what was the root cause once the issue is detected in the nightly. But, if we could measure it in the PR itself, the root cause would be likely the change introduced in the PR.

sigilioso (commented):
I haven't reviewed the action/script yet. I'd like us to discuss **when** the binary size is checked first. I've just left a comment related to that. Let me know what you think.

.github/workflows/push_pr_checks_tests.yml
vjripoll: Right now, the PR is always checked against main.

I think having it in PRs, nightlies, and releases helps us track how the binary size grows and keeps us aware of the increase right now, it's PR against Main, Main against Main day after day, and a Release against the previous one. Maybe we couldn't set allow_fail: true, but at least we would be aware of the size.

sigilioso: I probably got confused by the script and teh action in the push-pr pipeline, I thought it wasn't checked.

-  `bash .github/scripts/binary-size.sh measure-local ...`: I guess this measures what we have just built
- `./.github/actions/binary-size-baseline`: does this perform the check?

I'll review the details!

.github/scripts/binary-size.sh
```diff
@@ -0,0 +1,173 @@
+#!/usr/bin/env bash
+# Binary size tooling for CI. Subcommands:
+#   measure        <archives-dir> <output-file>  # extract sizes from packaged tar.gz/zip (nightly/prerelease)
+#   measure-local  <output-file> <name=path>...  # stat already-built local binaries (PR check)
+#   check          <current-file> <baseline-file> [threshold-percent, default 10]
```
danielorihuela, line 5: Nit: Maybe `measure-packaged`, `measure-local` and `check-difference` is easier to read?

danielorihuela:
Where does the size summary end? Is it added like a comment in the PR? Maybe we could add this on a following PR and remove the fail? So that we have 100% control over if we want to merge something fast at the expense of size?

.github/workflows/push_pr_checks_tests.yml
```diff
@@ -183,6 +183,22 @@ jobs:
       - name: Build onhost in release mode
         run: cargo zigbuild --release --package newrelic_agent_control --bin newrelic-agent-control --target x86_64-unknown-linux-musl
 
+      - name: Measure release binary sizes
+        run: |
+          mkdir -p binary-size
+          bash .github/scripts/binary-size.sh measure-local binary-size/current.txt \
+            "newrelic-agent-control-k8s=target/x86_64-unknown-linux-musl/release/newrelic-agent-control-k8s" \
+            "newrelic-agent-control=target/x86_64-unknown-linux-musl/release/newrelic-agent-control"
```
danielorihuela, line 191: We should also take windows into account, shouldn't we?

.github/actions/binary-size-baseline/action.yml
```diff
@@ -0,0 +1,58 @@
...
+  baseline-file:
+    description: 'Path to write/read the baseline sizes file'
+    required: false
+    default: 'binary-size/baseline.txt'
+  cache-key:
+    description: 'Exact cache key to save the baseline under'
+    required: true
+  cache-restore-prefix:
+    description: 'Cache key prefix used to restore the latest baseline (restore-keys)'
+    required: true
```
danielorihuela, line 17: This could be simplified into a single input named `cache-key-prefix`. Then, add the `github.sha` here in the file directly.

.github/actions/binary-size-baseline/action.yml
```diff
@@ -0,0 +1,58 @@
...
+description: 'Restores a cached binary-size baseline, checks against it, and optionally republishes it'
+inputs:
+  current-file:
+    description: 'Path to the current sizes file (name size-in-bytes per line)'
+    required: false
+    default: 'binary-size/current.txt'
+  baseline-file:
+    description: 'Path to write/read the baseline sizes file'
+    required: false
+    default: 'binary-size/baseline.txt'
```
danielorihuela, line 11: These two can be constants right? They are never passed from outside.

.github/actions/binary-size-baseline/action.yml
```diff
@@ -0,0 +1,58 @@
...
+    description: 'Exact cache key to save the baseline under'
+    required: true
+  cache-restore-prefix:
+    description: 'Cache key prefix used to restore the latest baseline (restore-keys)'
+    required: true
+  threshold:
+    description: 'Max allowed percent growth over baseline before failing'
+    required: false
+    default: '10'
+  enforce:
```
danielorihuela, line 22: Nit: Can't this be named `continue-on-error`?

danielorihuela:
@vjripoll what we do is the following:

* Compare PR binary size against main binary size
* Compare nightly vs the previous nightly
* Compare prelease vs last release

I wonder what we really need. I believe the prerelease comparison is needed, a 100%. I wouldn't fail at that point. The comparison on each PR is also valuable. We can quickly see if what we did increases the size a lot, and try to amend it if needed.
I don't like the check on the nightlies. I don't think it helps us in any way. We've already seen the increase for each PR, the nightly will show the sum of all the increases of that day. At that point we have nothing to do. If we already accepted the increase, we will accept it here too.

That's my opinion.

.github/workflows/nightly.yml
```diff
@@ -25,6 +25,40 @@ jobs:
...
+            exit 1
+          fi
+
+      # Only advances the baseline if the check passed, so a regression isn't baked in.
+      - name: Binary size baseline check
+        uses: ./.github/actions/binary-size-baseline
+        with:
+          cache-key: binary-size-nightly-${{ github.run_id }}
+          cache-restore-prefix: binary-size-nightly-
+          publish-baseline: 'true'
```
gsanchezgavier, line 60: is it reliable to use this cache? i think it gets wiped everyday, once the max amount of cache has been used , which we were reaching it everyday (at least in the past)

PR #2698, opened 2026-07-15
feat: agent control self-instrumentation OTLP pipeline foundation (NR-477218, NR-580189)

MunhozThiago:
PoC - Self Instrumentation

Productionizes the OTLP export pipeline itself: region-derived endpoint, correct HTTP client feature flags, config plumbing, and the two hardening fixes this pipeline needed. Deliberately excludes business metrics (record_* instruments) and a tracing-span expansion / id field fix for future implementations - this is the foundation those build on.

OTLP pipeline (NR-580189):
- Region-derived OTLP endpoint (US/EU/JP/staging) via OtelConfig::with_region_endpoint; port aligned to 4318 (HTTP/protobuf) end to end; no hardcoded staging endpoint
- reqwest-blocking-client (not reqwest-rustls/async): the PeriodicReader background thread uses futures_executor::block_on, which has no Tokio context - the async reqwest client would panic or deadlock from there
- Per-signal enable/disable (metrics, logs, traces) via self_instrumentation.opentelemetry config; traces default off
- Example configs and CONFIG.md updated; integration test exercises the full logs+metrics export path against a mock OTLP endpoint

Hardening (found in post-implementation review, folded in here since both depend directly on this pipeline):
- Bound OTLP force_flush() calls with a 5s timeout (instrumentation::flush) so a hung/unreachable endpoint can't stall graceful shutdown or a self-update, which previously had no bound beyond the 30s client timeout buried inside the SDK call
- Region-based endpoint derivation is now actually reachable: the ENDPOINT_SENTINEL fallback only ever triggered from test-only code because the `endpoint` config field had no serde default, so omitting it from a real config was a hard deserialization error instead of the intended region-derivation path

instrumentation::metrics is included here only as plumbing (register_provider + a bounded flush()) - the otel.rs pipeline needs it to flush metrics before self-update's exec(). The actual business `record_*` instruments land in the next PR on top of this one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Remember that the keywords that you can use are the following:
 ### enhancement
 - On-host and Kubernetes: when an agent's type is bumped via a live remote config update, Agent Control now reconciles both per-agent and shared filesystems, deleting paths declared by the old type that no longer appear in any active agent's declarations. Co-owned shared paths still claimed by another active agent are preserved. On Kubernetes, k8s objects annotated with the old type are garbage-collected while new-type resources are left intact.
 - Adds `oci-utils` CLI to `test/crates/oci-test-utils` for pushing agent packages and agent type definitions to OCI registries from the command line (dev/test tooling; not published).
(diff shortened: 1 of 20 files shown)


MunhozThiago:
Reopening to re-trigger CI checks after pushing fixes.

PR #2699, opened 2026-07-15
feat: redis agentType and e2e

paologallinaharbur:
This PR introduces the agentTypes for nri-redis for Windows and Linux.

Moreover, it adds an e2e for both linux and windows running a local instance of redis and of the infra agent+nri-redis

We check that metrics are delivered and that all the binaries and configurations are where they are expected to be


agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure.nri_redis-0.1.0.yaml
```diff
@@ -0,0 +1,45 @@
...
+    required: true
+  oci:
+    repository:
+      description: "OCI repository name for nri-redis"
+      type: string
+      required: false
+      default: newrelic/newrelic-agent-control-redis-dev
+      variants:
+        ac_config_field: "oci_repository_urls"
+        values: [ "newrelic/newrelic-agent-control-redis-dev" ]
```
sigilioso: We need to change the default when the package is available, right? Is it OK to merge it as it is?

test/e2e-runner/src/common/ohi.rs
```diff
@@ -0,0 +1,62 @@
...
+pub const SHARED_WINDOWS_FILESYSTEM_DIR: &str =
+    r"C:\ProgramData\New Relic\newrelic-agent-control\shared-filesystem";
+pub const EMBEDDED_LINUX_OHI_BINARIES: [&str; 3] = ["nri-flex", "nri-prometheus", "nri-docker"];
+pub const EMBEDDED_WINDOWS_OHI_BINARIES: [&str; 5] = [
+    "nri-flex.exe",
+    "nri-prometheus.exe",
+    "nr-winpkg.exe",
+    "nri-winservices.exe",
+    "windows_exporter.exe",
+];
```
sigilioso, line 15: Do we need to check these?

test/e2e-runner/src/linux/redis.rs
sigilioso, line 1: I was thinking about the possibility of filtering the infra-agent logs in order to simply check if the integration is actually being called with the expected parameters. If that was feasible we wouldn't need to set up any service (like this redis setup for Linux and Windows) and we could reuse the same approach to check any supported OHI.

That said, I don't know if fetching and filtering the infra-agent logs is something easy to do...

sigilioso (dismissed):
I've left some questions just in case we want to simplify it a little bit.

If we are OK with the complexity of setting up the service in the e2e and it is OK to merge the Agent Type with the testing package we can merge it as it is.

test/e2e-runner/src/common/ohi.rs
paologallinaharbur: We do not need to, but since @danielorihuela was checking that for linux I decided to check them for windows as well in the ohais

agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure.nri_redis-0.1.0.yaml
paologallinaharbur: They have just created `https://hub.docker.com/r/newrelic/nri-redis-artifacts` the test is overriding this anyway, I am changing it, but we should change the test to point ot it

test/e2e-runner/src/linux/redis.rs
paologallinaharbur: I think it was feasable, since we are already building the service for redis we can keep it, if it is flacky for any reason we can consider removing it

paologallinaharbur: but it is not a true e2e since it is not checking the binary "installation"

alvarocabanas (dismissed):
LGTM

agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure.nri_redis-0.1.0.yaml
```diff
@@ -0,0 +1,45 @@
...
+platform: host
+operating_system: linux
+protocol_version: "1.0"
+variables:
+  config_integration:
+    description: "nri-redis integration YAML consumed by the infra-agent"
+    type: yaml
+    required: true
+  version:
+    description: "OCI package version for nri-redis"
```
danielorihuela: Other agent types say `"Agent version"`. I don't know if we want to maintain any kind of consistency on this front.

I think it would make sense, given that `version` is outside `oci` in the variables.

paologallinaharbur: Done!

agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure.nri_redis-0.1.0.yaml
paologallinaharbur: I am creating a task for it

danielorihuela (dismissed):

sigilioso (dismissed):

agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure.nri_redis-0.1.0.yaml
```diff
@@ -0,0 +1,45 @@
...
+  version:
+    description: "Agent version"
+    type: string
+    required: true
+  oci:
+    repository:
+      description: "OCI repository name for nri-redis"
+      type: string
+      required: false
+      default: newrelic/nrdot-agent-artifacts
```
gsanchezgavier: nrdot?

agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure.nri_redis-0.1.0.yaml
```diff
@@ -0,0 +1,45 @@
...
+          repository: ${nr-var:oci.repository}
+          version: ${nr-var:version}
+          public_key_url: https://publickeys.newrelic.com/g/agent-control-oci/global/nri-redis/jwks.json
+  shared_filesystem:
+    infra-agent-ohi-configs:
+      kind: dir
+      entries:
+        nri-redis.yaml:
+          kind: file
+          text: ${nr-var:config_integration}
```
gsanchezgavier: 🤔  shouldn't this be 
```
text: |
  ${nr-var:config_integration}
```
other agent types declare it like that

test/e2e-runner/src/common.rs
```diff
@@ -10,6 +10,7 @@ pub mod fleet_control_api;
 pub mod logs;
 pub mod nrql;
 pub mod oci;
+pub mod ohai;
```
gsanchezgavier: nit: ohi

agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure.nri_redis-0.1.0.yaml
paologallinaharbur: Thanks for noticing, I've removed the test one for the new one and I took the wrong one 🤦

test/e2e-runner/src/linux/scenarios/nri_redis.rs
```diff
@@ -0,0 +1,134 @@
...
+use crate::linux::install::{install_agent_control_from_recipe, tear_down_test};
+use crate::linux::redis::Redis;
+use std::time::Duration;
+use tracing::info;
+
+// Dev OCI packages pre-populated on ghcr.io.
+// When production nri-redis OCI ships, delete this scenario's dev-registry wiring and point at the real registry/repos.
+const DEV_OCI_REGISTRY: &str = "ghcr.io";
+const DEV_INFRA_AGENT_REPO: &str = "newrelic/newrelic-agent-control-infrastructure-dev";
+const DEV_INFRA_AGENT_VERSION: &str = "v1.78.0";
```
gsanchezgavier, line 20: isn't these versions been passed from cli on other tests?

test/e2e-runner/src/linux/scenarios/nri_redis.rs
```diff
@@ -0,0 +1,134 @@
...
+        .expect("--redis-version is required for this scenario");
+
+    let test_id = format!(
+        "onhost-e2e-nri-redis_{}",
+        chrono::Local::now().format("%Y-%m-%d_%H-%M-%S%.3f"),
+    );
+
+    let recipe_data = RecipeData {
+        args: args.clone(),
+        monitoring_source: "infra-agent".to_string(),
```
gsanchezgavier, line 38: i think this is deprecated

test/e2e-runner/src/linux/scenarios/nri_redis.rs
gsanchezgavier: oh i now see that we are pointing to ghcr ... , all this is going to be simplified once we have the redis package

test/e2e-runner/src/linux/scenarios/nri_redis.rs
```diff
@@ -0,0 +1,134 @@
...
+config_integration:
+  integrations:
+    - name: nri-redis
+      env:
+        HOSTNAME: 127.0.0.1
+        PORT: "6379"
+        REMOTE_MONITORING: "true"
+      interval: 15s
+      labels:
+        host.id: {test_id}
```
gsanchezgavier: i think the host.id is a key that the NR BE is already using as metadata, i would rename is to `test.id` just to avoid any confusion

agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure.nri_redis-0.1.0.yaml
paologallinaharbur: I have no strong opinion on that, 
However, I am changing also the variable name to config only, no?

test/e2e-runner/src/windows/redis.rs
```diff
@@ -0,0 +1,49 @@
...
+use crate::common::test::retry_panic;
+use crate::windows::powershell::{download_file, extract};
+use crate::windows::utils::as_user_dir;
+use std::net::TcpStream;
+use std::process::Command;
+use std::time::Duration;
+use tracing::info;
+
+const REDIS_URL: &str =
+    "https://github.com/tporadowski/redis/releases/download/v5.0.14.1/Redis-x64-5.0.14.1.zip";
```
gsanchezgavier: isn't a more official source? maybe on some package manager?

gsanchezgavier (commented):
thanks , i left some minor comments / concerns

test/e2e-runner/src/windows/redis.rs
paologallinaharbur: I am testing out https://www.memurai.com/faq. It is not a real redis, but it is more official and if compatible would work perfectly for us

gsanchezgavier (approved):
thanks!

PR #2701, opened 2026-07-17
feat: remove persistent/ephemeral flag for filesystem [NR-579376]

sigilioso:
## Summary

This PR removes the `persistent:` field from on-host `file`/`dir` filesystem entries. Declared filesystem entries now always survive sub-agent stop, restart, and config-apply, instead of requiring every entry down a branch to opt in with `persistent: true`. There was no known use case for the `persistent: false` and removing the field prevents some potential footguns (the expected behavior when persistent/non-persistent elements were nested was hard to understand.

## Changes

- Drop support for `persistent: field` and remove the corresponding validators.
- Simplify `delete_not_declared` to apply only on top-level elements (everything else persist over restarts)
- Update the corresponding Agent Types to remove the `persistent` field.
- Improve unit test coverage for filesystem related operations.

### DIFF
--- a/agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
+++ b/agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
@@ -105,13 +105,6 @@ impl FileSystem {
     /// Deletes the on-disk path of every top-level entry that is not declared. Never descends into
     /// a declared `dir`'s own contents: those are left to the agent (or an earlier agent-type
     /// version) to own, and are only re-rendered by `write`, not pruned.
-    ///
-    /// # Known limitation: previously-declared entries inside a declared directory
-    ///
-    /// When AC configuration is changed **locally**, a file which **was** declared inside a directory
-    /// in a previous agent-type version but is no longer declared in the current one survives
-    /// since directory contents are never pruned. We would need to save a separate state of what was
-    /// written in order to handle this.
     pub fn delete_not_declared(
         &self,
         file_ops: &impl FileDeleter,
--- a/agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
+++ b/agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
@@ -2,10 +2,10 @@
 use fs::file::copier::FileCopier;
 use fs::file::deleter::FileDeleter;
 use fs::{directory_manager::DirectoryManager, file::writer::FileWriter};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::path::{Path, PathBuf};
 use thiserror::Error;
-use tracing::{trace, warn};
+use tracing::trace;
 
 /// Rendered filesystem tree, ready to be materialized on disk.
 ///
@@ -33,15 +33,11 @@ pub enum RenderedEntry {
     File {
         /// Where the file's bytes come from.
         content: FileContent,
-        /// The persistency attribute marking its lifecycle.
-        persistent: bool,
     },
     /// A directory containing child entries keyed by their relative path.
     Dir {
         /// The dictionary containing each children path and the entry.
         children: HashMap<PathBuf, RenderedEntry>,
-        /// The persistency attribute marking its lifecycle.
-        persistent: bool,
     },
     /// A directory whose files were projected from a map (filename to content).
     DirContentFromMap {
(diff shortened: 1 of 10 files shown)


agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -147,28 +102,16 @@ impl FileSystem {
...
-    }
-
-    /// Deletes the on-disk path of every entry in the filesystem that is not declared and is not a
-    /// child of a persistent declared directory.
+    /// Deletes the on-disk path of every top-level entry that is not declared. Never descends into
+    /// a declared `dir`'s own contents: those are left to the agent (or an earlier agent-type
+    /// version) to own, and are only re-rendered by `write`, not pruned.
     ///
-    /// # Known limitation — previously-declared entries inside persistent directories
+    /// # Known limitation: previously-declared entries inside a declared directory
```
danielorihuela: Aren't we missing the consequence?


```suggestion
    /// # Known limitation: previously-declared entries inside a declared directory are not pruned
```

And isn't this explained above?

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
```diff
@@ -147,28 +102,16 @@ impl FileSystem {
...
-            entry.delete_ephemeral(path, file_ops, dir_manager)?;
-        }
-        Ok(())
-    }
-
-    /// Deletes the on-disk path of every entry in the filesystem that is not declared and is not a
-    /// child of a persistent declared directory.
+    /// Deletes the on-disk path of every top-level entry that is not declared. Never descends into
+    /// a declared `dir`'s own contents: those are left to the agent (or an earlier agent-type
+    /// version) to own, and are only re-rendered by `write`, not pruned.
```
danielorihuela, line 107: When was this decided? Shouldn't we remove also the not declared files in "lower" folders?
I'm okay with the current approach, is way simpler.

docs/INTEGRATING_AGENTS.md
```diff
@@ -512,31 +507,15 @@ logging.d/
...
+**`dir_content_from_map` is the one exception**: instead of reconciling individual files like a `dir`, its whole directory is treated as a single unit, fully derived from its `map[string]yaml` variable. Every write deletes the directory and recreates it from the current map, rather than merging changes into what's already there. Like a `file: text` entry, its on-disk content is entirely determined by the current variable value, so unchanged variables produce the exact same files across a stop, restart, or config-apply, and a key removed from the map is gone after the next write. Unlike `file`/`dir`, though, this reconciliation is unconditional: any content that lands in the directory by other means (for example, the agent process writing into it) is wiped on every write too, not only when the map itself changes.
```
danielorihuela, line 514: > Unlike `file`/`dir`, though, this reconciliation is unconditional: any content that lands in the directory by other means (for example, the agent process writing into it) is wiped on every write too, not only when the map itself changes.

Is this true? Isn't this contradictory to the following

>  /// Deletes the on-disk path of every top-level entry that is not declared. Never descends into
    /// a declared `dir`'s own contents: those are left to the agent (or an earlier agent-type
    /// version) to own, and are only re-rendered by `write`, not pruned.

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
sigilioso: I kept the same behavior we had for _persistent_ folders.

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
sigilioso: You are right, it is redundant. Thanks!

docs/INTEGRATING_AGENTS.md
sigilioso: That's why it says `dir_content_from_map` is the one exception: those directories are threated as one single unit that is re-built each time the agent is rendered.

```yaml
some-regular-dir: # This directory's contents are kept (it is "persistent")
  kind: dir
  # ...
some-dir-from-map: # This directory is completely re-build when the agent is rendered (a file for each entry on the map)
  kind: dir_content_from_map
  # ...
```

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
sigilioso: Simplified here: [aab6d05](https://github.com/newrelic/newrelic-agent-control/pull/2701/commits/aab6d05031b2b30204a8da8285c8ebdbd91babf9)

agent-control/src/agent_type/runtime_config/on_host/filesystem/rendered.rs
alvarocabanas: Yes, this is the behavior we want, since we want to keep all the data that was created by a subagent.

alvarocabanas (approved):
LGTM

PR #2703, opened 2026-07-17
feat: remove orphan directories on shared-fs cleanup [NR-579376]

sigilioso:
## Summary

Shared-filesystem cleanup previously only reclaimed files, the _co-owned_ `kind: dir` directories lingered forever once every agent that declared them was removed. This PR includes the full set of active agents into `ResourceCleaner` so directories can be safely reclaimed once no active agent declares them anymore.

## Changes

- Renamed `ResourceCleaner::clean` to `on_agent_removed` and added an `active_agents: &SubAgentsMap` parameter to both `on_agent_removed` and `on_agent_type_changed`.
- Added `declared_dirs` to `DeclaredPaths` to track co-owned `kind: dir` nodes separately from individually-owned files and whole-owned managed dirs.
- `reconcile_shared_dir` and `delete_stale_paths` now remove a directory once no configured/active agent declares it, and still recurse into directories that remain declared to prune stray contents.
- Added unit tests covering directory reclamation on agent removal, on type change, and during startup reconciliation.

### DIFF
--- a/agent-control/src/agent_control/resource_cleaner/on_host.rs
+++ b/agent-control/src/agent_control/resource_cleaner/on_host.rs
@@ -205,6 +205,7 @@ where
             let paths = self.declared_shared_paths(&definition);
             all_shared_paths.files.extend(paths.files);
             all_shared_paths.managed_dirs.extend(paths.managed_dirs);
+            all_shared_paths.declared_dirs.extend(paths.declared_dirs);
         }
         Ok(all_shared_paths)
     }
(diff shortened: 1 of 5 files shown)


agent-control/src/agent_control/resource_cleaner/on_host.rs
```diff
@@ -235,33 +236,33 @@ where
...
-        let mut owned_managed_dirs = HashSet::new();
+        let mut managed_dirs = HashSet::new();
+        let mut declared_dirs = HashSet::new();
         for agent_config in configured.values() {
             match self.get_definition(&agent_config.agent_type) {
                 Ok(definition) => {
                     let declared = self.declared_shared_paths(&definition);
                     expected_files.extend(declared.files);
-                    owned_managed_dirs.extend(declared.managed_dirs);
+                    managed_dirs.extend(declared.managed_dirs);
```
gsanchezgavier, line 264: why this is called managed_dirs? i found it a bit confusing to understand the differences between declared managed and active, old,  new, owned and co-owned

gsanchezgavier (approved):
LGTM just a nit on naming

agent-control/src/agent_control/resource_cleaner/on_host.rs
sigilioso: These paths are defined in the filesystem module: https://github.com/newrelic/newrelic-agent-control/blob/37f4db2b0632b7f7ddff8aadd1f6056ffaf37054/agent-control/src/agent_type/runtime_config/on_host/filesystem.rs#L166-L176

and most of the code tries to follow the same naming. I think the confusing part is `managed_dirs` vs. `declared_dirs`: the first represents the directories that cannot collide between agents and are completely re-created and the second the directories that are simply declared or referenced and can exists in several agents at the same time.

I tried to simplify in this PR and use the same naming in the `DeclaredPaths` and the rest of the code. Did I make it worse?

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -170,6 +170,9 @@ pub struct DeclaredPaths {
     /// Directories owned as a whole: the entire subtree is managed as a unit, and no
     /// other agent may own anything at or under the path.
     pub managed_dirs: HashSet<PathBuf>,
+    /// Co-owned `kind: dir` nodes. Several agents may declare the same one; a directory is only
+    /// safe to remove once no agent declares it here anymore.
+    pub declared_dirs: HashSet<PathBuf>,
```
danielorihuela, line 175: Why not name it `coowned_dirs`? Wouldn't it be simpler to understand the code?
All files are co-owned by default in this design.

sigilioso:
I'm merging to avoid blocking the release, I'll take care of the naming issues on a different PR.

agent-control/src/agent_control/resource_cleaner/on_host.rs
gsanchezgavier: indeed all code was there, and the pr is unifying some of it, forgot to mention that. It was just the first time i read the code related to the shared fs , i need to check it out and read it properly not from the ui

sigilioso:
I tried to improve naming here https://github.com/newrelic/newrelic-agent-control/pull/2715 @danielorihuela, @gsanchezgavier Let me know what you think 🙂

PR #2704, opened 2026-07-17
feat: add alert to ensure logs are produced on canaries

alvarocabanas:
- Label onhost canaries logs.
- Add alert both for k8s and onhost canaries to check logs are sent.

### DIFF
--- a/test/k8s-canaries/terraform/production/alert_nrql_templates/log_presence.tftpl
+++ b/test/k8s-canaries/terraform/production/alert_nrql_templates/log_presence.tftpl
@@ -1,4 +1,4 @@
 SELECT count(*)
 FROM Log
 WHERE cluster_name = '${instance_id}'
-AND entity.name = 'agent-control-self-instrumentation'
+AND service.name = 'agent-control-self-instrumentation'
--- a/test/k8s-canaries/terraform/production/alert_nrql_templates/log_presence.tftpl
+++ b/test/k8s-canaries/terraform/production/alert_nrql_templates/log_presence.tftpl
@@ -1,0 +1,3 @@
+SELECT count(*)
+FROM Log
+WHERE cluster_name = '${instance_id}'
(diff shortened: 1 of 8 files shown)


test/k8s-canaries/terraform/production/alert_nrql_templates/log_presence.tftpl
```diff
@@ -0,0 +1,3 @@
+SELECT count(*)
+FROM Log
+WHERE cluster_name = '${instance_id}'
```
sigilioso, line 3: Should we also check attributes that are specific to self-instrumentation logs? I foresee false-negatives if something else reporting logs to the same cluster/host is added (For instance, if fluent-bit is added).

alvarocabanas: You wer absolutely right asking for this filter. Thanks to it, I realized that on k8s all our logs come from fluentbit. 
I've enabled k8s self-instrumentation logs and I've added an extra filter to the alert queries to ensure they come from this module and not fluentbit.

sigilioso: Do you know the difference between `service.name` and `entity.name`? (Just in case one is more stable than the other) Apparently the log entries have both attributes as `'agent-control-self-instrumentation'` 🤔.

Regarding the k8s thing, don't we need to enable the self-instrumentation there in order to prevent the alert from triggering right after creating it?

alvarocabanas: The entity.name seems to be synthetized from the service.name, unless there is and entity synthesis definition to match, so I think in this case is indistinct, but maybe it's better to use service.name to cover ourselves from the unlikely scenario of creating that specific synthesis rule, I've changed it in last commit.

Regarding your second question, I don't understand what you comment, I did modify the makefile that does the helm-upgrade with:
```
--set agentControlDeployment.chartValues.config.override.self_instrumentation.opentelemetry.logs.enabled=true
```
so the k8s cluster will start emitting logs.

sigilioso: Sorry, I totally overlook the Makefile change 😓

I'd use the `service.name` as well. Thanks! 👍

sigilioso (approved):
LGTM

PR #2705, opened 2026-07-17
feat(k8s-infra): reduce number of variables

paologallinaharbur:
This is a draft PR to discuss a possible change on the nr-infra agentType.

I do see, if multiple configs as separate files becomes a thing, that we want to allow the user to set a single file for the whole chart_values instead of two.

The same change is not possible for the different agentTypes, this is not a big deal since we could still point to the inner variable and expose directly `chart_values.newrelic-prometheus-agent`


gsanchezgavier (approved):
the infra change looks compatible to me 👍 ,not the prometheus change but that looks more like a fix , perhaps is worth to instroduce a changelog line for it

paologallinaharbur:
closing for now

PR #2709, opened 2026-07-20
fix(renovate): wrong suffix was messing up with renovate rule

paologallinaharbur:
Due to this reason we have multiple values for the same variable 

```
env:
  HOST_E2E_INFRA_AGENT_VERSION: "1.73.0"

env:
  HOST_E2E_INFRA_AGENT_VERSION: "1.72.4"

env:
  HOST_E2E_INFRA_AGENT_VERSION: "1.78.0"
```


sigilioso (approved):
Thanks!

PR #2713, opened 2026-07-20
chore: add docker image make targets [NR-566516]

sigilioso:
## Summary

* Adds Makefile targets (`image/agent-control`, `image/agent-control-cli`, `image`) that wrap the `docker buildx build` calls for the Agent Control container images, so they can be easily built locally (without `minikube`/`tilt`) and reused as-is by CI. The corresponding workflows now call `make image/...` instead of duplicating the `docker buildx build` invocation inline.

* `make image` usage is documented in `docs/DEVELOPMENT.md`.

```shell
❯ make image
# ...
❯ docker images
IMAGE                                                                                                 ID             DISK USAGE   CONTENT SIZE   EXTRA
newrelic/newrelic-agent-control-cli:local                                                             6e827bba7b96        356MB         89.3MB    U   
newrelic/newrelic-agent-control:local                                                                 c4a6356f266f        215MB         57.8MB        
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,59 @@ build-%:
 tilt-up:
 	tilt up ; tilt down
 
+##########################################
+# 		     Image targets 			 #
+##########################################
+DOCKER_IMAGE_NAME_AGENT_CONTROL ?= newrelic/newrelic-agent-control
+DOCKER_IMAGE_NAME_AGENT_CONTROL_CLI ?= newrelic/newrelic-agent-control-cli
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -33,6 +33,59 @@ build-%:
 tilt-up:
 	tilt up ; tilt down
 
+##########################################
```
danielorihuela, line 36: More make targets 🤮

danielorihuela (approved):

PR #2715, opened 2026-07-21
chore(fs): improve declared-paths names

sigilioso:
This PR is a follow-up of #2703. It addresses the following comments regarding names

- https://github.com/newrelic/newrelic-agent-control/pull/2703#discussion_r3614677619
- https://github.com/newrelic/newrelic-agent-control/pull/2703#discussion_r3615606783

### DIFF
--- a/agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
+++ b/agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
@@ -165,14 +165,14 @@ impl Templateable for SharedFileSystem {
 #[derive(Debug, Default, Clone, PartialEq, Eq)]
 pub struct DeclaredPaths {
     /// Files owned individually. Several agents may drop sibling files into the same
-    /// co-owned directory, but no two may own the exact same file path.
-    pub files: HashSet<PathBuf>,
+    /// shared directory, but no two may own the exact same file path.
+    pub owned_files: HashSet<PathBuf>,
     /// Directories owned as a whole: the entire subtree is managed as a unit, and no
     /// other agent may own anything at or under the path.
-    pub managed_dirs: HashSet<PathBuf>,
-    /// Co-owned `kind: dir` nodes. Several agents may declare the same one; a directory is only
+    pub owned_dirs: HashSet<PathBuf>,
+    /// Shared `kind: dir` nodes. Several agents may declare the same one; a directory is only
     /// safe to remove once no agent declares it here anymore.
-    pub declared_dirs: HashSet<PathBuf>,
+    pub shared_dirs: HashSet<PathBuf>,
 }
 
 impl SharedFileSystem {
@@ -189,21 +189,21 @@ impl SharedFileSystem {
 
 /// Recursively gathers the paths declared by `entry` (rooted at `path`) into `declared`.
 ///
-/// `Dir` entries are co-owned drop zones: the directory path is tracked separately in
-/// `declared_dirs` (not `files`/`managed_dirs`), since several agents may declare the same one.
+/// `Dir` entries are shared drop zones: the directory path is tracked separately in
+/// `shared_dirs` (not `owned_files`/`owned_dirs`), since several agents may declare the same one.
 fn collect_declared_paths(path: &Path, entry: &FilesystemEntry, declared: &mut DeclaredPaths) {
     match entry {
         FilesystemEntry::File { .. } => {
-            declared.files.insert(path.to_path_buf());
+            declared.owned_files.insert(path.to_path_buf());
         }
         FilesystemEntry::Dir { entries, .. } => {
-            declared.declared_dirs.insert(path.to_path_buf());
+            declared.shared_dirs.insert(path.to_path_buf());
             for (key, child) in entries {
                 collect_declared_paths(&path.join(key), child, declared);
             }
         }
         FilesystemEntry::DirContentFromMap { .. } => {
-            declared.managed_dirs.insert(path.to_path_buf());
+            declared.owned_dirs.insert(path.to_path_buf());
         }
     }
 }
@@ -1162,7 +1162,7 @@ projected:
         let declared = shared.declared_paths(base);
 
         assert_eq!(
-            declared.files,
+            declared.owned_files,
             HashSet::from([
                 base.join("top-file"),
                 base.join("co-owned").join("nri-redis.yaml"),
@@ -1171,12 +1171,12 @@ projected:
             "every `kind: file` must be reported, and the co-owned dir path must not be"
         );
         assert_eq!(
-            declared.managed_dirs,
+            declared.owned_dirs,
             HashSet::from([base.join("projected")]),
             "`dir_content_from_map` must be reported as a whole-directory owner"
         );
         assert_eq!(
-            declared.declared_dirs,
+            declared.shared_dirs,
             HashSet::from([base.join("co-owned"), base.join("co-owned").join("nested"),]),
             "every `kind: dir` node, including nested ones, must be reported as co-owned"
         );
@@ -1187,8 +1187,8 @@ projected:
     fn declared_paths_of_empty_is_empty() {
         let tmp_dir = TempDir::new().unwrap();
         let declared = SharedFileSystem::default().declared_paths(tmp_dir.path());
-        assert!(declared.files.is_empty());
-        assert!(declared.managed_dirs.is_empty());
-        assert!(declared.declared_dirs.is_empty());
+        assert!(declared.owned_files.is_empty());
+        assert!(declared.owned_dirs.is_empty());
+        assert!(declared.shared_dirs.is_empty());
     }
 }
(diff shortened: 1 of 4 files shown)


agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
gsanchezgavier: is this comment correct? this is used for both filesystems right?

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -165,14 +165,14 @@ impl Templateable for SharedFileSystem {
...
+    pub owned_files: HashSet<PathBuf>,
     /// Directories owned as a whole: the entire subtree is managed as a unit, and no
     /// other agent may own anything at or under the path.
-    pub managed_dirs: HashSet<PathBuf>,
-    /// Co-owned `kind: dir` nodes. Several agents may declare the same one; a directory is only
+    pub owned_dirs: HashSet<PathBuf>,
+    /// Shared `kind: dir` nodes. Several agents may declare the same one; a directory is only
     /// safe to remove once no agent declares it here anymore.
-    pub declared_dirs: HashSet<PathBuf>,
+    pub shared_dirs: HashSet<PathBuf>,
```
gsanchezgavier: perhaps just `dirs` because all are shared... or not if is used in both

gsanchezgavier: unowned word also exists but sounds off 😆 . it is actually shared ownership

sigilioso: I'd use `dirs` it also maps to the `kind: dir` in the yaml.

```yaml
shared-filesystem:
  some-dir:
    kind: dir # Can be in many agents
    entries:
      - some-file: # Can be in one agent only (owned)
        kind: file
         # ...
      - some-map-dir:
        kind: dir_content_from_map # Can be in one agent only (owned)
        # ...
```

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
sigilioso: It used to be correct, but It is not anymore. Too many details in comments are counterproductive! Fixed: [b5db5e0](https://github.com/newrelic/newrelic-agent-control/pull/2715/commits/b5db5e0e58a910e95e2dd14eb442c2cff9bc1fd7)

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
sigilioso: Used `dirs` here: [b5db5e0](https://github.com/newrelic/newrelic-agent-control/pull/2715/commits/b5db5e0e58a910e95e2dd14eb442c2cff9bc1fd7)

gsanchezgavier (dismissed):

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -160,19 +160,19 @@ impl Templateable for SharedFileSystem {
...
+    pub owned_files: HashSet<PathBuf>,
     /// Directories owned as a whole: the entire subtree is managed as a unit, and no
     /// other agent may own anything at or under the path.
-    pub managed_dirs: HashSet<PathBuf>,
-    /// Co-owned `kind: dir` nodes. Several agents may declare the same one; a directory is only
+    pub owned_dirs: HashSet<PathBuf>,
+    /// Shared `kind: dir` nodes. Several agents may declare the same one; a directory is only
     /// safe to remove once no agent declares it here anymore.
-    pub declared_dirs: HashSet<PathBuf>,
+    pub dirs: HashSet<PathBuf>,
```
paologallinaharbur: is DeclaredPaths used both for filesystem and sharedFilesystem? I am a bit confused by descriptions 😅

paologallinaharbur: IMO the description does not clarify the behaviour with "normal" filesystem

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -189,21 +189,21 @@ impl SharedFileSystem {
 
 /// Recursively gathers the paths declared by `entry` (rooted at `path`) into `declared`.
 ///
-/// `Dir` entries are co-owned drop zones: the directory path is tracked separately in
-/// `declared_dirs` (not `files`/`managed_dirs`), since several agents may declare the same one.
+/// `Dir` entries are shared drop zones: the directory path is tracked separately in
```
paologallinaharbur: also this, this function is called also in the "normal" filesystem, no?
I find confusing the "`Dir` entries are shared drop zones"

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -160,19 +160,19 @@ impl Templateable for SharedFileSystem {
...
     /// Directories owned as a whole: the entire subtree is managed as a unit, and no
     /// other agent may own anything at or under the path.
-    pub managed_dirs: HashSet<PathBuf>,
-    /// Co-owned `kind: dir` nodes. Several agents may declare the same one; a directory is only
+    pub owned_dirs: HashSet<PathBuf>,
+    /// Shared `kind: dir` nodes. Several agents may declare the same one; a directory is only
     /// safe to remove once no agent declares it here anymore.
-    pub declared_dirs: HashSet<PathBuf>,
+    pub dirs: HashSet<PathBuf>,
 }
```
danielorihuela, line 173: I like it better than before. I still think it's confusing. `dirs` don't contain all paths, but the name makes me think this should be a superset of `owned_dirs`, but it's okay for now.

I was wondering if we could simplify this even more. Now that we remove shared folders once no agent points to it, could we have a single `dirs` field? I don't know if that would be possible. I think it makes sense. If in the future we remove the `dir_content_from_map` after @gsanchezgavier work on logging (if that's possible), the code would remain the same.

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
sigilioso: I'll review all of them, there are probably more leftovers! Thanks!

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
paologallinaharbur: I agree, it was the first time I was looking at the code and I ended up checking out the whole branch to understand what was going on 😕

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
sigilioso: Addressed here: [1718d25](https://github.com/newrelic/newrelic-agent-control/pull/2715/commits/1718d25eb175761bbfde8f9c18e073b791217d06)

sigilioso: Also addressed here: [1718d25](https://github.com/newrelic/newrelic-agent-control/pull/2715/commits/1718d25eb175761bbfde8f9c18e073b791217d06)

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
sigilioso: Also addressed here: [1718d25](https://github.com/newrelic/newrelic-agent-control/pull/2715/commits/1718d25eb175761bbfde8f9c18e073b791217d06)

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
sigilioso: [1718d25](https://github.com/newrelic/newrelic-agent-control/pull/2715/commits/1718d25eb175761bbfde8f9c18e073b791217d06) changes it to two fields (`owned_files`/`owned_dirs` -> `exclusive`).

It also fixes a point of failure: `write` now clears a stale entry when its on-disk shape doesn't match what's being written, so this self-heals on the next apply regardless of what reconcile did. Example: on an agent-type update something declared as `dir` becomes something declared as `file`.

@danielorihuela, @paologallinaharbur LMKWYT

danielorihuela: I'm not gonna lie, it's still confusing. But I like it more!!! I'm not approving in case @paologallinaharbur wants to say anything.

danielorihuela (approved):

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
paologallinaharbur: I do not want to be the blocker, I saw it quickly and it looked better! having exclusive seems simplifying the overall solution!

PR #2716, opened 2026-07-21
fix: remove set permission on hook

gsanchezgavier:
Removes the auto chmod change on the binary being executed ( that could be a shell or whatever system binary )


sigilioso (approved):
Thanks! LGTM!

Do we know if there was some use-case for this to be added (and unit-tested)?

gsanchezgavier:
> Thanks! LGTM!
> 
> Do we know if there was some use-case for this to be added (and unit-tested)?

Asked this to @vjripoll, and he mentioned it was just in case the script had't execution permissions. But this is changing the permissions for the file declared in the path which in case of a bash script will be bash. Anyway I don't think the solution for that use-case is to mutate the file permissions on AC

PR #2717, opened 2026-07-21
feat: improve hook execution troubleshooting

gsanchezgavier:
Post-download hooks previously discarded stdout and only surfaced stderr on failure, making install-script issues hard to diagnose. This captures both streams on dedicated drain threads and logs them on success, failure, and timeout.

**Changes:**
- Pipe + drain stdout/stderr on separate threads to avoid pipe-buffer deadlock
- Log `stdout`/`stderr` fields on the success, failure, and timeout paths
- Extract `DEFAULT_TIMEOUT`/`POLL_INTERVAL` constants and make `timeout` a struct field (test-injectable)
- Add cross-platform (unix/windows) tests for the three logging paths via per-platform script helpers
- Replaces a 5 minute integration tests for a unit test

**Notes for reviewers:**
- On timeout the executor now `join()`s the reader threads after killing the child; a grandchild that inherits the stdio pipe (e.g. a backgrounded process) can keep those `read_to_end` calls blocked past the timeout. Process-tree teardown is still a TODO i'm working for a follow up PR .

### DIFF
--- a/agent-control/src/package/post_download_hook_executor.rs
+++ b/agent-control/src/package/post_download_hook_executor.rs
@@ -8,6 +8,9 @@ use tracing::{debug, warn};
 
 use crate::agent_type::runtime_config::on_host::package::rendered::PostDownloadHook;
 
+const DEFAULT_TIMEOUT: Duration = Duration::from_secs(300);
+const POLL_INTERVAL: Duration = Duration::from_millis(100);
+
 /// Errors that can occur while executing a post-download hook.
 #[derive(thiserror::Error, Debug)]
 pub enum PostDownloadHookExecutionError {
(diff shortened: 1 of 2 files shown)


vjripoll (dismissed):

agent-control/src/package/post_download_hook_executor.rs
```diff
@@ -240,6 +305,85 @@ mod tests {
...
+    fn test_execute_timeout_logs_partial_output() {
+        let temp_dir = TempDir::new().unwrap();
+        let script_path = temp_dir
+            .path()
+            .join(format!("hanging_hook.{}", script_extension()));
+
+        let content = join_commands(&[
+            print_stdout_no_newline("partial-stdout"),
+            print_stderr_no_newline("partial-stderr"),
+            sleep_command(),
```
danielorihuela, line 364: I think it makes sense to print something more after the sleep command and later assert that this doesn't appear in the logs.

danielorihuela (approved):

PR #2718, opened 2026-07-21
fix: kill process group on hook execution

gsanchezgavier:
### Summary
On timeout, the post-download hook executor only killed the direct child, leaving
grandchildren alive and holding the piped stdout/stderr open — blocking `execute()`. This
introduces a cross-platform `ProcessGroup` primitive so the entire process tree is terminated.

### Changes
- Add `utils::process_group::ProcessGroup` (unix: pgid + `killpg`; windows: Job Object).
- Post-download hook timeout now group-kills the tree instead of `child.kill()` (removes TODO).
- Move `job_object` from `sub_agent/on_host/command` to `utils`; give it its own `JobObjectError`.
- Replace `CommandError::WinError(String)` with `#[from] JobObjectError` (windows-gated).
- Add unix/windows tests asserting grandchildren are killed and pipes close promptly.
- Removes failure scenario from e2e

### Notes for reviewers
- Stacked on `gsanchez/feat/hook-logs`; review targets that base, not `main`.

### DIFF
--- a/agent-control/src/utils/process_group.rs
+++ b/agent-control/src/utils/process_group.rs
@@ -1,0 +1,210 @@
+//! Cross-platform primitive for killing a spawned process together with any descendants it
+//! spawned, not just the direct child.
+use std::process::{Child, Command};
+
+#[cfg(target_family = "windows")]
+use crate::utils::job_object::JobObject;
+
+/// Error produced while spawning, attaching to, or killing a process group.
+#[derive(thiserror::Error, Debug)]
+#[error("{0}")]
+pub struct ProcessGroupError(String);
+
+/// Handle to the process group of a spawned child, allowing the whole tree (the child and any
+/// descendants it spawned) to be killed together.
+pub struct ProcessGroup {
+    #[cfg(target_family = "unix")]
+    pgid: i32,
+    #[cfg(target_family = "windows")]
+    job_object: JobObject,
+}
+
+impl ProcessGroup {
+    /// Spawns `command` as the leader of a new process group and attaches to it, returning the
+    /// child together with a handle that can kill the whole group. If the group cannot be
+    /// established after a successful spawn, the child is killed and reaped before returning.
+    pub fn spawn(mut command: Command) -> Result<(Child, Self), ProcessGroupError> {
+        Self::prepare(&mut command);
+        let mut child = command
+            .spawn()
+            .map_err(|e| ProcessGroupError(format!("failed to spawn process: {e}")))?;
+        match Self::attach(&child) {
+            Ok(process_group) => Ok((child, process_group)),
+            Err(e) => {
+                let _ = child.kill();
+                Err(e)
+            }
+        }
+    }
+
+    /// Kills every process in the group.
+    pub fn kill(self) -> Result<(), ProcessGroupError> {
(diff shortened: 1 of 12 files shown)


test/e2e-runner/src/linux/scenarios/post_download_hook.rs
```diff
@@ -17,7 +17,6 @@ const AGENT_TYPE: &str = "newrelic/com.newrelic.posthook_e2e:0.1.0";
 const AGENT_TYPE_FILE: &str = "/etc/newrelic-agent-control/dynamic-agent-types/posthook_e2e.yaml";
```
danielorihuela, line 17: We could even remove the happy path, given that the ebpf agent e2e test includes a postdownload script.

agent-control/src/utils/process_group.rs
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    /// Kills every process in the group.
+    pub fn kill(self) -> Result<(), ProcessGroupError> {
+        #[cfg(target_family = "unix")]
+        {
+            use nix::sys::signal::{Signal, killpg};
+            use nix::unistd::Pid;
+
+            killpg(Pid::from_raw(self.pgid), Signal::SIGKILL)
```
danielorihuela, line 47: I remember a while ago we did some work to do a graceful shutdown of the subagent processes or something like that. Maybe it doesn't have anything to do, but... have you thought about that? Like, maybe there's some refactor that we could do there for the future?

danielorihuela (approved):

agent-control/src/utils/process_group.rs
gsanchezgavier: yes , we worked on introducing the Job Object for the windows executor to avoid the orphan process as the graceful shutdown signal was not something standard and supported by agents. 

I moved the job object to make some unification but stopped there as i'm not sure the goals are the same. I think the current implementation is ok for now, we can iterate and add more features in the future if needed and make sense

test/e2e-runner/src/linux/scenarios/post_download_hook.rs
gsanchezgavier: oh it already does? cool i will do it then ! also i don't think e2e are the place to test this features as it can be perfectly covered by integration test

danielorihuela: Well, maybe I'm too naive. We have the e2e https://github.com/newrelic/newrelic-agent-control/blob/main/test/e2e-runner/src/linux/scenarios/ebpf_agent.rs which is using the embedded type right?
The embedded type has a post install script configured there https://github.com/newrelic/newrelic-agent-control/blob/0964d23e735c4e630a5b532b552223bc85d6a586/agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.ebpf-0.1.0.yaml#L44.

I downloaded the 1.5.0 package with oras and the script is there.

Captura de pantalla 2026-07-27 a las 9 23 25

gsanchezgavier: yes is ok, also we have a set of integration tests running on linux/windows, that is the way to cover all feature like this . We have some exceptions in specific cases like self update but not need for this

PR #2719, opened 2026-07-21
feat: nri mysql agent type

danielorihuela:
Adds nri mysql agent type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,9 @@ Remember that the keywords that you can use are the following:
 ## Unreleased
 
 ### enhancement
-- On-host: added apache and nginx agentTypes
+- On-host: added apache and nginx agentTypes.
 - On-host: added nri-postgresql agentType.
+- On-host: added nri-mysql agentType.
 - Added `agent-type validate --file <path>` subcommand to `newrelic-agent-control-cli` and `newrelic-agent-control-k8s-cli` for schema-level validation of agent type definition files.
 - Remove static regex validator for newrelic-infra Agent Type.
 
--- a/agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure.nri_mysql-0.1.0.yaml
+++ b/agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure.nri_mysql-0.1.0.yaml
@@ -1,0 +1,46 @@
+namespace: newrelic
+name: com.newrelic.infrastructure.nri_mysql
+version: 0.1.0
+platform: host
+operating_system: linux
+protocol_version: "1.0"
+variables:
+  config:
+    description: "nri-mysql integration YAML consumed by the infra-agent"
+    type: yaml
+    required: true
+  version:
+    description: "Agent version"
+    type: string
+    required: true
+  oci:
+    repository:
+      description: "OCI repository name for nri-mysql"
+      type: string
+      required: false
+      default: newrelic/nri-mysql-artifacts
+      variants:
+        ac_config_field: "oci_repository_urls"
+        values: [ "newrelic/nri-mysql-artifacts" ]
+deployment:
+  packages:
+    nri-mysql:
+      download:
+        oci:
+          repository: ${nr-var:oci.repository}
+          version: ${nr-var:version}
+          public_key_url: https://publickeys.newrelic.com/g/agent-control-oci/global/nri-mysql/jwks.json
+  shared_filesystem:
+    infra-agent-ohi-configs:
+      kind: dir
+      entries:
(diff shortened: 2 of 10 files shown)


agent-control/src/agent_type/registry/local/agent_type_validation_tests.rs
sigilioso, line 1: Totally unrelated to the PR: I'm starting to think that these tests are slightly deprecated (or they will become deprecated when we stop doing changes in the embedded Agent Types).

Kind of related to the PR: something that could be useful instead would be checking if all OHI agent types can be used at the same type.

It would be basically having some sort of AC configuration with all of them and check it with the corresponding validator.

```yaml
agents:
  nri_redis:
    agent_type: newrelic/com.newrelic.infrastructure.nri_redis:0.1.0
  nri_mysql:
     agent_type: newrelic/com.newrelic.infrastructure.nri_mysql:0.1.0
# ...
```

Maybe @paologallinaharbur is already covering this in the e2e he is working on.

agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure.nri_mysql-0.1.0.yaml
```diff
@@ -0,0 +1,46 @@
...
+    required: true
+  oci:
+    repository:
+      description: "OCI repository name for nri-mysql"
+      type: string
+      required: false
+      default: newrelic/nri-mysql-artifacts
+      variants:
+        ac_config_field: "oci_repository_urls"
+        values: [ "newrelic/nri-mysql-artifacts" ]
```
sigilioso, line 24: Is it already available? If not, do we need to wait for it to be available?

sigilioso (commented):
I have only one doubt concerning the package availability, everything else looks good to me.

agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure.nri_mysql-0.1.0.yaml
danielorihuela: No, it's not yet. It's the same issue that @paologallinaharbur had for `nri-redis`. That metadata needs this for the release, and we need the package to test it.

I wouldn't merge this as an embedded package, but since this is what we did with redis, I'm adding it here.

I don't really know how to proceed. My idea was to merge this and then ask metadata to release.

agent-control/src/agent_type/registry/local/agent_type_validation_tests.rs
paologallinaharbur: I totally agree, they seems quite useless, I am adding them as well since it is easier to generate them rather then checking if they still make sense

paologallinaharbur: I would keep for now, but obviously stop once the agentTypes are extracted

agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure.nri_mysql-0.1.0.yaml
paologallinaharbur: > wouldn't merge this as an embedded package

what do you mean? That you would not include the agentType?

paologallinaharbur: The issue is that someone needs to start supporting a new type, and I do not see any problem being us the first ones

danielorihuela: I mean that we should use remote agent types. Push the new oci package, push the new agent type, and test it from there. But we decided to have them embedded for now.

CHANGELOG.md
```diff
@@ -15,8 +15,9 @@ Remember that the keywords that you can use are the following:
 ## Unreleased
 
 ### enhancement
-- On-host: added apache and nginx agentTypes
+- On-host: added apache and nginx agentTypes.
```
paologallinaharbur, line 18: can you add nri- to this lines to both agents. I think I missed those

paologallinaharbur (approved):
ignore my nit if that is the only one to avoid waiting again all pipelines 😄

CHANGELOG.md
danielorihuela: Maybe @sigilioso can do them as part of his branch which is out of date.

sigilioso: Sure!

sigilioso (approved):

PR #2720, opened 2026-07-22
feat: refactor onhost canarie alerts

vjripoll:
Refactors the onhost canary alerting in `test/onhost-canaries/terraform/main.tf` to reuse the shared `../../terraform/modules/nr_alerts` module instead of duplicating the policy/notification/workflow/condition resources inline for each canary instance.

### DIFF
--- a/test/onhost-canaries/terraform/alert_nrql_templates/generic_metric_derivative.tftpl
+++ b/test/onhost-canaries/terraform/alert_nrql_templates/generic_metric_derivative.tftpl
@@ -1,4 +1,4 @@
-SELECT max(${metric}) OR 0
+SELECT ${metric}
 FROM ${sample}
 WHERE (
   hostname = '${instance_id}'
--- a/test/onhost-canaries/terraform/alert_nrql_templates/generic_metric_derivative.tftpl
+++ b/test/onhost-canaries/terraform/alert_nrql_templates/generic_metric_derivative.tftpl
@@ -1,5 +1,5 @@
 SELECT ${metric}
-FROM ${sample}
+FROM ProcessSample
 WHERE (
   hostname = '${instance_id}'
   AND processDisplayName LIKE 'newrelic-agent-control%'
--- a/test/onhost-canaries/terraform/alert_nrql_templates/generic_metric_threshold.tftpl
+++ b/test/onhost-canaries/terraform/alert_nrql_templates/generic_metric_threshold.tftpl
@@ -1,5 +1,5 @@
 SELECT max(${metric}) OR 0
-FROM ${sample}
+FROM ProcessSample
 WHERE (
   hostname = '${instance_id}'
   AND processDisplayName LIKE 'newrelic-agent-control%'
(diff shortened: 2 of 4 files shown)


test/onhost-canaries/terraform/alert_nrql_templates/generic_metric_threshold.tftpl
danielorihuela, line 1: I was thinking... Would it make sense to rename that one to `generic_metric_max.tfpl`?
Maybe on a different PR, because we should change k8s one too.

danielorihuela (commented):
I like what I see. Good job!!! Did you try to launch it in a clean "test" environment to see that it works?

test/onhost-canaries/terraform/alert_nrql_templates/generic_metric_threshold.tftpl
vjripoll: yes i tested in staging one at least :)

test/onhost-canaries/terraform/main.tf
```diff
@@ -361,143 +349,19 @@ output "ansible_inventory_content" {
   sensitive   = true
 }
 
+# Create alert policy, workflow, notification channels and NRQL conditions for each instance via the shared module
+module "alerts" {
+  source = "../../terraform/modules/nr_alerts"
```
gsanchezgavier, line 356: cool i guess this module has all of the removed resources

test/onhost-canaries/terraform/alert_nrql_templates/generic_metric_derivative.tftpl
```diff
@@ -0,0 +1,9 @@
+SELECT ${metric}
+FROM ${sample}
```
gsanchezgavier: i would also pin the process sample here , i don't think any other sample would work given the where condition containing `processDisplayName`

gsanchezgavier (commented):
thanks looks good! i just left a minor thing

test/onhost-canaries/terraform/main.tf
```diff
@@ -160,22 +155,20 @@ locals {
...
         name               = "Memory growth (bytes/hour)"
         metric             = "derivative(memoryResidentSizeBytes, 1 hour)"
-        sample             = "ProcessSample"
         aggregation_window = 10800
         slide_by           = 3600
         threshold          = 210000
         duration           = 21600
         operator           = "above"
-        template_name      = "./alert_nrql_templates/generic_metric_threshold.tftpl"
+        template_name      = "./alert_nrql_templates/generic_metric_derivative.tftpl"
```
gsanchezgavier, line 173: this doesn't look like have any derivative specific right?

gsanchezgavier (dismissed):
thanks , please after merge make sure alerts are created as they should be, and old ones are removed

test/onhost-canaries/terraform/main.tf
vjripoll: yep, nothing derivative-specific in the template itself, it's just the version without the max() OR 0 wrapper the name reflects the use case, not the NRQL. I'll fold a rename of this one into the same follow-up PR as the generic_metric_threshold -> generic_metric_max rename (plus the k8s copy).

gsanchezgavier (approved):

PR #2727, opened 2026-07-23
feat(cli): add agent type semantic validation

gsanchezgavier:
## Summary
Extends the `agent-type validate` CLI subcommand with a semantic check on top of the existing
schema validation: every `${nr-var:X}` reference in a definition's `deployment` block must resolve
to a declared `variables` entry. This catches typos and missing declarations before Agent Control
ever runs the definition.

## Notes for reviewers
- Only `nr-var` references are validated; `nr-env` / `nr-sub` / `nr-ac` are intentionally ignored.
- Unused *declared* variables are allowed (validation is one-directional: referenced ⊆ declared).


agent-control/src/agent_type/semantic_validation.rs
```diff
@@ -0,0 +1,219 @@
...
+    variables: &VariableDefinitionTree,
+) -> Result<(), SemanticValidationError> {
+    #[derive(Deserialize)]
+    struct Deployment {
+        deployment: serde_json::Value,
+    }
+
+    let Deployment { deployment } = serde_saphyr::from_slice(raw_agent_type_content).expect(
+        "content should already be valid, having been parsed by AgentTypeDefinition::from_slice",
+    );
```
sigilioso: I wonder if it would be worth it to have some specific method to get both the `AgentTypeDefinition` and the raw deployment in the module in charge of parsing (just to avoid the re-parsing and the `except`) 🤔

agent-control/src/agent_type/semantic_validation.rs
```diff
@@ -0,0 +1,219 @@
...
+use thiserror::Error;
+
+use crate::agent_type::definition::VariableDefinitionTree;
+use crate::agent_type::templates::template_re;
+use crate::agent_type::variable::namespace::Namespace;
+
+#[derive(Error, Debug, PartialEq, Eq)]
+#[error(
+    "undeclared variable(s) referenced in deployment: {}",
+    .undeclared_variables.iter().cloned().collect::<Vec<_>>().join(", ")
```
sigilioso: Really nit: we could use `.map(String::as_str)` instead of `cloned()` and avoid cloning

agent-control/src/agent_type/semantic_validation.rs
```diff
@@ -0,0 +1,219 @@
...
+        return Ok(());
+    }
+
+    Err(SemanticValidationError {
+        undeclared_variables,
+    })
+}
+
+/// Collects the set of `nr-var` names referenced anywhere in `deployment`, ignoring other
+/// namespaces (`nr-env`, `nr-sub`, `nr-ac`, ...) and any pipe function suffix.
```
sigilioso: Even if we skip them, would it make sense to check that all namespaces in variables (`${<namespace>:<name>}`) are known (valid/supported namespaces)?

sigilioso (dismissed):
Just some _nits_ and a question. Thanks for taking care of it!

agent-control/src/agent_type/semantic_validation.rs
gsanchezgavier: yesss good catch

agent-control/src/agent_type/semantic_validation.rs
gsanchezgavier: ohhh i forgot to remove that except! good catch , we should defenetly should throw an error here. Regarding the method you say i will explore it

agent-control/src/cli/common/agent_type_validation.rs
```diff
@@ -28,6 +30,16 @@ pub fn validate(args: Args) -> Result<(), CliError> {
         ))
     })?;
 
+    semantic_validation::validate_variable_references(&content, &definition.variables).map_err(
```
danielorihuela: Nit: Instead of passing the `content`, we could pass as input `definition.runtime_config.deployment`.

This avoids an extra deserialization.

danielorihuela:
Would it make sense to create a sustain task to add validation for `nr-ac` and `nr-sub`?

agent-control/src/cli/common/agent_type_validation.rs
gsanchezgavier: is not than straight forward as we would need to serialize it again so we can perform a generic check over all things declared. Anyway i was revisiting the desing again and made a refactor that it somehow improve what you are pointing , but the main driver was that the interface of that function was a bit odd (asking for the raw content and a deserialized content at the same time) and thought to be convinient

agent-control/src/agent_type/semantic_validation.rs
gsanchezgavier: done

agent-control/src/agent_type/validation.rs
```diff
@@ -0,0 +1,412 @@
...
+    let unknown_namespaces = unknown_namespaces(&references);
+    if !unknown_namespaces.is_empty() {
+        return Err(ValidationError::UnknownNamespaces(unknown_namespaces));
+    }
+
+    let declared_variables: BTreeSet<String> = definition.variables.flatten().into_keys().collect();
+    let undeclared_variables = undeclared_variables(&references, &declared_variables);
+    if !undeclared_variables.is_empty() {
+        return Err(ValidationError::UndeclaredVariables(undeclared_variables));
+    }
```
sigilioso, line 60: super-nit: we could keep the variables scoped.
```suggestion
    if let unknown_ns = unknown_namespaces(&references) && !unknown_ns.is_empty() {
        return Err(ValidationError::UnknownNamespaces(unknown_ns));
    }

    let declared_variables: BTreeSet<String> = definition.variables.flatten().into_keys().collect();
    if let undeclared = undeclared_variables(&references) && !undeclared.is_empty() {
        return Err(ValidationError::UndeclaredVariables(undeclared));
    }
```

agent-control/src/agent_type/validation.rs
```diff
@@ -0,0 +1,412 @@
...
+    Definition(#[from] AgentTypeDefinitionParseError),
+    #[error(
+        "undeclared variable(s) referenced in deployment: {}",
+        .0.iter().map(String::as_str).collect::<Vec<_>>().join(", ")
+    )]
+    UndeclaredVariables(BTreeSet<String>),
+    #[error(
+        "unknown namespace(s) referenced in deployment: {}",
+        .0.iter().map(String::as_str).collect::<Vec<_>>().join(", ")
+    )]
```
sigilioso, line 24:
```suggestion
    #[error(
        "unknown namespace(s) referenced in variables in deployment: {}",
        .0.iter().map(String::as_str).collect::<Vec<_>>().join(", ")
    )]
```

agent-control/src/agent_type/variable/namespace.rs
```diff
@@ -28,6 +28,17 @@ pub enum Namespace {
...
+    /// All supported namespace variants.
+    pub const ALL: [Namespace; 7] = [
+        Namespace::Variable,
+        Namespace::SubAgent,
+        Namespace::AgentControl,
+        Namespace::EnvironmentVariable,
+        Namespace::Vault,
+        Namespace::File,
+        Namespace::K8sSecret,
+    ];
```
sigilioso, line 40: Nit: I'd add/extend the unit-tests to make things fail if we ever forget to add a variant in the `ALL` list.

sigilioso (approved):
I've left some _nits_ that can be ignored.

LGTM

agent-control/src/agent_type/variable/namespace.rs
gsanchezgavier: ohh i actually wanted to make this more robust so it forces you to keep it up to date and forgot, thanks!

agent-control/src/agent_type/validation.rs
gsanchezgavier: tried it gets a bit harder to read IMO

agent-control/src/agent_type/validation.rs
gsanchezgavier: i'll add it in follow up PR

agent-control/src/agent_type/variable/namespace.rs
gsanchezgavier: will add it to a follow up PR with some extra refactor

PR #2729, opened 2026-07-23
feat(types): added types and test

paologallinaharbur:
Uploading apache and nginx and adding a common test for all ohis without service

### DIFF
--- a/agent-control/src/agent_type/registry/local/agent_type_validation_tests.rs
+++ b/agent-control/src/agent_type/registry/local/agent_type_validation_tests.rs
@@ -719,6 +719,102 @@ static AGENT_TYPE_REDIS: LazyLock<AgentTypeValuesTestCase> =
         ..Default::default()
     });
 
+static AGENT_TYPE_NGINX: LazyLock<AgentTypeValuesTestCase> =
+    LazyLock::new(|| AgentTypeValuesTestCase {
+        agent_type: "newrelic/com.newrelic.infrastructure.nri_nginx:0.1.0",
+        values_linux: AgentTypeValues {
+            cases: HashMap::from([
+                (
+                    "mandatory fields only",
+                    r#"
+                config: "integrations: []"
+                version: "v1.15.2"
+                "#,
+                ),
+                (
+                    "check all value types are correct",
+                    r#"
+                config: "integrations: []"
+                version: "v1.15.2"
+                oci.repository: "newrelic/nri-nginx"
+                "#,
+                ),
+            ]),
+            ..Default::default()
+        }
+        .into(),
+        values_windows: AgentTypeValues {
+            cases: HashMap::from([
+                (
+                    "mandatory fields only",
+                    r#"
+                config: "integrations: []"
+                version: "v1.15.2"
+                "#,
+                ),
+                (
+                    "check all value types are correct",
+                    r#"
+                config: "integrations: []"
+                version: "v1.15.2"
+                oci.repository: "newrelic/nri-nginx"
+                "#,
+                ),
+            ]),
+            ..Default::default()
+        }
+        .into(),
+        ..Default::default()
+    });
+
+static AGENT_TYPE_APACHE: LazyLock<AgentTypeValuesTestCase> =
+    LazyLock::new(|| AgentTypeValuesTestCase {
+        agent_type: "newrelic/com.newrelic.infrastructure.nri_apache:0.1.0",
+        values_linux: AgentTypeValues {
+            cases: HashMap::from([
+                (
+                    "mandatory fields only",
+                    r#"
+                config: "integrations: []"
+                version: "v1.15.2"
+                "#,
+                ),
+                (
+                    "check all value types are correct",
+                    r#"
+                config: "integrations: []"
+                version: "v1.15.2"
+                oci.repository: "newrelic/nri-apache"
+                "#,
+                ),
+            ]),
+            ..Default::default()
+        }
+        .into(),
+        values_windows: AgentTypeValues {
+            cases: HashMap::from([
+                (
+                    "mandatory fields only",
+                    r#"
+                config: "integrations: []"
+                version: "v1.15.2"
+                "#,
+                ),
+                (
+                    "check all value types are correct",
+                    r#"
+                config: "integrations: []"
+                version: "v1.15.2"
+                oci.repository: "newrelic/nri-apache"
+                "#,
+                ),
+            ]),
+            ..Default::default()
+        }
+        .into(),
+        ..Default::default()
+    });
+
 fn get_agent_type_test_cases() -> impl Iterator<Item = &'static AgentTypeValuesTestCase> {
     [
         &AGENT_TYPE_APM_DOTNET,
@@ -737,6 +833,8 @@ fn get_agent_type_test_cases() -> impl Iterator<Item = &'static AgentTypeValuesT
         &AGENT_TYPE_PIPELINE_CONTROL_GATEWAY_CONFIG,
         &AGENT_TYPE_EBPF,
         &AGENT_TYPE_REDIS,
+        &AGENT_TYPE_NGINX,
+        &AGENT_TYPE_APACHE,
     ]
     .into_iter()
     .map(Deref::deref)
--- a/test/e2e-runner/src/common/ohi.rs
+++ b/test/e2e-runner/src/common/ohi.rs
@@ -117,9 +117,14 @@ pub fn get_all_ohi_to_test(args: &InstallationArgs) -> Vec<Ohi> {
     ]
 }
 
+pub const TEST_LABEL: &str = "test.label";
+pub const TEST_LABEL_VALUE: &str = "1.2.3";
+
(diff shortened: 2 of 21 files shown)


test/e2e-runner/src/common/ohi.rs
```diff
@@ -59,3 +62,144 @@ pub fn check_ohi_shared_filesystem(
...
+    //     .expect("--apache-version is required for this scenario");
+
+    vec![
+        Ohi {
+            name: "nri-redis",
+            agent_type_id: "newrelic/com.newrelic.infrastructure.nri_redis:0.1.0",
+            repo: "newrelic/newrelic-agent-control-redis-dev",
+            version: redis_version,
+        },
+        // Ohi {
```
paologallinaharbur: I still need to uplaod this packages

test/e2e-runner/src/common/ohi.rs
```diff
@@ -59,3 +62,144 @@ pub fn check_ohi_shared_filesystem(
...
+version: {DEV_INFRA_AGENT_VERSION}
+oci:
+  repository: {DEV_INFRA_AGENT_REPO}
+"#
+        ),
+    );
+
+    // Per-OHI local config: a single integration entry. No real
+    // service target — each OHI binary will fail its health check, but the infra-agent still logs
+    // the invocation with the label, which is what we assert on.
```
sigilioso: I'd simplify the comment, saying that "each OHI binary will fail its health check" can lead to confusion

test/e2e-runner/src/common/ohi.rs
```diff
@@ -59,3 +62,144 @@ pub fn check_ohi_shared_filesystem(
...
+        write_agent_local_config(
+            &config_path,
+            format!(
+                r#"
+config:
+  integrations:
+    - name: {}
+      interval: 15s
+      labels:
+        test.label: 1.2.3
```
sigilioso: Would it make sense to use a constant and reference it from here and the log check?

test/e2e-runner/src/linux/scenarios/all_ohis_no_service.rs
```diff
@@ -0,0 +1,54 @@
...
+use tracing::info;
+
+/// Runs a generic OHI e2e scenario for every declared OHI at once, without starting any monitored
+/// service.
+///
+/// Installs Agent Control with the infra-agent and every OHI sub-agent side-by-side, wires each
+/// one's config, and then verifies from the Agent Control logs that
+/// the infra-agent invoked each integration binary with the expected config. Since no real
+/// services are running, the integrations will not succeed at monitoring anything — the test only
+/// checks that the binaries were triggered with the expected configs.
```
sigilioso: I'd also simplify this comment: the _expected config_ is merely `test.label=1.2.3`, right?

test/e2e-runner/src/linux/scenarios/all_ohis_no_service.rs
```diff
@@ -0,0 +1,54 @@
...
+        60,
+        Duration::from_secs(2),
+        "every OHI invocation in AC log with expected label",
+        || {
+            for ohi in &ohis {
+                let name_needle = format!("integration_name={}", ohi.name);
+                expect_log_line_contains(
+                    linux::DEFAULT_LOG_PATH,
+                    &[name_needle.as_str(), "test.label=1.2.3"],
+                )?;
```
sigilioso, line 50: AC requires a specific log level to show the infra-agent logs, right?

Could we add a comment stating it just to avoid digging in the future if we ever decide to change the log level in e2e?

test/e2e-runner/src/linux/scenarios/all_ohis_no_service.rs
```diff
@@ -0,0 +1,54 @@
...
+    retry_panic(
+        60,
+        Duration::from_secs(2),
+        "every OHI invocation in AC log with expected label",
+        || {
+            for ohi in &ohis {
+                let name_needle = format!("integration_name={}", ohi.name);
+                expect_log_line_contains(
+                    linux::DEFAULT_LOG_PATH,
+                    &[name_needle.as_str(), "test.label=1.2.3"],
```
sigilioso: Do the infra-agent lines have some sort of prefix? If would be nice to know that these lines are actually written by the infra-agent.

sigilioso (commented):
I like the approach! I've left some questions

test/e2e-runner/src/linux/scenarios/all_ohis_no_service.rs
paologallinaharbur: I think it does, let me check that!

test/e2e-runner/src/linux/scenarios/all_ohis_no_service.rs
paologallinaharbur: sure, I'll add a comment about that

test/e2e-runner/src/common/ohi.rs
paologallinaharbur: why not!

paologallinaharbur:
> I like the approach! I've left some questions

eheh I think it was the approach you proposed, it is an easy win 🤣

test/e2e-runner/src/main.rs
```diff
@@ -95,6 +99,10 @@ enum WindowsScenarios {
     /// Installs Agent Control with a mirrored infra-agent OCI and a custom nri-redis OCI,
     /// spins up a local Redis instance, and verifies that RedisSample data lands in NRDB.
     NriRedis(InstallationArgs),
+    /// Installs Agent Control with the infra-agent and every OHI sub-agent side-by-side (nri-nginx,
```
danielorihuela: Nit: Don't mention nri-nginx and nri-apache, this will change shortly

test/e2e-runner/src/windows.rs
```diff
@@ -16,7 +16,7 @@ const AGENT_CONTROL_DIRS: &[&str] = &[
     r"C:\ProgramData\New Relic\newrelic-agent-control\",
 ];
 
-const DEFAULT_AC_CONFIG_PATH: &str =
+pub(crate) const DEFAULT_AC_CONFIG_PATH: &str =
```
danielorihuela: Nit: `pub` is enough I think. It will aso be consistent with the linux counterpart with only uses `pub`

test/e2e-runner/src/main.rs
paologallinaharbur: sure! makes sense!

test/e2e-runner/src/linux/scenarios/all_ohis_no_service.rs
paologallinaharbur: it is not a "real prefix" but I added a line that comes only from the infra agent

agent-control/src/agent_type/registry/local/agent_type_validation_tests.rs
```diff
@@ -719,6 +719,102 @@ static AGENT_TYPE_REDIS: LazyLock<AgentTypeValuesTestCase> =
...
+static AGENT_TYPE_NGINX: LazyLock<AgentTypeValuesTestCase> =
+    LazyLock::new(|| AgentTypeValuesTestCase {
+        agent_type: "newrelic/com.newrelic.infrastructure.nri_nginx:0.1.0",
+        values_linux: AgentTypeValues {
+            cases: HashMap::from([
+                (
+                    "mandatory fields only",
+                    r#"
+                config: "integrations: []"
+                version: "v1.15.2"
```
danielorihuela, line 731: the latest nri-nginx version is other https://github.com/newrelic/nri-nginx/releases

agent-control/src/agent_type/registry/local/agent_type_validation_tests.rs
```diff
@@ -719,6 +719,102 @@ static AGENT_TYPE_REDIS: LazyLock<AgentTypeValuesTestCase> =
...
+static AGENT_TYPE_APACHE: LazyLock<AgentTypeValuesTestCase> =
+    LazyLock::new(|| AgentTypeValuesTestCase {
+        agent_type: "newrelic/com.newrelic.infrastructure.nri_apache:0.1.0",
+        values_linux: AgentTypeValues {
+            cases: HashMap::from([
+                (
+                    "mandatory fields only",
+                    r#"
+                config: "integrations: []"
+                version: "v1.15.2"
```
danielorihuela, line 779: The latest nri-apache version is another https://github.com/newrelic/nri-apache/releases

sigilioso (approved):
LGTM

agent-control/src/agent_type/registry/local/agent_type_validation_tests.rs
paologallinaharbur: These are going to be outdated super soon anyway, I actually agree with you that we should remove these uni tests

agent-control/src/agent_type/registry/local/agent_type_validation_tests.rs
paologallinaharbur: as above!

PR #2730, opened 2026-07-23
feat: add nri-postgresql agent-type and tests

alvarocabanas:
add nri-postgresql agent-type and tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Remember that the keywords that you can use are the following:
 
 ### enhancement
 - On-host: added apache and nginx agentTypes
+- On-host: added nri-postgresql agentType.
 - Added `agent-type validate --file <path>` subcommand to `newrelic-agent-control-cli` and `newrelic-agent-control-k8s-cli` for schema-level validation of agent type definition files.
 - Remove static regex validator for newrelic-infra Agent Type.
 
--- a/agent-control/src/agent_type/registry/local/agent_type_validation_tests.rs
+++ b/agent-control/src/agent_type/registry/local/agent_type_validation_tests.rs
@@ -815,6 +815,54 @@ static AGENT_TYPE_APACHE: LazyLock<AgentTypeValuesTestCase> =
         ..Default::default()
     });
 
+static AGENT_TYPE_POSTGRESQL: LazyLock<AgentTypeValuesTestCase> =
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ Remember that the keywords that you can use are the following:
 - On-host: added support for nri-redis agentType for both Linux and Windows.
 - On-host health config: replaced the flat `health.http:` / `health.file:` fields with an explicit `checks:` list. Each entry is discriminated by `kind:` (`Process`, `Http`, or `File`). The previously implicit supervised-process health check is now declared as `kind: Process`. An omitted or empty `checks:` list disables health reporting entirely.
 - On-host filesystem entries (`file`/`dir`) now always survive sub-agent stop, restart, and config-apply. A stray `persistent:` key in existing agent-type YAML is ignored rather than rejected.
+- nri-postgresql agent type support.
```
danielorihuela: we don't have to modify sections of the changelog that have already been released

agent-control/src/agent_type/registry/local/agent_type_validation_tests.rs
```diff
@@ -815,6 +815,54 @@ static AGENT_TYPE_APACHE: LazyLock<AgentTypeValuesTestCase> =
...
+static AGENT_TYPE_POSTGRESQL: LazyLock<AgentTypeValuesTestCase> =
+    LazyLock::new(|| AgentTypeValuesTestCase {
+        agent_type: "newrelic/com.newrelic.infrastructure.nri_postgresql:0.1.0",
+        values_linux: AgentTypeValues {
+            cases: HashMap::from([
+                (
+                    "mandatory fields only",
+                    r#"
+                config: "integrations: []"
+                version: "v1.15.2"
```
danielorihuela, line 827: this is not the latest version of https://github.com/newrelic/nri-postgresql/releases.

CHANGELOG.md
alvarocabanas: It's fixed.

agent-control/src/agent_type/registry/local/agent_type_validation_tests.rs
alvarocabanas: No, and it isn't also for nri-nginx and other because it's not needed for this test, this is just unit test that doesn't require a real version.

danielorihuela (approved):

PR #2731, opened 2026-07-24
feat: add new check for binary size in prs

vjripoll:
## Summary

Adds a binary size check for Agent Control: a PR comment that tracks size against `main`, plus reporting and alerting for nightly/release builds.

- **PR check**: measures `newrelic-agent-control` / `newrelic-agent-control-cli` on every PR build and posts a sticky comment with size and Δ vs the last successful build on `main` (bytes/% delta). Purely informational, doesn't block merge. Falls back gracefully to "no baseline found" when there's nothing to compare against yet (first run after enabling this, or the baseline artifact expired).
- **Nightly/release reporting**: every built binary (control + CLI, all platforms) is reported as an `AgentControlStats` custom event to New Relic on `nightly.yml` and `prerelease.yml` runs, queryable via NRQL.
- **Growth alert**: before reporting, looks up the previous build's size for that same binary+target via NerdGraph. If it grew more than 10%, warns on Slack by reusing the existing `component_send_warning_via_slack.yml`.

## Changes

- `.github/workflows/push_pr_checks_tests.yml` binary size measurement, baseline artifact (published on push to `main`, downloaded on PRs), sticky PR comment.
- `.github/workflows/scripts/report_binary_size.sh` new script: reports size to the Events API, checks growth via NerdGraph, writes a warning file if >10%.
- `.github/workflows/component_packages.yml` new step calling the script per binary; new `notify-binary-size-growth` job that warns on Slack when triggered.
- `.github/workflows/component_send_warning_via_slack.yml` accepts an explicit `slack_webhook_url` secret.

### DIFF
--- a/.github/workflows/component_packages.yml
+++ b/.github/workflows/component_packages.yml
@@ -22,6 +22,18 @@ on:
       pfx_passphrase:
         description: 'Passphrase of the PFX certificate'
         required: false
+      nr_account_id:
+        description: 'New Relic account ID for binary size reporting (optional)'
+        required: false
+      nr_license_key:
+        description: 'New Relic license key, Events API ingest key (optional)'
+        required: false
+      nr_user_api_key:
+        description: 'New Relic User API key for the growth check via NerdGraph (optional)'
+        required: false
+      slack_webhook_url:
+        description: 'Slack webhook for binary size growth warnings (optional)'
+        required: false
     inputs:
       pre-release:
         description: 'set to true if running a real pre-release'
@@ -41,6 +53,8 @@ jobs:
   build:
     runs-on: ubuntu-latest
     name: Build and upload
+    outputs:
+      binary-size-warnings: ${{ steps.binary-size.outputs.warnings }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
@@ -117,6 +131,29 @@ jobs:
           GORELEASER_CURRENT_TAG: ${{ inputs.tag_name }}
           SKIP_WINDOWS_SIGN: ${{ env.SKIP_WINDOWS_SIGN }}
 
+      - name: Report binary sizes to New Relic
+        id: binary-size
+        env:
+          BINARY_VERSION: ${{ inputs.tag_name }}
+          NR_ACCOUNT_ID: ${{ secrets.nr_account_id }}
+          NR_LICENSE_KEY: ${{ secrets.nr_license_key }}
+          NR_USER_API_KEY: ${{ secrets.nr_user_api_key }}
+          WARNINGS_FILE: binary-size-warnings.txt
+        run: |
+          : > "$WARNINGS_FILE"
+          for path in dist/*/newrelic-agent-control dist/*/newrelic-agent-control-cli dist/*/newrelic-agent-control.exe dist/*/newrelic-agent-control-cli.exe; do
+            [ -f "$path" ] || continue
+            export BINARY_PATH="$path"
+            bash .github/workflows/scripts/report_binary_size.sh
+          done
+          if [ -s "$WARNINGS_FILE" ]; then
+            {
+              echo "warnings<<EOF"
+              cat "$WARNINGS_FILE"
+              echo "EOF"
+            } >> "$GITHUB_OUTPUT"
+          fi
+
       - name: Upload assets to pipeline
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
         with:
@@ -125,6 +162,22 @@ jobs:
           path: |
             ./bin/*
             ./dist/*
+
+  notify-binary-size-growth:
+    name: Warn on binary size growth
+    needs: build
+    # Keep this event allow-list in sync with TRACKED_TRIGGER_EVENTS in report_binary_size.sh:
+    # only nightly (schedule/workflow_dispatch) and real pre-releases (release) build comparable
+    # binaries. Explicit here too so a PR run can never trigger this, even if warnings weren't empty.
+    if: |
+      needs.build.outputs.binary-size-warnings != '' &&
+      contains(fromJSON('["schedule", "workflow_dispatch", "release"]'), github.event_name)
+    continue-on-error: true
+    uses: ./.github/workflows/component_send_warning_via_slack.yml
+    with:
+      message: ${{ needs.build.outputs.binary-size-warnings }}
+    secrets:
+      slack_webhook_url: ${{ secrets.slack_webhook_url }}
 
   verify-windows-signatures:
     runs-on: windows-latest
--- a/.github/workflows/scripts/report_binary_size.sh
+++ b/.github/workflows/scripts/report_binary_size.sh
@@ -1,0 +1,100 @@
+#!/bin/bash
+# Reports one AgentControlStats event to New Relic; appends a warning to WARNINGS_FILE
+# if the binary grew >10%. Sending that warning to Slack is the caller's job, not this script's.
+#
+# BINARY_NAME and BINARY_TARGET are derived from BINARY_PATH, assuming the goreleaser
+# layout dist/<build-id>_<target>/<binary>.
+#
+# Required env: BINARY_PATH, BINARY_VERSION, NR_ACCOUNT_ID, NR_LICENSE_KEY
+# Optional env: NR_USER_API_KEY (NerdGraph, skips growth check if unset), WARNINGS_FILE (default: binary-size-warnings.txt)
+
+set -euo pipefail
+
+GROWTH_THRESHOLD_PCT=10
+
+if [[ -z "${BINARY_PATH:-}" ]]; then
+  echo "BINARY_PATH is required" >&2
+  exit 1
+fi
+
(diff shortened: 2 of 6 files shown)


.github/workflows/component_packages.yml
```diff
@@ -117,6 +131,32 @@ jobs:
...
+          NR_USER_API_KEY: ${{ secrets.nr_user_api_key }}
+          WARNINGS_FILE: binary-size-warnings.txt
+        run: |
+          : > "$WARNINGS_FILE"
+          for path in dist/*/newrelic-agent-control dist/*/newrelic-agent-control-cli dist/*/newrelic-agent-control.exe dist/*/newrelic-agent-control-cli.exe; do
+            [ -f "$path" ] || continue
+            dir=$(basename "$(dirname "$path")")
+            export BINARY_NAME=$(basename "$path" .exe)
+            export BINARY_TARGET="${dir#*_}"
+            export BINARY_PATH="$path"
```
danielorihuela: Why don't we just pass the `BINARY_PATH` and compute the rest internally?

.github/workflows/scripts/report_binary_size.sh
```diff
@@ -0,0 +1,77 @@
...
+#
+# Required env: BINARY_NAME, BINARY_TARGET, BINARY_PATH, BINARY_VERSION, NR_ACCOUNT_ID, NR_LICENSE_KEY
+# Optional env: NR_USER_API_KEY (NerdGraph, skips growth check if unset), WARNINGS_FILE (default: binary-size-warnings.txt)
+
+set -euo pipefail
+
+GROWTH_THRESHOLD_PCT=10
+
+if [[ -z "${NR_ACCOUNT_ID:-}" || -z "${NR_LICENSE_KEY:-}" ]]; then
+  echo "NR_ACCOUNT_ID/NR_LICENSE_KEY not set, skipping binary size report for ${BINARY_PATH:-unknown}"
```
danielorihuela: `BINARY_PATH` is mandatory. Should we check it's passed as well on a different `if`, at the top?

.github/workflows/component_packages.yml
```diff
@@ -126,13 +166,24 @@ jobs:
...
+  notify-binary-size-growth:
+    name: Warn on binary size growth
+    needs: build
+    if: needs.build.outputs.binary-size-warnings != ''
+    continue-on-error: true
+    uses: ./.github/workflows/component_send_warning_via_slack.yml
+    with:
+      message: ${{ needs.build.outputs.binary-size-warnings }}
+    secrets:
+      slack_webhook_url: ${{ secrets.slack_webhook_url }}
```
danielorihuela, line 180: Why not use secrets inherit and avoid changes on the component?
I'm not against it, but curious about the reasoning.

vjripoll: nr_account_id, nr_license_key, gpg_*, and pfx_* are already passed one by one, and neither this workflow or nightly/prerelease use inherit anywhere else. Switching would break that pattern just to save two lines.

It would also give component_send_warning_via_slack.yml access to every secret this workflow currently has the GPG signing key, the PFX cert, the license key even though it only needs the webhook URL. That's a lot of extra secrets exposed for no real reason.

danielorihuela: > nr_account_id, nr_license_key, gpg_, and pfx_ are already passed one by one, and neither this workflow or nightly/prerelease use inherit anywhere else. Switching would break that pattern just to save two lines.

The nightly uses

```
notify-failure:
    if: ${{ always() && failure() }}
    needs:
      - onhost-e2e
      - k8s_canaries
      - onhost_canaries
      - security-image
      - security-source-code
      - build-image
      - build-packages
      - k8s-e2e-tests
    uses: ./.github/workflows/component_send_warning_via_slack.yml
    with:
      message: "Nightly workflow failed"
    secrets: inherit
```

> It would also give component_send_warning_via_slack.yml access to every secret this workflow currently has the GPG signing key, the PFX cert, the license key even though it only needs the webhook URL. That's a lot of extra secrets exposed for no real reason.

Okey, I see the issue. It's not a security issue though, because we control the other component and what secrets we use. Would it make sense to keep it consistent and inherit all secrets? Then in a different PR, replace all calls to this component and just pass the needed secret? Should we then also change `component_image` or other workflows to just pass in the required secrets?

danielorihuela: Oh, seems like a security issue 

https://cheatsheetseries.owasp.org/cheatsheets/GitHub_Actions_Security_Cheat_Sheet.html#eliminate-secrets-inherit-while-reusing-workflows
https://cycode.com/blog/github-actions-supply-chain-attack/

danielorihuela: I think we need a sustain task. Your call

* Use secrets inherit for consistency + create security task
* Leave as it is + create security task

.github/workflows/scripts/report_binary_size.sh
```diff
@@ -0,0 +1,89 @@
...
+
+curl -s -X POST \
+  "https://insights-collector.newrelic.com/v1/accounts/${NR_ACCOUNT_ID}/events" \
+  -H "Content-Type: application/json" \
+  -H "Api-Key: ${NR_LICENSE_KEY}" \
+  -d "$event"
+
+if [[ -n "$PREVIOUS_SIZE_BYTES" ]]; then
+  growth_pct=$(awk -v cur="$SIZE_BYTES" -v prev="$PREVIOUS_SIZE_BYTES" \
+    'BEGIN { printf "%.2f", (prev == 0) ? 0 : ((cur - prev) * 100.0 / prev) }')
```
danielorihuela: Should we print something if the prev size bytes is 0?

It's not the same going from 21 bytes to 21 bytes than going from 0 bytes to 21 bytes. The first case is legitimate, the second seems like an error.

.github/workflows/scripts/report_binary_size.sh
```diff
@@ -0,0 +1,89 @@
...
+
+SIZE_BYTES=$(stat -c%s "$BINARY_PATH")
+
+# GITHUB_HEAD_REF is set for PRs; GITHUB_REF_NAME covers push/schedule/dispatch.
+BRANCH="${GITHUB_HEAD_REF:-${GITHUB_REF_NAME:-}}"
+
+# Fetch the previous size before reporting the current one, so we don't race our own insert.
+PREVIOUS_SIZE_BYTES=""
+if [[ -n "${NR_USER_API_KEY:-}" ]]; then
+  nrql_query="SELECT sizeBytes FROM AgentControlStats WHERE binaryName = '${BINARY_NAME}' AND target = '${BINARY_TARGET}' SINCE 90 days ago ORDER BY timestamp DESC LIMIT 1"
```
danielorihuela, line 46: This query is getting the last `AgentControlStats`, but we are not using `triggerEvent` nor `"$GITHUB_EVENT_NAME"` anywhere.

I don't understand how we make sure that we compare a `nightly` size with the previous `nightly` size.

.github/workflows/component_packages.yml
```diff
@@ -117,6 +131,29 @@ jobs:
           GORELEASER_CURRENT_TAG: ${{ inputs.tag_name }}
           SKIP_WINDOWS_SIGN: ${{ env.SKIP_WINDOWS_SIGN }}
 
+      - name: Report binary sizes to New Relic
```
danielorihuela, line 134: This file is called on `component_onhost_e2e`, which is called on:

* nightly
* prerelease
* push_pr_test_extended_labels
* push_pr_test_extended_paths_host

Does this mean that we will also get a slack message when the check fails on a PR executing onhost e2e tests? I think it does and we don't want that.

danielorihuela: If that's true, one solution I can think of is to just rely on the nightly slack message we already send on error. Then, add something similar to prerelease.

```
notify-failure:
    if: ${{ always() && failure() }}
    needs:
      - onhost-e2e
      - k8s_canaries
      - onhost_canaries
      - security-image
      - security-source-code
      - build-image
      - build-packages
      - k8s-e2e-tests
    uses: ./.github/workflows/component_send_warning_via_slack.yml
    with:
      message: "Nightly workflow failed"
    secrets: inherit
```

.github/workflows/scripts/report_binary_size.sh
vjripoll: we now log explicitly when the previous size is 0 and skip the growth check, instead of silently treating 0→N as 0% growth

.github/workflows/scripts/report_binary_size.sh
vjripoll: only nightly (schedule/workflow_dispatch) and real pre-releases (release) ever populate AgentControlStats every other caller leaves nr_account_id/nr_license_key unset, so the script exits before reporting. Both build the binary identically, so they're comparable. Added an explicit GITHUB_EVENT_NAME allow-list in the script so this isn't just implicit.

.github/workflows/component_packages.yml
vjripoll: component_onhost_e2e.yml never forwards the NR secrets to this workflow, so the script exits before writing any warning no Slack fire today on PR runs. Also added an explicit github.event_name check on the notify-binary-size-growth job itself, so the protection isn't just implicit secret-wiring.

PR #2744, opened 2026-07-28
chore(redis): point to prod in tests

paologallinaharbur:
Changing tests to point to prod


danielorihuela (dismissed):

sigilioso:
We need to update the [all-ohis-no-service](https://github.com/newrelic/newrelic-agent-control/blob/51e0e9ecebdad48ef9e7906d45243e535d015a07/test/e2e-runner/src/common/ohi.rs#L67) test as well, right?

paologallinaharbur:
>We need to update the [all-ohis-no-service](https://github.com/newrelic/newrelic-agent-control/blob/51e0e9ecebdad48ef9e7906d45243e535d015a07/test/e2e-runner/src/common/ohi.rs#L67) test as well, right?

yes, but I cannot point to prod there yet, since different agents cannot point to different registries!

test/e2e-runner/src/common/ohi.rs
```diff
@@ -99,12 +95,14 @@ pub fn get_all_ohi_to_test(args: &InstallationArgs) -> Vec<Ohi> {
...
 
+    // TODO once all integrations are available we get migrate the whole test to prod, for the time being
+    // we need to keep leveraging the dev package
     vec![
         Ohi {
             name: "nri-redis",
             agent_type_id: "newrelic/com.newrelic.infrastructure.nri_redis:0.1.0",
             repo: "newrelic/newrelic-agent-control-redis-dev",
-            version: redis_version,
+            version: "0.0.1".into(),
```
danielorihuela, line 105: why this change?

sigilioso: It seems a temporal workaround for this: https://github.com/newrelic/newrelic-agent-control/pull/2744#issuecomment-5104870659. I'd suggest adding a comment / TODO if we hadn't a task to update all of them anyway.

paologallinaharbur: since the redis version tag available in the custom repo is different from the one in the official one

paologallinaharbur: only the 0.0.1 exist

sigilioso (approved):

test/e2e-runner/src/common/ohi.rs
danielorihuela: Aaaaah okay. I was confused with one test using the prod version while the other didn't. Got it. Thanks!!!

danielorihuela (approved):

PR #2745, opened 2026-07-28
chore: introduce strum for enum iteration and parsing

gsanchezgavier:
## Summary

Replaces hand-written enum helpers (`FromStr`, `Namespace::ALL`, `Environment::all()`) with `strum` derives. Reduces boilerplate and makes variant sets compiler-enforced.

While refactoring i noticed that there were dead `Deserialize`/`deserialize_with` attributes from the backoff types (parsing actually flows through `TemplateableValue`), so they are removed too.

## Changes

- Add workspace `strum` 0.28.0 dependency
- `BackoffStrategyType`: derive `EnumString` instead of a manual `FromStr` + drop `Deserialize`
- `Namespace`/`Environment`: derive `EnumIter`, remove `Namespace::ALL` and the test-only `Environment::all()`
- Rework `Namespace::is_secret_variable` around an exhaustive `is_secret` match (identical secret set)
- Add a deserialize-and-resolve test for the backoff strategy config

## Notes for reviewers

- `AgentTypeError::UnknownBackoffStrategyType` was already unreachable (errors funnel to `ValueNotParseableFromString` via `TemplateableValue`); its removal is safe.
- `strum` 0.27.2 remains in the tree transitively, so the lockfile now carries two `strum` majors.

Follow up from #2727

### DIFF
--- a/agent-control/src/agent_type/variable/namespace.rs
+++ b/agent-control/src/agent_type/variable/namespace.rs
@@ -1,12 +1,13 @@
 //! Namespaces used to prefix agent type variable names (e.g. `nr-var`, `nr-env`, `nr-vault`).
 use std::fmt::Display;
+use strum::{EnumIter, IntoEnumIterator};
 
 /// Holds the variable name prefixed with the namespace.
 /// Example: "nr-env:MY_ENV_VAR" for the environment variable "MY_ENV_VAR".
 pub type NamespacedVariableName = String;
 
 /// Namespace defines the supported namespace names for variables definition.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, EnumIter)]
 pub enum Namespace {
     /// Variables defined in the agent type.
     Variable,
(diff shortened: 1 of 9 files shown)


agent-control/src/agent_type/variable/namespace.rs
```diff
@@ -65,14 +55,20 @@ impl Namespace {
...
+
+    /// Whether this namespace holds "secret" variables (loaded on every remote config fetch).
+    fn is_secret(&self) -> bool {
+        match self {
+            Namespace::Variable | Namespace::SubAgent | Namespace::AgentControl => false,
+            Namespace::EnvironmentVariable
+            | Namespace::Vault
+            | Namespace::File
+            | Namespace::K8sSecret => true,
+        }
```
DavSanchez, line 71: I think a `matches!` with only the `true` cases should also work.

DavSanchez (approved):
What we have to do for using a language without [`Enum`](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Prelude.html#t:Enum) and [`Bounded`](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Prelude.html#t:Bounded) 😅

agent-control/src/agent_type/variable/namespace.rs
danielorihuela: there's also https://docs.rs/strum/latest/strum/trait.EnumProperty.html

sigilioso (approved):
LGTM

agent-control/src/agent_type/variable/namespace.rs
gsanchezgavier: I did it this way to force the new variant implementor take a decision on this, so there is no default behaviour of a variant. I might have understood wrong but only matching true cases will make false the default right?

DavSanchez: ah yes if we want this to stop compiling if a new variant is added (until a decision of `true`/`false` is needed then I'd say this is the way to go 👌  💯

PR #2746, opened 2026-07-28
feat: enable cache for Agent Type remote registry

sigilioso:
This PR wires up the cache implementation for the remote Agent Type registry. 

It also introduces an integration test to check that the registry built by the Agent Control Runner is resolving embedded, is resolving remote ant that the cache is in place. The integration test involved a small refactor of the function building the registry: it was made public (to be callable from integration tests) and the arguments were better scoped.

### DIFF
--- a/agent-control/src/agent_type/registry.rs
+++ b/agent-control/src/agent_type/registry.rs
@@ -9,6 +9,7 @@ use std::path::PathBuf;
 use thiserror::Error;
 use tracing::{debug, warn};
 
+use self::caching::CachingRegistry;
 use self::local::LocalRegistry;
 use self::remote::RemoteRegistry;
 use super::agent_type_id::AgentTypeID;
@@ -83,8 +84,9 @@ impl<R: AgentTypeRegistry> Registry<R> {
 pub enum SupportedRegistry {
     /// In-memory registry of embedded and custom agent types.
     Local(LocalRegistry),
-    /// Registry that pulls agent types from a remote OCI registry.
-    Remote(RemoteRegistry<OCIAgentTypeArtifactDownloader>),
+    /// Registry that pulls agent types from a remote OCI registry, memoizing successful lookups
+    /// since a downloaded artifact is immutable for a given [AgentTypeID].
+    Remote(CachingRegistry<RemoteRegistry<OCIAgentTypeArtifactDownloader>>),
 }
 
 impl AgentTypeRegistry for SupportedRegistry {
@@ -108,7 +110,7 @@ impl Registry<SupportedRegistry> {
         downloader: OCIAgentTypeArtifactDownloader,
     ) -> Self {
         let local = LocalRegistry::new(env, config.dynamic_agent_types_path);
-        let remote = RemoteRegistry::new(env, downloader);
+        let remote = CachingRegistry::new(RemoteRegistry::new(env, downloader));
         Self::new(vec![
             SupportedRegistry::Local(local),
             SupportedRegistry::Remote(remote),
--- a/agent-control/tests/on_host/agent_type_remote_registry.rs
+++ b/agent-control/tests/on_host/agent_type_remote_registry.rs
@@ -1,7 +1,10 @@
 use crate::common::runtime::tokio_runtime;
 use crate::on_host::tools::oci_package_manager::TestDataHelper;
 use assert_matches::assert_matches;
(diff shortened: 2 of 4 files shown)


agent-control/src/agent_type/registry.rs
```diff
@@ -83,8 +84,9 @@ impl<R: AgentTypeRegistry> Registry<R> {
 pub enum SupportedRegistry {
     /// In-memory registry of embedded and custom agent types.
     Local(LocalRegistry),
-    /// Registry that pulls agent types from a remote OCI registry.
-    Remote(RemoteRegistry<OCIAgentTypeArtifactDownloader>),
+    /// Registry that pulls agent types from a remote OCI registry, memoizing successful lookups
+    /// since a downloaded artifact is immutable for a given [AgentTypeID].
+    Remote(CachingRegistry<RemoteRegistry<OCIAgentTypeArtifactDownloader>>),
```
paologallinaharbur, line 89: I assumed we had already caching in place, do you know why it was disabled?

sigilioso: @gsanchezgavier and I discussed about it and It seems we just forgot to wire it up

gsanchezgavier (approved):
thanks!

agent-control/tests/on_host/agent_type_remote_registry.rs
```diff
@@ -148,3 +208,50 @@ fn test_remote_registry_errors_on_missing_agent_type_with_oci_registry() {
         Err(AgentTypeRegistryError::Remote(_))
     );
 }
+
+#[test]
+#[ignore = "needs oci registry (use *with_oci_registry suffix)"]
+fn test_build_agent_type_registry_resolves_embedded_and_caches_remote_with_oci_registry() {
```
danielorihuela, line 214: Do we need a whole integration test with OCI, when a unit test using mocks similar to `test_no_layer_hit` would suffice?

agent-control/tests/on_host/agent_type_remote_registry.rs
```diff
@@ -148,3 +208,50 @@ fn test_remote_registry_errors_on_missing_agent_type_with_oci_registry() {
...
+        test_oci_client(),
+    )
+    .expect("registry should build");
+
+    // Check embedded agent-type
+    let definition = registry
+        .get(&embedded_id)
+        .expect("embedded agent type should resolve without an OCI registry");
+    assert_eq!(definition.metadata.id, embedded_id);
+    assert_eq!(definition.metadata.environment, Environment::K8s);
```
danielorihuela, line 240: If you still want the integration test, I think we don't need to test hits and misses. This is already done in tests like `test_error_falls_through_to_next_layer`

agent-control/tests/on_host/agent_type_remote_registry.rs
danielorihuela: It's just checking that we hit the `CachingRegistry` and the `RemoteRegistry` is never called.

sigilioso: I think the main use case for the integration test is using the same Registry (the composite with Local + Cached Remote) that is used in the production code. We had unit tests covering most of the features of each registry implementation but we didn't have any testing the _production_ part.

agent-control/tests/on_host/agent_type_remote_registry.rs
sigilioso: Same reasoning as the previous comment, this is just a test to check the expected behavior of the Registry returned by `build_agent_type_registry` which is the _builder_ actually used.

Unit tests (and previous integration tests) where testing each component individually, but we lacked an integration test testing that. There is definitely overlap (and it is expected) and the coverage of this integration test is not good (we have unit tests for that) but it will help if we change the builder.

danielorihuela: You are right, and I still don't like it haha. I feel like it's unnecessary.

We have "unit" tests (mocks) to check layering order.
We have integration tests to check remote registry works.
We have e2e tests to check the product works.

We are now adding integration tests to check that the production registries are performed in the order we want. This could be done with a unit tests (mocks) to check the resulting registries types and order.

I'm not sure what, but there's something in the current shape that doesn't convince me.

danielorihuela (approved):

PR #2747, opened 2026-07-28
feat(canaries): alert when the Agent Control supervisor is unhealthy

DavSanchez:
The canary alerts only detected Agent Control (AC) disappearing **entirely** — they never warned when AC was alive but a supervised sub-agent was **unhealthy** (e.g. the Windows infra-agent crash-looping on a log-file permission error). On Windows, AC self-instrumentation was never enabled, so no AC telemetry reached New Relic at all.

This PR adds real "agent unhealthy" detection, enables self-instrumentation for windows and performs some minor adjustments.


sigilioso:
> Agent Control error logs (new) — AC-internal hard errors (severity.text = 'ERROR').

Have we missed such alerts in the past or do we have any particular situation in mind?

I fear of introducing noise in case of transient situations. Eg: temporal networking issues.

DavSanchez:
> Have we missed such alerts in the past or do we have any particular situation in mind?

Not really, just improving over the existing "log presence" alert.

> I fear of introducing noise in case of transient situations. Eg: temporal networking issues.

Yeah I agree this might cause noise with e.g. random OpAMP hiccups or whatever... not sure if we have a number that we can trust will be valuable per time window (like 5-10 over 5 minutes)?

test/k8s-canaries/terraform/production/alert_nrql_templates/log_presence.tftpl
```diff
@@ -1,4 +1,4 @@
 SELECT count(*)
 FROM Log
-WHERE cluster_name = '${instance_id}'
+WHERE clusterName = '${instance_id}'
```
sigilioso, line 3: Out of curiosity, how was this alert not triggering before? 🤔

test/k8s-canaries/terraform/production/main.tf
```diff
@@ -111,5 +111,28 @@ module "alerts" {
...
     },
+    {
+      # Fires when a supervised sub-agent reports unhealthy. AC logs an `AgentHealthInfo` event on
+      # every health-check tick (~30s) whose `health` attribute reads `Healthy(...)`/`Unhealthy(...)`;
+      # we match the unhealthy variant. This is the direct "AC is alive but a sub-agent is unhealthy"
+      # signal — e.g. an executable crash-looping or a failing health check — which the metric/log
+      # presence conditions miss because AC itself keeps running and emitting.
+      name               = "Agent Control supervisor unhealthy"
+      threshold          = 0
+      duration           = 300
```
danielorihuela: I run `SELECT * FROM Log WHERE clusterName = 'Agent_Control_Canaries_Staging-Cluster' AND service.name = 'agent-control-self-instrumentation' AND message = 'AgentHealthInfo' AND health LIKE '%Unhealthy%'` and saw the following:

Unhealthy at 12:34 (missing pods)
Unhealthy at 12:40 (missing pods)

Image

Image

Won't we get false positives with this query? I don't know if we can check for a more specific unhealthy error.

DavSanchez: Maybe we can increase match amounts per time window for this to be meaningful? What I mean is, with that same example, if these unhealthy situations do not recover then this _**is**_ an actionable issue no?

danielorihuela: Yes, it's tricky.

test/k8s-canaries/terraform/production/alert_nrql_templates/log_presence.tftpl
DavSanchez: I wondered the same over a call with @alvarocabanas , I think that absence of signal (because the `WHERE` does not match) is either not properly matching the "0 or less" threshold or it matched at some point, was muted or similar and given it "never recovers" there was no new alert to trigger, it was the same alert all along that we just ignored.

DavSanchez: (Claude says it checked the history of the alert and there has not been any trigger 🤔)

test/k8s-canaries/terraform/production/main.tf
danielorihuela: Maybe we can increase the time and call it a day? Like, if it's unhealthy for 20 minutes straight, it should be an error.

Or maybe the 5 minutes is okay and what I'm seeing on the console is a real error 🤷

test/k8s-canaries/terraform/production/alert_nrql_templates/agent_unhealthy.tftpl
```diff
@@ -0,0 +1,6 @@
+SELECT count(*)
+FROM Log
```
gsanchezgavier: have you explored the heartbeat samples? i think there is health there

test/k8s-canaries/terraform/production/alert_nrql_templates/agent_unhealthy.tftpl
```diff
@@ -0,0 +1,6 @@
...
+WHERE clusterName = '${instance_id}'
+AND service.name = 'agent-control-self-instrumentation'
+AND message = 'AgentHealthInfo'
+AND health LIKE '%Unhealthy%'
```
gsanchezgavier: not sure how easy would be to add a metric instead of this, which we would think twice before modifying it

test/k8s-canaries/terraform/production/main.tf
DavSanchez: I just did: https://github.com/newrelic/newrelic-agent-control/pull/2747/commits/baf254ceaa814494da19233304880f6b044e4291

test/k8s-canaries/terraform/production/alert_nrql_templates/agent_unhealthy.tftpl
DavSanchez: do you have an example query?

test/k8s-canaries/terraform/production/alert_nrql_templates/agent_unhealthy.tftpl
DavSanchez: a metric that we would emit via self-instrumentation? I'd say it should be possible, as separate PR before this one 🤔

test/k8s-canaries/terraform/production/alert_nrql_templates/log_presence.tftpl
sigilioso: Should we dig deeper into it? Just in case the _fixed_ alert presents the same issue. The alert is there to be triggered precisely in the absence of the signal, right?

DavSanchez: I explored this for a bit and it seemed like a pretty involved change, I'll try to give it another shot but on a separate PR so I don't mix concerns on this one.

test/k8s-canaries/terraform/production/alert_nrql_templates/agent_unhealthy.tftpl
DavSanchez: Alerts tested in staging and working! Created in our account, pointing to the Fleet account. We can fine-tune the query and any messages as per your review.

test/fleet-canary-alerts/terraform/production/alert_nrql_templates/agent_heartbeat_unhealthy.tftpl
```diff
@@ -0,0 +1,3 @@
+SELECT filter(count(*), WHERE agentHealthStatus IS FALSE)
+FROM AgentHeartbeat
+FACET agentType, fleetGuid
```
gsanchezgavier, line 4: if we are not filtering by fleet id how we can false positives from any other fleet in that org?

test/fleet-canary-alerts/terraform/production/main.tf
```diff
@@ -0,0 +1,82 @@
...
+
+variable "account_id" {
+  description = "New Relic canary telemetry account ID the alert resources are created in (the production canary account, i.e. NEW_RELIC_PROD_ACCOUNT_ID). Write access required."
+  type        = string
+}
+
+variable "data_account_id" {
+  description = "New Relic fleet-data account ID that receives AgentHeartbeat events; the NRQL condition queries it cross-account. Production = 6425865 (read access required)."
+  type        = string
+  default     = "6425865"
```
gsanchezgavier: would make sense to pass all account ids trhough vars that are taken from gh secrets/var ?

test/fleet-canary-alerts/terraform/production/alert_nrql_templates/agent_heartbeat_unhealthy.tftpl
DavSanchez: Fixed!

test/fleet-canary-alerts/terraform/production/main.tf
```diff
@@ -24,6 +24,21 @@ variable "data_account_id" {
...
+# customer's agents and page us for their unhealthy agents. These are the production canary fleet entity
+# GUIDs, i.e. the FLEET_ID_PRODUCTION / WINDOWS_FLEET_ID_PRODUCTION values the canaries are provisioned
+# with. Keep in sync with component_onhost_canaries.yml and component_k8s_canaries.yml.
+variable "fleet_guids" {
+  description = "Canary fleet entity GUIDs to restrict the AgentHeartbeat query to (production)."
+  type        = list(string)
+  default = [
+    "NjQyNTg2NXxOR0VQfEZMRUVUfDAxOWFlM2EyLTA3OTctNzczYS05Y2JjLWMzNzZkMjAwMWFkZg", # on-host Linux (FLEET_ID_PRODUCTION)
+    "NjQyNTg2NXxOR0VQfEZMRUVUfDAxOWMyMjljLWVhYTQtN2JmNi04YWIyLTU2ZWI0YjE2ZWZjZQ", # on-host Windows (WINDOWS_FLEET_ID_PRODUCTION)
+    "NjQyNTg2NXxOR0VQfEZMRUVUfDAxOTVlYzZhLTYwNjItNzZiZS1iOWE0LTA4MzZlZmE2MGY4YQ", # k8s (FLEET_ID_PRODUCTION)
```
gsanchezgavier: aren't we passing these from the CI for the e2e? i would try to have a single source of truth

gsanchezgavier (commented):
oh sorry i forgot to submit these comments I think it looks good, i would just remove all the pinned data and try to pass it from env vars that are declared in github vars or secrets so we have a single source for these

test/fleet-canary-alerts/terraform/production/main.tf
DavSanchez: Fixed — moved the fleet GUIDs and fleet-data account IDs into repo variables (ONHOST_FLEET_ID_*, ONHOST_WINDOWS_FLEET_ID_*, K8S_FLEET_ID_*, FLEET_DATA_ACCOUNT_ID_*), wired into component_onhost_canaries.yml, component_k8s_canaries.yml, and the fleet-canary-alerts plan/apply/destroy targets. The terraform defaults are gone, single source of truth now.

test/fleet-canary-alerts/terraform/production/main.tf
DavSanchez: Fixed — same fix as below: data_account_id is now required (no default), passed in from CI via the FLEET_DATA_ACCOUNT_ID_STAGING/PRODUCTION repo variables.

gsanchezgavier (approved):
lgtm

PR #2748, opened 2026-07-28
feat: add postgresql custom queries support

alvarocabanas:
Add postgresql custom queries support
----------------------------------------

We are setting an env var in the infra-agent  `AC_SHARED_CFG: "${nr-sub:shared_filesystem_dir}/infra-agent-ohi-configs"`.
Then, we propagate it to the integrations thanks to the fact that `{{ AC_SHARED_CFG }}` is rendered by the infra agent both in the infra config and integrations config.
We will need to set the template for postgresql with something like:
```
  config:
    integrations:
    - env:
        HOSTNAME: 127.0.0.1
        PORT: "1123"
        # Path to custom SQL queries for additional metric collection;                                                                                                                                                              
        # resolved at runtime by Agent Control from the shared config directory
        CUSTOM_METRICS_CONFIG: {{ AC_SHARED_CFG }}/postgresql-custom-query.yaml
      interval: 15s
      name: nri-postgresql
```

### DIFF
--- a/agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure-0.1.0.yaml
+++ b/agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure-0.1.0.yaml
@@ -129,6 +129,7 @@ deployment:
         NRIA_STATUS_SERVER_ENABLED: true
         NRIA_STATUS_SERVER_PORT: "${nr-var:health_port}"
         NR_HOST_ID: "${nr-ac:host_id}"
+        AC_SHARED_CFG: "${nr-sub:shared_filesystem_dir}/infra-agent-ohi-configs"
       restart_policy:
         backoff_strategy:
           type: fixed
--- a/agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure.nri_postgresql-0.1.0.yaml
+++ b/agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure.nri_postgresql-0.1.0.yaml
@@ -9,6 +9,10 @@ variables:
     description: "nri-postgresql integration YAML consumed by the infra-agent"
     type: yaml
     required: true
+  custom_queries:
+    description: "custom queries to be executed by the nri-postgresql integration"
+    type: yaml
+    required: false
   version:
     description: "Agent version"
     type: string
@@ -38,6 +42,10 @@ deployment:
           kind: file
           text: |
             ${nr-var:config}
+        postgresql-custom-query.yaml:
+          kind: file
+          text: |
+            ${nr-var:custom_queries}
     infra-agent-ohi-binaries:
       kind: dir
       entries:
--- a/agent-control/agent-type-registry/newrelic/host-windows-com.newrelic.infrastructure-0.1.0.yaml
+++ b/agent-control/agent-type-registry/newrelic/host-windows-com.newrelic.infrastructure-0.1.0.yaml
@@ -133,6 +133,7 @@ deployment:
         NRIA_STATUS_SERVER_ENABLED: true
         NRIA_STATUS_SERVER_PORT: "${nr-var:health_port}"
         NR_HOST_ID: "${nr-ac:host_id}"
+        AC_SHARED_CFG: "${nr-sub:shared_filesystem_dir}/infra-agent-ohi-configs"
       restart_policy:
         backoff_strategy:
           type: fixed
--- a/agent-control/agent-type-registry/newrelic/host-windows-com.newrelic.infrastructure.nri_postgresql-0.1.0.yaml
+++ b/agent-control/agent-type-registry/newrelic/host-windows-com.newrelic.infrastructure.nri_postgresql-0.1.0.yaml
@@ -9,6 +9,10 @@ variables:
     description: "nri-postgresql integration YAML consumed by the infra-agent"
     type: yaml
     required: true
+  custom_queries:
+    description: "custom queries to be executed by the nri-postgresql integration"
+    type: yaml
+    required: false
   version:
     description: "Agent version"
     type: string
@@ -38,6 +42,10 @@ deployment:
           kind: file
           text: |
             ${nr-var:config}
+        postgresql-custom-query.yaml:
+          kind: file
+          text: |
+            ${nr-var:custom_queries}
     infra-agent-ohi-binaries:
       kind: dir
       entries:


agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure-0.1.0.yaml
```diff
@@ -129,6 +129,7 @@ deployment:
         NRIA_STATUS_SERVER_ENABLED: true
         NRIA_STATUS_SERVER_PORT: "${nr-var:health_port}"
         NR_HOST_ID: "${nr-ac:host_id}"
+        AC_SHARED_CFG: "${nr-sub:shared_filesystem_dir}/infra-agent-ohi-configs"
```
gsanchezgavier, line 132: was enough to be env var of the infra agent? I remember some `passthrough` configurations

gsanchezgavier, line 132: shouldn't we use another folder for this? i think the infra-agent will try to "execute" this file as an integration config and it might produce some warn logs

alvarocabanas: No, the passthrough is another thing, the passthrough allow the integration to access an env var, and it has the problem that if you define it in the agent_type, it overwrites whatever the user has set on their config.
This instead sets an env var that the template engine from the infra-agent will substituted when it finds {{ NAME_OF_WHATEVER_ENV_VARIABLE_ACCESSIBLE_BY_THE_INFRA_AGENT }}

alvarocabanas: The infra-agent only parses configs matching the right schema with the `integrations:`  key. Based on what I could see from most customers, they tended to put it in the same place as the integration config, I would do it like this, but for sure we could set another directory if needed.

gsanchezgavier: cool , no passthrough then 👍

gsanchezgavier: 👍  as long there is no recurrent logs sounds good, and is even they place for the sample config https://github.com/newrelic/nri-postgresql/blob/master/build/.goreleaser.yml#L80

gsanchezgavier, line 132: what about `AC_INTEGRATIONS_CONFIG_PATH` ?

sigilioso:
Nice approach! @alvarocabanas, @gsanchezgavier.

Considering that these names will be likely _exposed_ in remote configs. I'd double-check them with PM(s) before merging.

I'd like to suggest some cosmetic changes:

- `AC_SHARED_CFG` --> `AC_SHARED_CONFIG_DIR`
- `postgresql-custom-query.yml` -->  `nri-postgresql-custom-queries.yml`

agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure.nri_postgresql-0.1.0.yaml
```diff
@@ -38,6 +42,10 @@ deployment:
           kind: file
           text: |
             ${nr-var:config}
+        postgresql-custom-query.yaml:
```
gsanchezgavier, line 45: it would be even better to pass the full path including this as an envar , not sure if it is possible, otherwise this is a bit hidden for the user configuring the agent

agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure-0.1.0.yaml
sigilioso, line 132: I'd document somewhere (maybe as a comment in the infrastructure Agent Types) why is this exposes as variable and how can it be used.

agent-control/agent-type-registry/newrelic/host-linux-com.newrelic.infrastructure.nri_postgresql-0.1.0.yaml
alvarocabanas: We don't have access to the env vars passed to the integration from infra-agent, we can only use passthrough, but if we set it in the agent_type, any possible passthrough variable defined by the user will be overwritten and ignore by our custom passthrough.
So we require to write in the config yaml.
The only modification that I am not a big fan of, is having something like `POSTGRESQL_CUSTOM_QUERIES_PATH: "${nr-sub:shared_filesystem_dir}/infra-agent-ohi-configs/postgresql-custom-query.yaml"`
in the infra agent instead of AC_SHARED_CFG so then we require:
`CUSTOM_METRICS_CONFIG: {{ POSTGRESQL_CUSTOM_QUERIES_PATH }}`
But I don't like that the infra-agent will have something specific of postgresql, and we will need to do the same for any integration requiring it.

gsanchezgavier: yes there is no much we can improve with the current tools. All this should be documented in the agent type, for instance a big note that says how breaking would be to change that config file name give how this is used

agent-control/agent-type-registry/newrelic/host-windows-com.newrelic.infrastructure-0.1.0.yaml
```diff
@@ -133,6 +133,7 @@ deployment:
         NRIA_STATUS_SERVER_ENABLED: true
         NRIA_STATUS_SERVER_PORT: "${nr-var:health_port}"
         NR_HOST_ID: "${nr-ac:host_id}"
+        AC_SHARED_CFG: "${nr-sub:shared_filesystem_dir}/infra-agent-ohi-configs"
```
paologallinaharbur, line 136: what I do not like is that in case of a password the user will need to leverage mixed notation and know the name of an environmentVariable:
```
config:
    integrations:
    - env:
        HOSTNAME: 127.0.0.1
        PORT: "1123"
        PASSWORD: ${nr-env:PASSWORD}
        CUSTOM_METRICS_CONFIG: {{ AC_SHARED_CFG }}/postgresql-custom-query.yml
```

I wonder if it would be better instead to expose just the base path
```
config:
    integrations:
    - env:
        HOSTNAME: 127.0.0.1
        PORT: "1123"
        PASSWORD: ${nr-env:PASSWORD}
        CUSTOM_METRICS_CONFIG: ${nr-sub:base-dir}/shared-filesystem/infra-agent-ohi-configs/postgresql-custom-query.yml
```

gsanchezgavier:
Hey! Just tidying up the board a bit 🧹 — I'm closing this for now, but no worries, this isn't a rejection at all! Feel free to reopen it whenever it's ready for another review/merge. Just wanted to keep things clean and easy to track. 🙌

PR #2749, opened 2026-07-28
feat: send supported protocol version on release

jcoscollanr:


bfollon-new-relic (approved):

gsanchezgavier (changes requested):
[here](https://github.com/newrelic/newrelic-agent-control/blob/main/agent-control/Cargo.toml#L13) we specify the protocol version . we should reuse it or refactor it to have a single source of truth

PR #2750, opened 2026-07-29
feat(logs): add maximum retention of 30 days

paologallinaharbur:
This PR is:
 - adding .max_log_files(30) when building the loggers
> When constructing a RollingFileAppender or starting a new log file, the appender will delete the oldest matching log files until at most n files remain. 
- introduced a new helper struct to clarify that `FileLoggingConfigSubAgent` is completely different from `FileLoggingConfig`. 

Note that with @gsanchezgavier we decided not to expose the max_file since we currently do not have a section in the config for the log of the subagents, moreover the "DAILY" rotation was already hardcoded

### DIFF
--- a/agent-control/src/sub_agent/on_host/command/logging/file_logger.rs
+++ b/agent-control/src/sub_agent/on_host/command/logging/file_logger.rs
@@ -23,7 +23,7 @@ const MAX_LOG_FILES_SUB_AGENT: usize = 30;
 pub struct FileLoggerError(String);
 
 #[derive(Debug, Deserialize, Default, PartialEq, Clone)]
-pub struct FileLoggingConfigSubAgent {
+pub struct SubAgentFileLoggingConfig {
     pub enabled: bool,
     // Default value is being set by `ConfigPatcher` right after deserialization.
     pub base_path: PathBuf,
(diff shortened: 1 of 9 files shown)


docs/CONFIG.md
```diff
@@ -66,6 +66,8 @@ log:
   show_spans: false # Show spans information as logs. It includes additional details which may be useful for debugging purposes.
 ```
 
+Notice that rotation and retention cannot be configured and is fixed to be daily with 15 days of retention.
```
sigilioso:
```suggestion
Notice that rotation and retention cannot be configured and is fixed to be daily with 30 days of retention.
```

sigilioso (commented):
I've just pointed out a _typo_ in the docs.

I was about to say that we should make it configurable, but the reason why not making it configurable is already stated in the PR description.

agent-control/src/sub_agent/on_host/command/logging/file_logger.rs
```diff
@@ -12,17 +15,33 @@ use tracing_subscriber::{
...
 
+const MAX_LOG_FILES_SUB_AGENT: usize = 30;
+
 /// Error produced while building a file logger.
 #[derive(Debug, Error)]
 #[error("{0}")]
 pub struct FileLoggerError(String);
 
+#[derive(Debug, Deserialize, Default, PartialEq, Clone)]
+pub struct FileLoggingConfigSubAgent {
```
DavSanchez: nit: `SubAgentFileLoggingConfig`?

agent-control/src/sub_agent/on_host/command/logging/file_logger.rs
```diff
@@ -12,17 +15,33 @@ use tracing_subscriber::{
...
+    pub base_path: PathBuf,
+}
+
 /// Creates a new file logger writing to a file in the provided directory with the provided suffix.
 /// The file will be rotated daily and the file name will be in the format `<timestamp>.<suffix>`
-/// e.g. `2027-12-01.stdout.log`.
-pub fn file_logger(file_dir: &Path, file_name_suffix: &str) -> Result<FileLogger, FileLoggerError> {
+/// e.g. `2027-12-01.stdout.log`. Only the MAX_LOG_FILES_SUB_AGENT most recent rotated files are retained.
+pub fn file_logger(
+    agent_id: AgentID,
```
DavSanchez: take a reference instead?

docs/CONFIG.md
paologallinaharbur: done!

agent-control/src/sub_agent/on_host/command/logging/file_logger.rs
paologallinaharbur: done!

DavSanchez (approved):

PR #2758, opened 2026-07-30
fix: improve oci client error logs

gsanchezgavier:
Make use of the existent `OciErrorMessage` conversion improving the logging in situations where the target tag is missing in the registry for packages and agent types. 
Fixes a regression introduced with the retries #2610 related to log improvements.


danielorihuela (dismissed):

sigilioso (approved):
LGTM

PR #2759, opened 2026-07-30
chore(vars): refactor NamespacedVariableName

paologallinaharbur:
The main change of this PR is having a concrete type for NamespacedVariableName

Therefore
```rust
pub type NamespacedVariableName = String;
```

Became
```rust
#[derive(Debug, Clone, PartialEq, Eq, Hash)]
pub struct NamespacedVariableName {
    namespace: Namespace,
    name: String,
}
```

This is important since in many parts of the code we were using String as if it was NamespacedVariableName and since it was just alias it was impossible to tell the difference between:
```rust
        variables: HashMap<String, Variable>,   with prefix
        env_vars: HashMap<String, Variable>,   without prefix
```


agent-control/src/agent_type/variable/namespace.rs
```diff
@@ -4,7 +4,33 @@ use strum::{EnumIter, IntoEnumIterator};
 
 /// Holds the variable name prefixed with the namespace.
 /// Example: "nr-env:MY_ENV_VAR" for the environment variable "MY_ENV_VAR".
-pub type NamespacedVariableName = String;
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct NamespacedVariableName {
```
sigilioso: Nit: I'd consider the namespace an "implementation detail". We could name this struct `VariableName` and maybe move it to the parent module (`crate::agent_type::variable`). WDYT?

sigilioso (dismissed):
Thanks for introducing the type!

I've left a naming suggestion, feel free to ignore it!

agent-control/src/agent_type/variable/namespace.rs
paologallinaharbur: let me check, but we have already have "variable"

paologallinaharbur: done!

sigilioso (approved):
LGTM

PR #2760, opened 2026-07-30
fix(fs): make Windows admin ACE inheritable

DavSanchez:
# Windows: infra sub-agent couldn't open its own log after an upgrade

## What was broken
On Windows, the infrastructure sub-agent kept crash-restarting because it couldn't open its own log file — Windows returned **"Access is denied"**, even though the agent runs as the highest-privilege system account.

## Why, in plain terms
Every file on Windows carries a list of "who is allowed to touch this." Agent Control locks its folders down so only **administrators** can access them. A file created inside such a folder normally **inherits** that "administrators allowed" entry automatically.

The bug: when Agent Control locked a folder, it marked the "administrators allowed" entry as **"do not pass this down to the files inside."** When Windows then recalculates an inside file's permissions and finds nothing to inherit, it leaves that file with an **empty list. An empty list means "nobody is allowed" — not even the agent itself.** So the sub-agent could no longer open its log, failed to start, and crash-looped.

## Fresh install vs upgrade — why only the upgrade breaks

```mermaid
flowchart TD
    subgraph FRESH["Fresh install (works)"]
        direction TB
        A1["AC locks the folder<br/>(entry NOT passed down)"] --> A2["THEN agent creates the log"]
        A2 --> A3["log gets its OWN explicit<br/>'administrators allowed' entry"]
        A3 --> A4["later re-locks only recalc<br/>INHERITED entries"]
        A4 --> A5["explicit entry survives<br/>✅ log stays openable"]
    end
    subgraph UPGRADE["Upgrade older to 1.18.0 (breaks)"]
        direction TB
        B1["old version: folder NOT locked"] --> B2["agent creates the log<br/>log INHERITS entry from parent"]
        B2 --> B3["upgrade: 1.18.0 locks the folder<br/>for the FIRST time, not-passed-down"]
        B3 --> B4["Windows recalculates:<br/>nothing left to inherit"]
        B4 --> B5["log left with EMPTY list<br/>= nobody allowed, incl. SYSTEM"]
        B5 --> B6["❌ agent can't open its log<br/>crash loop"]
    end
```

- **A fresh install is safe.** The folder is locked **before** the log is created, so the log gets its **own explicit** entry that survives re-locking.
- **Restarting the same version changes nothing.** A healthy log stays healthy; an already-emptied log stays empty. The damage happens exactly once — at the upgrade, when a folder-locking version first meets a log created before that folder was ever locked.

### Part 1 — make the entry inheritable (stops it happening)
Mark the folder's "administrators allowed" entry as **inheritable** ("pass this down to the files inside"). Files created inside then inherit administrator access instead of ending up with an empty list. This prevents the bug on fresh installs and future upgrades.

```mermaid
flowchart LR
    subgraph BEFORE["Before"]
        direction TB
        P1["Folder entry:<br/>administrators allowed<br/>DO NOT pass down"] -. "nothing to inherit" .-> C1["Log: EMPTY list<br/>❌ nobody allowed"]
    end
    subgraph AFTER["After (Part 1)"]
        direction TB
        P2["Folder entry:<br/>administrators allowed<br/>PASS DOWN (inheritable)"] == "inherited" ==> C2["Log: administrators allowed<br/>✅ agent (SYSTEM) can open"]
    end
```

### Part 2 — repair machines already broken by an upgrade
Part 1 **does not** fix machines that an upgrade already broke — verified in the field. Two reasons:
1. Agent Control **never re-locks a folder that already exists** — it only sets permissions when *creating* one. So on upgrade the new inheritable entry is never applied to existing folders.
2. The wiped files aren't limited to one folder. They're spread across **everything Agent Control stores under its data root** — the sub-agent's **log** and **data folders**, but also **stored remote (fleet) configs** and local data — and most of these are never re-touched after creation.

Result: the empty lists are never repaired. Symptoms range from a sub-agent that can't open its log (crash loop) to **"Access is denied" when deleting a stored remote config** while decommissioning a sub-agent via a fleet deployment. (Confirmed in the field: a Part‑1‑only binary let the log recover, but stored remote configs stayed locked and decommission still failed with `os error 5`.)

The fix adds an explicit **recursive repair at Agent Control startup, over the whole managed data tree** (sub-agent filesystem, stored remote configs, local data — everything under Agent Control's data root). It walks the tree once and re-stamps the administrators entry **only on entries that are actually broken** — an empty (or unreadable) permission list — leaving healthy entries untouched, so a healthy install isn't rewritten on every boot. Because Agent Control **owns** these files, the rewrite succeeds even on an empty list, and it restores read/write **and delete** — so both the crash loop and the decommission-cleanup failure are fixed (a healed file now permits delete). This is the **single** repair entry point — there's no per-render or per-delete repair to reason about.

```mermaid
flowchart TD
    S["Agent Control start<br/>(after upgrading to this fix)"] --> R["Recursive repair over the WHOLE data root:<br/>re-stamp 'administrators allowed' on every<br/>existing file & folder (we own them)"]
    R --> L["empty sub-agent log → allowed ✅<br/>(opens — no crash loop)"]
    R --> D["empty data / local-data → allowed ✅"]
    R --> F["empty stored remote configs (fleet-data) → allowed ✅"]
    F --> DEL["decommission can delete them now ✅<br/>(previously 'Access is denied' — empty list denied delete)"]
```

- **No new access is granted** — administrators-only throughout (SYSTEM qualifies because it is an administrator).
- **Broken machines heal on the next Agent Control start** after upgrading to this fix — no manual intervention.

## How we verified it
- **Reproduced** the exact failure on a clean Windows VM: agent-control **1.17.0** (works) → **1.18.0** (breaks). The resulting permissions matched a real affected machine byte-for-byte. Full detail in **NR-601065**.
- **Confirmed iteratively in the field**: a Part‑1‑only binary didn't heal at all (only a freshly-created temp folder picked up the new entry). Adding a render-time repair healed the sub-agent's log — but a **stored remote config under `fleet-data` stayed locked**, so decommissioning the sub-agent failed with `Access is denied` (`os error 5`). That's why the repair now runs recursively over the **whole data root at startup**, covering every managed location rather than just the sub-agent filesystem.

## Tests
- **`child_created_in_managed_directory_inherits_admin_access`** — regression test for Part 1: proves a runtime-created file inherits administrator access. Run against **unfixed** code it **fails** (the non-inheritable ACE), which is exactly what the throwaway PR #2763 demonstrates.
- **`recursive_repair_heals_existing_empty_dacl_child_from_older_version`** — a unit test *of the repair* (Part 2): recreates the empty permission list an older version leaves behind and proves the startup repair restores access and openability. It exercises the new repair code, so it lives with the fix (there's no "unfixed" counterpart to run on the throwaway).

**Throwaway PR #2763** holds the regression tests against **unfixed production code** so you can see them go red — proving they catch the bug. It intentionally contains *no* fix code.


agent-control/src/agent_control/run/on_host.rs
```diff
@@ -93,6 +93,21 @@ impl AgentControlRunner {
...
+
+        // Windows-only: check and repair the managed data tree before anything reads/writes/deletes
+        // it (see NR-601065). Fails early — we cannot ensure proper AC operation if permissions
+        // aren't right.
+        #[cfg(target_family = "windows")]
+        fs::win_permissions::ensure_managed_permissions([
+            remote_dir.as_path(),
+            local_dir.as_path(),
+        ])
+        .map_err(|e| {
```
paologallinaharbur: I would not exit in case of error with permissions, there is the risk of causing a deadlock of the application in which we are not able to update because of this, no?

DavSanchez: I see your point! I initially erred in the side of caution (just quit if we cannot ensure proper operation, but it's true that the self-upgrade can work as a healing path. I'm updating it to not stop AC but log any failures as warnings.

fs/src/win_permissions.rs
```diff
@@ -54,64 +76,261 @@ pub fn set_file_permissions_for_administrator(path: &Path) -> Result<(), Permiss
...
-        grfAccessPermissions: GENERIC_READ | GENERIC_WRITE | DELETE,
+        grfAccessPermissions: (FILE_GENERIC_READ
+            | FILE_GENERIC_WRITE
+            | FILE_GENERIC_EXECUTE
+            | DELETE)
+            .0,
         grfAccessMode: SET_ACCESS,
-        grfInheritance: NO_INHERITANCE,
+        grfInheritance: OBJECT_INHERIT_ACE | CONTAINER_INHERIT_ACE,
         Trustee: trustee,
```
gsanchezgavier, line 85: is this matching to what we set in the installation ps1?

fs/src/win_permissions.rs
```diff
@@ -157,38 +376,468 @@ pub mod tests {
...
+        fs::remove_dir_all(&managed_dir)
+            .expect("managed tree must be removable after the permission repair");
+        assert!(!managed_dir.exists());
+    }
+
+    /// Reproduces the permissions an *even older* agent-control applied to stored configs: a PROTECTED
+    /// Administrators-only ACE granting only read+write (mask `0x12019f`) with **no DELETE** and no
+    /// execute. This is the state that leaves a stored remote config undeletable on decommission
+    /// (NR-601065). Non-inheritable, matching what was observed in the field.
+    fn legacy_harden_read_write_only(path: &Path) {
```
sigilioso: Nit: I'd try to place helpers together.

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ Remember that the keywords that you can use are the following:
 
 ## Unreleased
 
+### bugfix
+- Windows: fixed files under Agent Control's managed directories being left with permissions it could not use. This could block a range of operations: a sub-agent could not modify its own assets (e.g. log rotation for the infrastructure-agent), and Agent Control could not delete a sub-agent's stored remote config or data when decommissioning it ("Access is denied"). Managed directories now use an inheritable Administrators ACE so runtime-created files inherit access, and Agent Control now preemptively checks and repairs its managed directories on startup by re-stamping any managed entry that does not already grant the required access.
```
sigilioso: Nit: I'd try to make the Changelog message shorter

agent-control/src/agent_control/run/on_host.rs
```diff
@@ -93,6 +93,21 @@ impl AgentControlRunner {
     pub fn run_onhost(self) -> Result<GracefulShutdownReason, RunError> {
         let local_dir = self.base_paths.local_dir;
         let remote_dir = self.base_paths.remote_dir;
+
+        // Windows-only: check and repair the managed data tree before anything reads/writes/deletes
+        // it (see NR-601065). Fails early — we cannot ensure proper AC operation if permissions
+        // aren't right.
+        #[cfg(target_family = "windows")]
+        fs::win_permissions::ensure_managed_permissions([
```
gsanchezgavier: i think it would be ideal if we have a FF to control this kind of behaviour. We could activated just for exiting customers for instance. Out of scope of this fix for sure

fs/src/win_permissions.rs
```diff
@@ -54,64 +76,261 @@ pub fn set_file_permissions_for_administrator(path: &Path) -> Result<(), Permiss
...
+/// Caveat: this looks for *any* matching Administrators `Allow` ACE with a sufficient mask, without
+/// regard to ACE order. It does not check whether an earlier `Deny` ACE in the same DACL would negate
+/// that `Allow` at Windows' actual access-check time — so a DACL with e.g. `[Deny Administrators:
+/// Delete][Allow Administrators: Modify]` would be reported as conforming even though delete is
+/// effectively denied. Not currently reachable: every DACL this code writes fully replaces the
+/// existing one (`SetEntriesInAclW` is called with no old ACL to merge), so
+/// `set_file_permissions_for_administrator` can never itself leave behind a stray `Deny` ACE for this
+/// function to misread. Noted here as a defense-in-depth gap in case a `Deny` ACE is ever introduced
+/// by something other than this code path (e.g. third-party security tooling).
+pub fn permissions_need_repair(path: &Path) -> bool {
```
sigilioso: Nit: do we need this to be public? (same for `ensure_permissions_recursive`)

fs/src/win_permissions.rs
```diff
@@ -1,50 +1,72 @@
...
+    DELETE, FILE_GENERIC_EXECUTE, FILE_GENERIC_READ, FILE_GENERIC_WRITE,
+};
+use windows::core::{PCWSTR, PWSTR};
 
 /// Error returned when setting Windows file permissions (ACLs) fails.
+///
+/// Internal to this crate: nothing outside `fs` calls [`set_file_permissions_for_administrator`]
+/// directly, so this never needs to cross the crate boundary. Callers within `fs` convert it to
+/// [`io::Error`] — either via the [`From`] impl below when no extra context is needed, or with a
+/// `.map_err` that adds the path being operated on.
```
sigilioso: Nit: Considering this, would it make sense to use `io::Error::other` directly?

DavSanchez: Done.

fs/src/win_permissions.rs
DavSanchez: Done.

CHANGELOG.md
DavSanchez: Done.

fs/src/win_permissions.rs
DavSanchez: Done.

agent-control/src/agent_control/run/on_host.rs
DavSanchez: This has been changed to not bail out on failure. This now logs a report of the operation outcome and continues the startup.

DavSanchez:
> [...] I'd take a look at the logs/Errors style conventions (there are some lowercase/uppercase inconsistencies).
> 
> [...] I'd address @paologallinaharbur comment regarding existing on error before merging.

Both done!

fs/src/win_permissions.rs
DavSanchez: Matches to a sub-set of it. The PowerShell script sets Full Control (`F`), here we use a least privilege approach and use only Modify (`M`).

DavSanchez:
Fixed conflicts

DavSanchez:
> ## 📦 Binary size
> Binary	main	PR	Δ
> newrelic-agent-control	21.92 MB	21.93 MB	⬆️ +0.00 MB (0.01%)
> newrelic-agent-control-cli	13.66 MB	13.66 MB	⬇️ -0.00 MB (-0.00%)

Now that I see this @vjripoll maybe it could be useful to have more rows to specify per OS/arch, because I added code that is Windows-only it won't be reflected here.

fs/src/win_permissions.rs
```diff
@@ -1,117 +1,402 @@
...
+                repaired = self.repaired.len(),
+                failed = self.failed.len(),
+                failures = %FailedEntries(&self.failed),
+                "some managed permissions could not be repaired; continuing startup"
+            );
+        } else if !self.repaired.is_empty() {
+            info!(
+                repaired = self.repaired.len(),
+                "repaired managed permissions on startup"
+            );
```
sigilioso, line 278:
```suggestion
        if !self.failed.is_empty() {
            warn!(
                repaired = self.repaired.len(),
                failed = self.failed.len(),
                failures = %FailedEntries(&self.failed),
                "Some managed permissions could not be repaired; continuing startup"
            );
        } else if !self.repaired.is_empty() {
            info!(
                repaired = self.repaired.len(),
                "Repaired managed permissions on startup"
            );
```

PR #2761, opened 2026-07-31
feat: send supported protocol version on release

jcoscollanr:
Adding to send the binding of protocol_version supported by NRAgentControl.

That allows us to decide which agents are compatible with each agent-control


gsanchezgavier (approved):
thanks for re-using the source

PR #2766, opened 2026-08-03
feat: replace map[string]yaml by string_map in agent types [NR-599201]

sigilioso:
## Summary
Introduces a `string_map` variable type that maps string keys to string values, replacing `map[string]yaml`. 

Non-string values assigned to a `string_map` are now YAML-encoded into a string rather than kept as arbitrary YAML, simplifying `TrivialValue` down to a single map variant and removing the ad-hoc re-parsing step previously needed by `dir_content_from_map`. This helps us avoiding a breaking change for existing configurations. Since non-string values are stored as its string representation the same file as before ends up stored in the corresponding entries of `dir_content_from_map`.

## Changes
- Removed `TrivialValue::MapStringYaml`; `VariableType::StringMap` now produces `TrivialValue::MapStringString`, encoding non-string merged values as YAML text via `parse_string_map`
- Simplified `DirContentFromMap` templating in `filesystem.rs` since sources are now always `HashMap<String, String>`, no per-value re-serialization needed
- Updated the `newrelic-infra` on-host agent type YAMLs and docs (`INTEGRATING_AGENTS.md`, `agent_type.md`) to use `string_map`


agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
```diff
@@ -314,34 +313,17 @@ impl Templateable for TemplateableValue<DirEntriesMap> {
         let value: HashMap<SafePath, String> = if templated_string.is_empty() {
             HashMap::new()
         } else {
-            let map_string_value: HashMap<SafePath, serde_json::Value> =
```
gsanchezgavier, line 317: are we removing the safePath protection?

agent-control/tests/on_host/scenarios/filesystem_ops.rs
```diff
@@ -195,15 +195,15 @@ deployment:
...
         agent_id.to_string(),
         format!(
             r#"
 yaml_file_contents:
   {yaml_var_content}
 some_string: "{string_var_content}"
-some_mapstringyaml:
+some_stringmap:
   file3.txt: "File 3 contents"
   file4.txt: |
```
gsanchezgavier, line 208: from your comments in the channel i though this was not possible before as we were expecting a yaml as content of the key

agent-control/src/agent_type/variable/variable_type.rs
```diff
@@ -159,4 +178,42 @@ mod tests {
...
+            final_value: None,
+        })
+    }
+
+    #[test]
+    fn string_map_merge_accepts_plain_string_value() {
+        let mut variable_type = empty_string_map();
+
+        variable_type
+            .merge_with_yaml_value(serde_json::json!({ "file.txt": "hello" }))
```
gsanchezgavier, line 195: was this working before? see the other comment

gsanchezgavier (commented):
hey looks good i left some minor comments

agent-control/src/agent_type/runtime_config/on_host/filesystem.rs
sigilioso: I totally missed that, thanks for pointing it out.

sigilioso: Actually, we are not removing it but maybe making the type explicit makes sense here because the diff also tricked me when reviewing your comment 😅

```rust 
let value: HashMap<SafePath, String> = if templated_string.is_empty() {
    HashMap::new()
} else {
    serde_saphyr::from_str(/*...*/)?
}
```

We keep deserializing to `HashMap<SafePath, String>` because of the `value`'s type.

sigilioso: While changing it to explicit, I'm having second thoughts: we just thought the `SafePath` protection was removed because of the diff, I think the codes reads better without the explicit `from_str<HashMap<SafePath, String>>`. WDYT?

agent-control/src/agent_type/variable/variable_type.rs
sigilioso: It was, the infra-agent's e2e was (and still is) using it: https://github.com/newrelic/newrelic-agent-control/blob/74007aba80253123ccc561351cbf7c4c44ba9e78/test/e2e-runner/src/linux/scenarios/infra_agent.rs#L59-L77

It is confusing because we were supporting both before and we keep supporting both:

**Before:**

- A nested object was supported and rendered as `serde_json::Value`
- A single string was supported and also rendered as `serde_json::Value` (a single string is a valid yaml after all)
- When writing the file, we serialized the `serde_json::Value` back to yaml.
- The type had `yaml` semantics.

**Now:**
- A nested object is still supported because we store its string representation as a string ([here](https://github.com/newrelic/newrelic-agent-control/blob/74007aba80253123ccc561351cbf7c4c44ba9e78/agent-control/src/agent_type/variable/variable_type.rs#L137-L141)). This _trick_ allows us supporting it and avoid a breaking change.
- A single string is still supported because we are directly storing the string value.
- When writing the file, we just write the string (for nested objects we already serialized it when storing the string representation.
- The type doesn't have yaml semantics anymore.

agent-control/tests/on_host/scenarios/filesystem_ops.rs
sigilioso: Let's discuss in the other comment, if that's OK

agent-control/src/agent_type/variable/variable_type.rs
gsanchezgavier: thanks for clarifying !

gsanchezgavier (approved):
it might worth to check with @jcoscollanr if this has any implications in the metadata service before merge it

sigilioso:
Thanks for pointing that out @gsanchezgavier. 

@jcoscollanr and I have just discussed it and it might impact on how JSON-Schemas are built from Agent Types. It is another thing to take into account when considering Agent Types ready.

PR #2767, opened 2026-08-03
chore: refactor secret retriever

paologallinaharbur:
The current location for `secret-retriever` was quite confusing and collided with `secret-providers` that was at the same level. 

Since the only method is used to "OpAMP authentication secret (Private Key)." I think the correct module was the opamp one


sigilioso (approved):
Thanks!

PR #2768, opened 2026-08-03
chore(ohi-test): point to prod

paologallinaharbur:
With This PR no new feature it is added, we are simply change the test to point to prod. 

HOST_E2E_POSTGRESQL_VERSION has been excluded since there is still no artifact for it


sigilioso (approved):
LGTM!

`nri-postgresql` and `nri-flex` can be added after updating/adding the corresponding Agent Type (when they are in the real repositories).

PR #2769, opened 2026-08-04
chore(injection): refactor envVar injection

paologallinaharbur:
The main objective is avoid reading from different points in code the environment variables. Now they are collected solely via secret retriever.
 -  now all values coming from the secret provider can be used at render time (not only environment variables)
 - a test were added to make sure the use-cases are clear and self-documented, since I was the first being confused by the implementation.

Review commit by commit

### DIFF
--- a/agent-control/src/agent_type/render.rs
+++ b/agent-control/src/agent_type/render.rs
@@ -1,7 +1,6 @@
 //! Renders an [`AgentType`] together with user values into a runtime configuration for a sub-agent.
 use crate::agent_type::definition::VariableTree;
 use crate::agent_type::{
-    agent_attributes::AgentAttributes,
     definition::AgentType,
     error::AgentTypeError,
     runtime_config::rendered::Runtime,
@@ -119,6 +118,7 @@ pub(crate) mod tests {
     use std::str::FromStr;
 
     use super::*;
+    use crate::agent_type::agent_attributes::AgentAttributes;
     use crate::agent_type::runtime_config::on_host::executable::rendered as exec_rendered;
     use crate::{
         agent_control::agent_id::AgentID,
--- a/agent-control/src/agent_type/render.rs
+++ b/agent-control/src/agent_type/render.rs
@@ -1,4 +1,5 @@
 //! Renders an [`AgentType`] together with user values into a runtime configuration for a sub-agent.
+use crate::agent_type::definition::VariableTree;
 use crate::agent_type::{
     agent_attributes::AgentAttributes,
     definition::AgentType,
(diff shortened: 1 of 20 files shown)


agent-control/src/sub_agent/effective_agents_assembler.rs
```diff
@@ -183,13 +183,22 @@ where
...
+            .try_into()
+            .map_err(|e: YAMLConfigError| SecretVariablesError::YamlParseError(e.to_string()))?;
+
+        let secret_variables_values = SecretVariables::from(config.as_str());
+        let secret_variables_runtime =
+            SecretVariables::from(format!("{:?}", agent_type.runtime_config).as_str());
+
+        let mut secrets: HashMap<VariableName, Variable> = HashMap::new();
+        secrets.extend(secret_variables_values.load_secrets(&self.secrets_providers)?);
+        secrets.extend(secret_variables_runtime.load_secrets(&self.secrets_providers)?);
```
sigilioso, line 196: Adding everything to secrets and then filtering out only env vars inside the `render` template also seems confusing to me. When reading the code I though we had introduced a behavioral change because everything is is secrets.

Maybe we should consider if `secrets_provider` should actually be an attribute of the `Renderer` implementation and don provide anything to the `render(...)` method:

```rust
let renderer = Renderer::new(/* ... */ secrets_provider);
// ...
renderer.render(agent_type, values, attributes)?
```

agent-control/src/sub_agent/effective_agents_assembler.rs
```diff
@@ -183,13 +183,22 @@ where
...
-        let env_vars = load_env_vars();
-        let secrets = secret_variables.load_secrets(&self.secrets_providers)?;
+        let config: String = values
+            .clone()
+            .try_into()
+            .map_err(|e: YAMLConfigError| SecretVariablesError::YamlParseError(e.to_string()))?;
+
+        let secret_variables_values = SecretVariables::from(config.as_str());
+        let secret_variables_runtime =
+            SecretVariables::from(format!("{:?}", agent_type.runtime_config).as_str());
```
sigilioso: I think that using the debug representation of the runtime_config to extract the referenced secret variables is risky: the debug representation is expected to be used for debugging purposes only and can changing it would lead into bugs here.

agent-control/src/sub_agent/effective_agents_assembler.rs
```diff
@@ -183,13 +183,22 @@ where
 
         // Values are expanded substituting all ${nr-env...} with environment variables.
         // Notice that only environment variables are taken into consideration (no other vars for example)
```
sigilioso: This comment was outdated before the change (We also consider secrets as `${nr-vault:...}` , `${nr-kubesec:...}`, ...). Could we take the opportunity to update it?

sigilioso (commented):
I've left some comments for us to discuss.

By the way, thanks for adding these unit tests!

agent-control/src/sub_agent/effective_agents_assembler.rs
paologallinaharbur: IMO the issue is the current behaviour, it is confusing, I still do no understand why we should threat them differently.

paologallinaharbur: > Adding everything to secrets and then filtering out only env vars inside the render template also seems confusing to me. When reading the code I though we had introduced a behavioural change because everything is is secrets.

renders uses the all, but just the env vars... IMO now it is clear and before it was confusing 😅

agent-control/src/sub_agent/effective_agents_assembler.rs
paologallinaharbur: Serialise should work, would you prefer it?

paologallinaharbur: is there a better way, that does not involve changing 100 structs?

sigilioso: Serialize, to_string, ... anything but debug! 🙂

agent-control/src/sub_agent/effective_agents_assembler.rs
paologallinaharbur: > because everything is is secrets

The are all the inputs coming from the secret provider!

paologallinaharbur: The TemplateRenderer itself could disappear, it has only one attribute (that does not make even sense IMO) and two help functions

sigilioso: As discussed _offline_, I agree! We can "merge" EffectiveAgentsAssembler and TemplateRenderer and keep only one component with a meaningful name. Let's do that in the next refactor 😄

agent-control/src/agent_type/render.rs
```diff
@@ -584,6 +543,82 @@ deployment:
...
+      kind: ObjectKind
+      metadata:
+        name: test
+        namespace: test-namespace
+      spec:
+        vault_field: ${nr-vault:V_KEY}
+        env_direct: ${nr-env:ENV}
+        from_sub_agent: ${nr-sub:agent_id}
+        values: ${nr-var:my_yaml}
+"#,
```
sigilioso, line 575: Now I see that we were half-supporting this. 

Before, an Agent type referencing to `${nr-vault:xxx}` was valid but we always failed to render it because the variables weren't valid. Now the Agent Type might work if the corresponding secret exists.

If we wanted to prevent such secrets from being used in Agent Types we would need to implement the corresponding validation.

sigilioso (approved):
LGTM

agent-control/src/agent_type/render.rs
paologallinaharbur: exactly!

PR #2770, opened 2026-08-04
fix: add default timeout to oci-client

gsanchezgavier:
adding a missing default timeout on the oci client .

### DIFF
--- a/agent-control/src/oci.rs
+++ b/agent-control/src/oci.rs
@@ -34,6 +34,12 @@ const DEFAULT_NO_RETRY_POLICY: BackoffPolicy = BackoffPolicy {
     jitter: false,
 };
 
+/// Default timeout for establishing the connection to the registry.
+const DEFAULT_OCI_CONNECT_TIMEOUT: Duration = Duration::from_secs(10);
+
+/// Default read timeout applied to every OCI operation..
+const DEFAULT_OCI_READ_TIMEOUT: Duration = Duration::from_secs(30);
+
 /// [oci_client::Client] wrapper with extended functionality.
 /// It also wraps all _async_ operations using the underlying runtime, all functions will
 /// block the current thread until completion.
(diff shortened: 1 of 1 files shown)


agent-control/src/oci.rs
```diff
@@ -34,6 +34,12 @@ const DEFAULT_NO_RETRY_POLICY: BackoffPolicy = BackoffPolicy {
     jitter: false,
 };
 
+/// Default timeout for establishing the connection to the registry.
+const DEFAULT_OCI_CONNECT_TIMEOUT: Duration = Duration::from_secs(10);
```
paologallinaharbur, line 38: are you adding this due to a failing test?

paologallinaharbur: I am asking since early this week the nightly failed, I asked claude to investigate and it told me that it was likely this missing timeone and added a series of extra logs as well

agent-control/src/oci.rs
```diff
@@ -53,7 +59,15 @@ impl Client {
         proxy_config: ProxyConfig,
         runtime: Arc<Runtime>,
     ) -> Result<Self, OciClientError> {
-        let client_config = proxy::setup_proxy(client_config, proxy_config.clone())?;
+        let mut client_config = proxy::setup_proxy(client_config, proxy_config.clone())?;
+
+        client_config
+            .connect_timeout
+            .get_or_insert(DEFAULT_OCI_CONNECT_TIMEOUT);
```
paologallinaharbur, line 66: but it was adding it in /oci/signature_verification.rs

paologallinaharbur: oh, but it is the same client, so I guess you are doing the same

agent-control/src/oci.rs
paologallinaharbur: This was the execution https://github.com/newrelic/newrelic-agent-control/actions/runs/30786396849/job/91603569158

gsanchezgavier: no, this is because when i was implementing the reachability check i thought about timing out to prevent hanging AC on initialization . As this is a feature of the client itself i did not consider adding a test was useful

agent-control/src/oci.rs
gsanchezgavier: yes this is the client abstraction, so any other abstractino using oci client should use this and will leverage the default timeout

PR #2771, opened 2026-08-04
feat: agent type repo reachability probe

gsanchezgavier:
## Summary

Fleet Control needs to know whether the configured agent type OCI repository is reachable and
accessible so it can limit remote agent types to eligible customers. This adds a startup probe that
lists repository tags and logs the outcome.

## Changes

- Add `oci::Client::reachability_probe`, checking repository access via `list_tags`
- Run the probe in `AgentControlRunner::try_new` and log the outcome at debug level

## Notes for reviewers

- WIP: the probe result is intentionally discarded; a follow-up PR wires it to OpAMP custom capabilities.


agent-control/src/oci.rs
```diff
@@ -163,6 +164,26 @@ impl Client {
...
+        &self,
+        registry: &Registry,
+        repository: &Repository,
+        auth: RegistryAuth,
+    ) -> Result<(), OciClientError> {
+        let reference = Reference::with_tag(
+            registry.to_string(),
+            repository.to_string(),
+            // This tag is just used to build the reference.
+            "unused-tag".to_string(),
```
DavSanchez, line 178: I guess `String::default()` just doesn't work.

gsanchezgavier: havent' check really, i though this was more clear. IIRC the library uses latest wheenver the tag is empty and i though that might create confusion

agent-control/src/agent_control/run.rs
```diff
@@ -142,6 +143,28 @@ impl AgentControlRunner {
...
+                context
+                    .bootstrap_config
+                    .oci
+                    .auth
+                    .as_ref()
+                    .map(RegistryAuth::from)
+                    .unwrap_or(RegistryAuth::Anonymous),
+            )
+            .inspect(|_| debug!("Agent type registry reachability probe succeeded"))
+            .inspect_err(|err| debug!(%err, "Agent type registry reachability probe failed"));
```
sigilioso: Nit:

```suggestion
            .inspect_err(|err| debug!("Agent type registry reachability probe failed: {err}"));
```

agent-control/src/oci.rs
```diff
@@ -163,6 +164,26 @@ impl Client {
...
+        );
+        // Tags are discarded so we just ask for 1.
+        self.runtime
+            .block_on(self.client.list_tags(&reference, &auth, Some(1), None))
+            .map_err(|err| OciClientError::ListTags(err.into()))?;
+        Ok(())
```
sigilioso, line 184: Should we have retries to handle transient errors?

gsanchezgavier: I would rather not as this this is called on initialisation blocking the AC start up

agent-control/src/agent_control/run.rs
gsanchezgavier: oh true, old standards

agent-control/src/oci.rs
sigilioso: Makes sense! I actually had doubts when writing the comment, false negatives worry me a bit but it is a tradeoff

